PR #5, opened 2016-12-20
handling escape delimiters

russorat:
Fix for: https://issues.cask.co/browse/HYDRATOR-1219

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -24,7 +24,7 @@
   <packaging>jar</packaging>
 
   <name>Wrangler Plugin</name>
-  <description>A interative tool for data cleansing and transformation.</description>
+  <description>A interactive tool for data cleansing and transformation.</description>
 
   <licenses>
     <license>
@@ -74,6 +74,7 @@
     <cdap.version>4.0.0-SNAPSHOT</cdap.version>
     <commons-csv.version>1.4</commons-csv.version>
     <guava.version>19.0</guava.version>
+    <commons-lang.version>2.6</commons-lang.version>
     <!-- properties for script build step that creates the config files for the artifacts -->
     <widgets.dir>widgets</widgets.dir>
     <docs.dir>docs</docs.dir>
@@ -121,6 +122,11 @@
       <groupId>com.google.guava</groupId>
       <artifactId>guava</artifactId>
       <version>${guava.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>commons-lang</groupId>
+      <artifactId>commons-lang</artifactId>
+      <version>${commons-lang.version}</version>
     </dependency>
   </dependencies>
 
--- a/pom.xml
+++ b/pom.xml
@@ -24,7 +24,7 @@
   <packaging>jar</packaging>
 
   <name>Wrangler Plugin</name>
-  <description>A interactive tool for data cleansing and transformation.</description>
+  <description>An interactive tool for data cleansing and transformation.</description>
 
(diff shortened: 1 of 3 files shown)


pom.xml
```diff
@@ -24,7 +24,7 @@
   <packaging>jar</packaging>
 
   <name>Wrangler Plugin</name>
-  <description>A interative tool for data cleansing and transformation.</description>
+  <description>A interactive tool for data cleansing and transformation.</description>
```
awholegunch: An interactive

pom.xml
```diff
@@ -74,6 +74,7 @@
     <cdap.version>4.0.0-SNAPSHOT</cdap.version>
```
awholegunch: <cdap.version>4.0.0</cdap.version>

pom.xml
```diff
@@ -74,6 +74,7 @@
...
     <commons-csv.version>1.4</commons-csv.version>
     <guava.version>19.0</guava.version>
+    <commons-lang.version>2.6</commons-lang.version>
```
awholegunch, line 76: this should be alphabetical, ie as line 76...

src/test/java/co/cask/wrangler/steps/PipelineTest.java
```diff
@@ -93,4 +95,26 @@ public void testSplitWithNull() throws Exception {
     Assert.assertEquals("1,2,a,A", row.getValue("firstCol"));
     Assert.assertNull(row.getValue("secondCol"));
   }
+
+  @Test
+  public void testEscapedStrings() throws Exception {
+    List<Step> steps = new ArrayList<>();
+    Row row = new Row("__col", StringEscapeUtils.unescapeJava("1\\ta"));
+
+    TextSpecification ts = new TextSpecification("set format csv \\t false\n" +
```
awholegunch, line 104: Why `\\t` and yet `\n`?

russorat: It has to do with how the parsing happens when the specifications come from the UI. The UI translates the "tab" character into "\t" but keeps the "newline" as a newline character.

src/test/java/co/cask/wrangler/steps/PipelineTest.java
```diff
@@ -93,4 +95,26 @@ public void testSplitWithNull() throws Exception {
     Assert.assertEquals("1,2,a,A", row.getValue("firstCol"));
     Assert.assertNull(row.getValue("secondCol"));
   }
+
+  @Test
+  public void testEscapedStrings() throws Exception {
+    List<Step> steps = new ArrayList<>();
+    Row row = new Row("__col", StringEscapeUtils.unescapeJava("1\\ta"));
```
abhinavbansal246, line 102: Should we escape once here and again at https://github.com/hydrator/wrangler-transform/pull/5/files#diff-01163bfca2584bef3a980fd885e184e1R101
Shouldn't we test with the replica of input as how it will be coming from FE

abhinavbansal246:
LGTM

src/test/java/co/cask/wrangler/steps/PipelineTest.java
russorat: So in cdap, the Wrangler Spec has the actual characters '\' and 't' in it, but the data coming through has the "tab" character in it. This test is mimicking that by unescaping the \\t into an actual "tab" character.

PR #8, opened 2017-01-08
Additional directives

nitinmotgi:
Following are the directives that have been added

[x] ```filter-row-by-regex <column> <regex>``` Filters row based on the regex applied on a column value. 
[x] ```filter-row-by-condition <condition>``` Filters row based on the condition. 
[x] ```mask-number <column> <mask-pattern>``` Substitution Masking 
[x] ```mask-shuffle <column>``` Shuffle Masking with same length as input
[x] ```format-date <column> <source-format> <destination-format>``` Formats date
[x] ```format-unixtimestamp <column> <destination-format>``` Formats unix timestamp
[x] ```set column <column-name> <jexl-expression>``` Executes the expression and applies it to the column.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,2 @@
+language: java
+mvn test
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,2 +1,29 @@
+#
+# Copyright © 2016 Cask Data, Inc.
+# 
+# Licensed under the Apache License, Version 2.0 (the "License"); you may not
+# use this file except in compliance with the License. You may obtain a copy of
+# the License at
+# 
+# http://www.apache.org/licenses/LICENSE-2.0
+# 
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations under
+# the License.
+
 language: java
-mvn test
+
+jdk:
+  - oraclejdk7
+  - oraclejdk8
+
+env:
+  - MAVEN_OPTS="-Xmx4096m -XX:MaxPermSize=128m" HADOOP_VERSION="2.4.1"
+
+sudo: false
+
+cache:
+  directories:
+    - $HOME/.m2
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
 #
-# Copyright © 2016 Cask Data, Inc.
+# Copyright © 2016-2017 Cask Data, Inc.
 # 
 # Licensed under the Apache License, Version 2.0 (the "License"); you may not
 # use this file except in compliance with the License. You may obtain a copy of
--- a/src/main/java/co/cask/wrangler/Wrangler.java
+++ b/src/main/java/co/cask/wrangler/Wrangler.java
@@ -27,6 +27,7 @@ import co.cask.cdap.etl.api.PipelineConfigurer;
 import co.cask.cdap.etl.api.Transform;
 import co.cask.cdap.etl.api.TransformContext;
 import co.cask.wrangler.api.Pipeline;
+import co.cask.wrangler.api.SkipRowException;
 import co.cask.wrangler.api.Specification;
 import co.cask.wrangler.internal.DefaultPipeline;
 import co.cask.wrangler.internal.TextSpecification;
@@ -127,10 +128,16 @@ public class Wrangler extends Transform<StructuredRecord, StructuredRecord> {
 
   @Override
   public void transform(StructuredRecord input, Emitter<StructuredRecord> emitter) throws Exception {
-    // Run through the wrangle pipeline.
-    StructuredRecord record = (StructuredRecord)pipeline.execute(input.get(config.field), oSchema);
-    StructuredRecord.Builder builder = StructuredRecord.builder(oSchema);
+    // Run through the wrangle pipeline, if there is a SkipRecord exception, don't proceed further
+    // but just return without emitting any record out.
+    StructuredRecord record;
+    try {
+      record = (StructuredRecord)pipeline.execute(input.get(config.field), oSchema);
+    } catch (SkipRowException e) {
+      return; // Skips the row.
+    }
 
+    StructuredRecord.Builder builder = StructuredRecord.builder(oSchema);
     // Iterate through output schema, if the 'record' doesn't have it, then
     // attempt to take if from 'input'.
     for (Schema.Field field : oSchema.getFields()) {
--- a/src/main/java/co/cask/wrangler/Wrangler.java
+++ b/src/main/java/co/cask/wrangler/Wrangler.java
@@ -31,8 +31,6 @@ import co.cask.wrangler.api.SkipRowException;
 import co.cask.wrangler.api.Specification;
 import co.cask.wrangler.internal.DefaultPipeline;
 import co.cask.wrangler.internal.TextSpecification;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 
 import java.io.IOException;
 import java.text.ParseException;
@@ -48,8 +46,6 @@ import java.text.ParseException;
 @Name("Wrangler")
 @Description("Wrangler - A interactive tool for data cleansing and transformation.")
 public class Wrangler extends Transform<StructuredRecord, StructuredRecord> {
-  private static final Logger LOG = LoggerFactory.getLogger(Wrangler.class);
-
   // Plugin configuration.
   private final Config config;
 
--- a/src/main/java/co/cask/wrangler/api/AbstractStep.java
+++ b/src/main/java/co/cask/wrangler/api/AbstractStep.java
@@ -30,7 +30,7 @@ public abstract class AbstractStep implements Step<Row, Row> {
 
   @Override
   public String toString() {
-    return String.format("[Step %d] - %s", lineno, detail);
+    return String.format("[Step %d] - <%s>", lineno, detail);
   }
 }
 
--- a/src/main/java/co/cask/wrangler/internal/TextSpecification.java
+++ b/src/main/java/co/cask/wrangler/internal/TextSpecification.java
@@ -1,5 +1,5 @@
 /*
- * Copyright © 2016 Cask Data, Inc.
+ * Copyright © 2016, 2017 Cask Data, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not
  * use this file except in compliance with the License. You may obtain a copy of
@@ -21,6 +21,7 @@ import co.cask.wrangler.api.Step;
 import co.cask.wrangler.steps.Columns;
 import co.cask.wrangler.steps.CsvParser;
 import co.cask.wrangler.steps.Drop;
+import co.cask.wrangler.steps.FormatDate;
 import co.cask.wrangler.steps.IndexSplit;
 import co.cask.wrangler.steps.Lower;
 import co.cask.wrangler.steps.Mask;
@@ -168,6 +169,15 @@ public class TextSpecification implements Specification {
         case "mask-shuffle":
           steps.add(new Mask(lineno, line, qualifier, null, Mask.MASK_SHUFFLE));
           break;
+
+        // format-date <column> <source-format> <destination-format>
+        case "format-date":
+          steps.add(new FormatDate(lineno, line, qualifier, options[2], options[3]));
+          break;
+
+        // format-unixtimestamp <column> <destination-format>
+        case "format-unixtimestamp":
+          steps.add(new FormatDate(lineno, line, qualifier, options[2]));
 
         default:
           throw new ParseException("Unknown command found in dsl", lineno);
--- a/src/main/java/co/cask/wrangler/internal/TextSpecification.java
+++ b/src/main/java/co/cask/wrangler/internal/TextSpecification.java
@@ -18,6 +18,7 @@ package co.cask.wrangler.internal;
 
 import co.cask.wrangler.api.Specification;
 import co.cask.wrangler.api.Step;
+import co.cask.wrangler.steps.Expression;
 import co.cask.wrangler.steps.Columns;
 import co.cask.wrangler.steps.CsvParser;
 import co.cask.wrangler.steps.Drop;
@@ -39,6 +40,7 @@ import java.text.ParseException;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
+import java.util.StringTokenizer;
 
 /**
  * Parses the DSL into specification containing steps for wrangling.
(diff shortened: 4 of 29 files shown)


.travis.yml
```diff
@@ -0,0 +1,29 @@
+#
+# Copyright © 2016 Cask Data, Inc.
```
abhinavbansal246: 2016-2017

src/main/java/co/cask/wrangler/api/AbstractStep.java
```diff
@@ -30,7 +30,7 @@ public AbstractStep(int lineno, String detail) {
 
   @Override
   public String toString() {
-    return String.format("[Step %d] - %s", lineno, detail);
+    return String.format("[Step %d] - <%s>", lineno, detail);
```
abhinavbansal246, line 33: why <>

src/main/java/co/cask/wrangler/Wrangler.java
```diff
@@ -127,10 +128,16 @@ public void initialize(TransformContext context) throws Exception {
...
-    // Run through the wrangle pipeline.
-    StructuredRecord record = (StructuredRecord)pipeline.execute(input.get(config.field), oSchema);
-    StructuredRecord.Builder builder = StructuredRecord.builder(oSchema);
+    // Run through the wrangle pipeline, if there is a SkipRecord exception, don't proceed further
+    // but just return without emitting any record out.
+    StructuredRecord record;
+    try {
+      record = (StructuredRecord)pipeline.execute(input.get(config.field), oSchema);
+    } catch (SkipRowException e) {
+      return; // Skips the row.
```
abhinavbansal246, line 135: We should log a debug message while skipping a row

src/main/java/co/cask/wrangler/steps/Mask.java
```diff
@@ -0,0 +1,164 @@
...
+    this.maskType = maskType;
+  }
+
+  /**
+   * Masks the column specified using either substitution method or shuffling.
+   *
+   * @param row Input {@link Row} to be wrangled by this step.
+   * @return A newly transformed {@link Row} with masked column.
+   * @throws StepException thrown when there is issue with masking
+   * @throws SkipRowException thrown when the row needs to be skipped
```
abhinavbansal246, line 93: says it throws skipRowException but never actually throws it

src/main/java/co/cask/wrangler/internal/TextSpecification.java
```diff
@@ -81,83 +86,158 @@ public TextSpecification(String dsl) {
...
                   if (unescapedStr == null) {
-                    throw new IllegalArgumentException("Invalid delimiter for CSV Parser: " + options[3]);
+                    throw new IllegalArgumentException("Invalid delimiter for CSV Parser: " + delimStr);
                   }
                   delimiter = unescapedStr.charAt(0);
                 }
+                boolean ignoreEmptyLines = false;
+                if (tokenizer.nextToken().equalsIgnoreCase("true")) {
+                  ignoreEmptyLines = true;
+                }
```
abhinavbansal246: can be combined to 
```
boolean ignoreEmptyLines = tokenizer.nextToken().equalsIgnoreCase("true");
```

src/main/java/co/cask/wrangler/internal/TextSpecification.java
```diff
@@ -81,83 +86,158 @@ public TextSpecification(String dsl) {
...
                 CsvParser.Options opt = new CsvParser.Options(delimiter, ignoreEmptyLines);
                 steps.add(new CsvParser(lineno, line, opt, STARTING_COLUMN, false));
                 steps.add(new Drop(lineno, line, STARTING_COLUMN));
               } else {
-                throw new ParseException("Unknown format " + options[3], lineno);
+                throw new ParseException("Unknown format ", lineno);
```
abhinavbansal246: It would be good to say in the error that what is the format it could not parse

src/main/java/co/cask/wrangler/internal/TextSpecification.java
```diff
@@ -81,83 +86,158 @@ public TextSpecification(String dsl) {
...
+              steps.add(new Expression(lineno, line, column, expr));
+            }
+            break;
 
             // set columns <name1, name2, ...>
-            case "columns":
-              String cols[] = options[2].split(",");
+            case "columns": {
+              String cols[] = tokenizer.nextToken().split(",");
               steps.add(new Columns(lineno, line, Arrays.asList(cols)));
```
abhinavbansal246: according to https://github.com/hydrator/wrangler-transform/blob/cd179d98e7d12154c07714fa5187f879eaf36718/src/main/java/co/cask/wrangler/steps/Columns.java#L61-L71

It will not stop from adding duplicate columns, is this the desired functionality? Are we checking at any other place if the duplicate columns are not being added to the row?

src/main/java/co/cask/wrangler/steps/Expression.java
```diff
@@ -0,0 +1,119 @@
...
+import org.slf4j.LoggerFactory;
+
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * A Wrangler step for apply expression results to a column.
+ */
+public class Expression extends AbstractStep {
+  private static final Logger LOG = LoggerFactory.getLogger(Columns.class);
```
abhinavbansal246: This doesnt seems to be used in this class, same for many other steps of wrangler

src/main/java/co/cask/wrangler/steps/Expression.java
```diff
@@ -0,0 +1,119 @@
...
+    try {
+      Object result = script.execute(context);
+      int idx = r.find(this.column);
+      if (idx == -1) {
+        r.add(this.column, result.toString());
+      } else {
+        r.setValue(idx, result.toString());
+      }
+    } catch (JexlException e) {
+      throw new StepException(toString() + " : '" + e.getMessage());
```
abhinavbansal246: Please put `e` as the parameter in the step exception being thrown.

src/main/java/co/cask/wrangler/steps/FormatDate.java
```diff
@@ -0,0 +1,91 @@
...
+import java.text.ParseException;
+import java.text.SimpleDateFormat;
+import java.util.Date;
+
+/**
+ * A Wrangle step for managing date formats.
+ *
+ */
+public class FormatDate extends AbstractStep {
+  private static final Logger LOG = LoggerFactory.getLogger(FormatDate.class);
```
abhinavbansal246: same here for LOG, not being used in the class

src/main/java/co/cask/wrangler/steps/FormatDate.java
```diff
@@ -0,0 +1,91 @@
...
+          date = new Date(Long.parseLong(datetimestamp) * 1000);
+        } else if (datetimestamp.matches("[1-9][0-9]{12}")) {
+          date = new Date(Long.parseLong(datetimestamp));
+        } else {
+          date = sourceFmt.parse((String) row.getValue(idx));
+        }
+        dt.setValue(idx, destinationFmt.format(date));
+      } catch (ParseException e) {
+        throw new StepException(toString() + " : '" +
+                                  column + ", " + e.getMessage());
```
abhinavbansal246: send e as a parameter to stepException.

src/main/java/co/cask/wrangler/steps/Mask.java
```diff
@@ -0,0 +1,164 @@
+/*
+ * Copyright © 2016 Cask Data, Inc.
```
abhinavbansal246: 2017

src/main/java/co/cask/wrangler/steps/Mask.java
```diff
@@ -0,0 +1,164 @@
...
+  @Override
+  public Row execute(Row row) throws StepException, SkipRowException {
+    Row masked = new Row(row);
+    int idx = row.find(column);
+    if (idx != -1) {
+      if (maskType == MASK_NUMBER) {
+        try {
+          masked.setValue(idx, maskNumber((String)row.getValue(idx), mask));
+        } catch (Exception e) {
+          System.out.println(e.getMessage());
```
abhinavbansal246, line 104: Log the exception and get rid of the SOUT statement

src/main/java/co/cask/wrangler/steps/Mask.java
```diff
@@ -0,0 +1,164 @@
...
+      }
+    } else {
+      throw new StepException(toString() + " : '" +
+                                column + "' column is not defined. Please check the wrangling step."
+      );
+    }
+    return masked;
+  }
+
+  private String maskNumber(String number, String mask) {
```
abhinavbansal246, line 117: Is this method or this functionality is only meant for numbers? If yes put a check If no change the method name

src/main/java/co/cask/wrangler/steps/Mask.java
```diff
@@ -0,0 +1,164 @@
...
+    int index = 0;
```
abhinavbansal246, line 118: We should check if the mask and the number are of same length, otherwise it can result in index out of bound exception. Or atleast that the length of the mask is less than the length of the number

src/main/java/co/cask/wrangler/steps/Mask.java
```diff
@@ -0,0 +1,164 @@
...
+      } else {
+        masked.append(c);
+      }
+    }
+    return masked.toString();
+  }
+
+  private String maskShuffle(String str, int seed) {
+    final String cons = "bcdfghjklmnpqrstvwxz";
+    final String vowel = "aeiouy";
```
abhinavbansal246, line 141: Why do we need to separate consonants and vowels? It is understandable in the case of digits and alphabets.

src/main/java/co/cask/wrangler/steps/RowRegexFilter.java
```diff
@@ -0,0 +1,66 @@
+/*
+ * Copyright © 2016 Cask Data, Inc.
```
abhinavbansal246: 2017

src/test/java/co/cask/wrangler/steps/PipelineTest.java
```diff
@@ -79,21 +80,61 @@ public void testSplit() throws Exception {
...
+    actual = (Row) step.execute(row);
+    Assert.assertEquals("x-8899", actual.getValue("ssn"));
+  }
+
+  @Test
+  public void testMaskSuffle() throws Exception {
+    Row row = new Row("address", "150 Mars Street, Mar City, MAR, 783735");
+    Step step = new Mask(0, "", "address", "", 2);
+    Row actual = (Row) step.execute(row);
+    Assert.assertEquals("089 Kyrp Czsyyr, Dyg Goci, FAG, 720322", actual.getValue("address"));
```
abhinavbansal246, line 127: Isnt't this test failing? How are you asserting something which is generated randomly?

src/main/java/co/cask/wrangler/steps/Mask.java
abhinavbansal246: "y" is not a vowel

src/test/java/co/cask/wrangler/steps/PipelineTest.java
```diff
@@ -117,4 +158,36 @@ public void testEscapedStrings() throws Exception {
...
+    }
+
+    Assert.assertEquals("Joltie, Root", row.getValue("name"));
+    Assert.assertEquals("1886.3999999999999", row.getValue("salary"));
+    Assert.assertEquals("no", row.getValue("isteen"));
+    Assert.assertEquals("oi.etiloj@toor", row.getValue("email"));
+    Assert.assertEquals("13.0", row.getValue("hrlywage"));
+  }
+
 }
```
abhinavbansal246, line 269: Please add some negative test cases

abhinavbansal246 (changes requested):

src/main/java/co/cask/wrangler/Wrangler.java
nitinmotgi: if there 10M, do we want 10M logs ?

src/main/java/co/cask/wrangler/steps/RowRegexFilter.java
nitinmotgi: done! It was written in 2016, so will change to 2016-2017.

src/test/java/co/cask/wrangler/steps/PipelineTest.java
nitinmotgi: Seed is the same! So when you have the same seed, the random sequence is the same. It's based on linear congruential random numbers.

.travis.yml
nitinmotgi: done!

src/main/java/co/cask/wrangler/api/AbstractStep.java
nitinmotgi: In case there are spaces, it shows up. Was useful in debugging.

src/main/java/co/cask/wrangler/steps/Mask.java
nitinmotgi: done!

src/main/java/co/cask/wrangler/steps/Mask.java
nitinmotgi: It's the contract, doesn't need to throw.

src/main/java/co/cask/wrangler/steps/Mask.java
nitinmotgi: Yeah like credit cards, SSN or things like that. What would you suggest I name this ? @abhinavbansal246

src/main/java/co/cask/wrangler/steps/Mask.java
nitinmotgi: Good point. Fixed.

src/main/java/co/cask/wrangler/steps/Mask.java
nitinmotgi: So that you consider from the same pool, depending on the length you might get the same string. So to avoid that, they are randomly chosen from different buckets.

src/main/java/co/cask/wrangler/internal/TextSpecification.java
nitinmotgi: fixed!

src/main/java/co/cask/wrangler/internal/TextSpecification.java
nitinmotgi: fixed.

src/main/java/co/cask/wrangler/internal/TextSpecification.java
nitinmotgi: ```replaceColumnNames``` is true, so it will replace it. That's only functionality that is needed. What's the use case you are thinking that would require merging ? @abhinavbansal246

src/main/java/co/cask/wrangler/steps/Expression.java
nitinmotgi: done!

src/main/java/co/cask/wrangler/steps/FormatDate.java
nitinmotgi: done!

src/main/java/co/cask/wrangler/Wrangler.java
abhinavbansal246: Makes sense.

src/main/java/co/cask/wrangler/steps/Mask.java
abhinavbansal246: Should be consistent with javadocs of execute methods in other steps e.g.
https://github.com/hydrator/wrangler-transform/blob/feature/additional-feature/src/main/java/co/cask/wrangler/steps/Merge.java#L57

src/main/java/co/cask/wrangler/steps/Mask.java
nitinmotgi: It's a contract, if it's not thrown then it's fine. There are other steps that might use this to skip rows.

nitinmotgi:
@abhinavbansal246 addressed all your comments. Please review again.

src/main/java/co/cask/wrangler/steps/Mask.java
abhinavbansal246: We can keep it as mask number if that is the use case here. the implementation in the method doesn't stop it to be used in non numerical fields e.g. we can use it to mask email ids, usernames or other PII information. If we encounter new use cases we can iterate in the future.

src/test/java/co/cask/wrangler/steps/PipelineTest.java
abhinavbansal246: ok

src/main/java/co/cask/wrangler/internal/TextSpecification.java
abhinavbansal246: I see, in the columns step class there is a constructor that takes `replaceColumnNames` as a parameter which is currently not being called fro any class yet. 

So unrelated to this code the use case I was thinking was that if some class initializes `Columns` with `replaceColumnNames` as false and try to add columns already existing in the row. There is no way to stop that.

src/main/java/co/cask/wrangler/steps/Expression.java
```diff
@@ -34,20 +34,35 @@
...
+ *
+ * <p>
+ *   Step step = new Expression(lineno, directive, column, "if (age > 24 ) { 'adult' } else { 'teen' }");
+ * </p>
  */
 public class Expression extends AbstractStep {
   private static final Logger LOG = LoggerFactory.getLogger(Columns.class);
 
-  // Columns of the column to be upper cased.
+  // Column to which the result of experience is applied to.
```
abhinavbansal246, line 47: expression

src/main/java/co/cask/wrangler/steps/RowConditionFilter.java
```diff
@@ -0,0 +1,92 @@
...
+        throw new SkipRowException();
+      }
+    } catch (JexlException e) {
+      // Generally JexlException wraps the original exception, so it's good idea
+      // to check if there is a inner exception, if there is wrap it in 'StepException'
+      // else just print the error message.
+      if (e.getCause() != null) {
+        throw new StepException(toString() + " : " + e.getMessage(), e.getCause());
+      } else {
+        throw new StepException(toString() + " : " + e.getMessage());
```
abhinavbansal246, line 82: We generally send e as a parameter while throwing new exception. that helps in debugging and provides stack trace.

src/main/java/co/cask/wrangler/steps/Mask.java
nitinmotgi: got it!

abhinavbansal246 (approved):

abhinavbansal246:
1 minor comment, otherwise LGTM

abhinavbansal246:
:+1:

src/main/java/co/cask/wrangler/steps/RowConditionFilter.java
nitinmotgi: If the internal cause is null, you see null. So it's of no use.

PR #12, opened 2017-01-14
Addition of Quantization Directive

nitinmotgi:
```
quantize <source-column-name> <dest-column-name>  [low:high=value][,low:high=value]*
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -209,6 +209,29 @@ available [here](http://commons.apache.org/proper/commons-jexl/reference/syntax.
   set column hrlywage Math:abs(toDouble(hrlywage))
 ```
 
+## Quantize
+This directive quantizes a continous value of a column through a range table
+specified. The quantization ranges are all real numbers, with low specifying the low end of the
+ range and high specifying the high end of the range. Associated with the range is the
+ value that if the incoming value falls in the range it would be assigned that value.
+ The range is a closed range - [low:high] = {x | low <= x <= high}. Also, the high endpoint
+ should be greater than low endpoint.
+
+**Specification**
+```
+  quantize {source-column} {destination-column} {quantization-table}
+```
+
+* source-column : Name of the column which has to be quantized
+* destination-column : Name of the column to which the quantized value should be added.
+* quantization-table : Specifies the quantization table in the following format low:high=value[,low:high=value]*
+the range specified in the quantization table is a closed range.
+
+**Example**
+```
+  quantize hrlywage wagecategory 0.0:4.99=LOW,5.0:13.99=NORMAL,14.0:29.99=HIGH,30.0:100.0=VERY HIGH
+```
+
 ### Mask Column
 Data masking (also known as data scrambling and data anonymization) is the process of replacing sensitive
 information with realistic, but scrubbed, data based on masking rules. This plugin supports two types of
--- a/README.md
+++ b/README.md
@@ -4,6 +4,8 @@
   <img alt="Coverity Scan Build Status"
        src="https://scan.coverity.com/projects/11434/badge.svg"/>
 </a>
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -208,6 +211,29 @@ available [here](http://commons.apache.org/proper/commons-jexl/reference/syntax.
   set column hrlywage Math:abs(toDouble(hrlywage))
 ```
 
+## Quantize
+This directive quantizes a continous value of a column through a range table
+specified. The quantization ranges are all real numbers, with low specifying the low end of the
+ range and high specifying the high end of the range. Associated with the range is the
+ value that if the incoming value falls in the range it would be assigned that value.
+ The range is a closed range - [low:high] = {x | low <= x <= high}. Also, the high endpoint
+ should be greater than low endpoint.
```
abhinavbansal246: all ranges are mutually exclusive.

README.md
```diff
@@ -208,6 +211,29 @@ available [here](http://commons.apache.org/proper/commons-jexl/reference/syntax.
...
+
+**Specification**
+```
+  quantize {source-column} {destination-column} {quantization-table}
+```
+
+* source-column : Name of the column which has to be quantized
+* destination-column : Name of the column to which the quantized value should be added.
+* quantization-table : Specifies the quantization table in the following format low:high=value[,low:high=value]*
```
abhinavbansal246, line 229: What does this mean `value[,low:high=value]*` ? Does this depicts that the user can have multiple ranges?

src/main/java/co/cask/wrangler/steps/Quantization.java
```diff
@@ -0,0 +1,87 @@
+/*
+ * Copyright © 2016 Cask Data, Inc.
```
abhinavbansal246: 2016-2017
Update your copyright profile intellij to have `$today.year` instead of 2016  so that it picks up current year.

src/main/java/co/cask/wrangler/steps/Quantization.java
```diff
@@ -0,0 +1,87 @@
...
+   */
+  @Override
+  public Row execute(Row row, PipelineContext context) throws StepException, SkipRowException {
+    Row r = new Row(row);
+    int idx = r.find(col1);
+
+    if (idx != -1) {
+      try {
+        double d = Double.parseDouble((String) r.getValue(idx));
+        String value = rangeMap.get(d);
```
abhinavbansal246, line 70: We should throw skiprow exception if value does not fall in any range

abhinavbansal246 (changes requested):

src/main/java/co/cask/wrangler/steps/Quantization.java
nitinmotgi: Column operations shouldn't skip the row right ? Why should we skip .. i want to have null value and I can operate on it later. If you look at the test case, it shows how it's handled.

README.md
nitinmotgi: Yes, the test case shows that.

README.md
nitinmotgi: good point.

src/main/java/co/cask/wrangler/steps/Quantization.java
nitinmotgi: it's 2017. done!

README.md
nitinmotgi: done!

nitinmotgi:
@abhinavbansal246 addressed your comments. Please approve to merge.

abhinavbansal246 (approved):

abhinavbansal246:
LGTM

PR #20, opened 2017-01-22
Handling of datetime type and parsing errors. 

nitinmotgi:
### List of 
- [x] Date format can now accept fields of type 'string', 'int' and 'long'.
- [x] Proper handling TextSpecification parsing (see below for type of error messages)

Missing field 'delimiter' at line number 1 for directive <set format> (usage: set format [csv|json] <delimiter> <skip empty lines>)
Missing field 'destination' at line number 1 for directive <rename> (usage: rename <source> <destination>)
Missing field 'column-name' at line number 1 for directive <drop> (usage: drop <column-name>)
Missing field 'col' at line number 1 for directive <uppercase> (usage: uppercase <col>)
Missing field 'col' at line number 1 for directive <titlecase> (usage: titlecase <col>)
Missing field 'source-column-name' at line number 1 for directive <indexsplit> (usage: indexsplit <source-column-name> <start> <end> <destination-column-name>)
Missing field 'start' at line number 1 for directive <indexsplit> (usage: indexsplit <source-column-name> <start> <end> <destination-column-name>)
Missing field 'end' at line number 1 for directive <indexsplit> (usage: indexsplit <source-column-name> <start> <end> <destination-column-name>)
Missing field 'destination-column-name' at line number 1 for directive <indexsplit> (usage: indexsplit <source-column-name> <start> <end> <destination-column-name>)
Missing field 'condition' at line number 1 for directive <filter-row-if-true> (usage: filter-row-if-true <condition>)
Missing field 'destination-format' at line number 1 for directive <format-unix-timestamp> (usage: format-unix-timestamp <column> <destination-format>)

### DIFF
--- a/src/main/java/co/cask/wrangler/Wrangler.java
+++ b/src/main/java/co/cask/wrangler/Wrangler.java
@@ -29,15 +29,16 @@ import co.cask.cdap.etl.api.Transform;
 import co.cask.cdap.etl.api.TransformContext;
 import co.cask.wrangler.api.Pipeline;
 import co.cask.wrangler.api.PipelineContext;
+import co.cask.wrangler.api.PipelineException;
 import co.cask.wrangler.api.Row;
 import co.cask.wrangler.api.SkipRowException;
 import co.cask.wrangler.api.Specification;
+import co.cask.wrangler.api.SpecificationParseException;
 import co.cask.wrangler.internal.DefaultPipeline;
 import co.cask.wrangler.internal.TextSpecification;
 import com.google.common.base.Strings;
 
 import java.io.IOException;
-import java.text.ParseException;
 
 /**
  * Wrangler - A interactive tool for data data cleansing and transformation.
@@ -58,6 +59,9 @@ public class Wrangler extends Transform<StructuredRecord, StructuredRecord> {
 
   // Output Schema associated with transform output.
   private Schema oSchema;
+
+  // Error counter.
+  private long errorCounter;
 
   // This is used only for tests, otherwise this is being injected by the ingestion framework.
   public Wrangler(Config config) {
@@ -90,7 +94,7 @@ public class Wrangler extends Transform<StructuredRecord, StructuredRecord> {
     Specification specification = new TextSpecification(config.specification);
     try {
       specification.getSteps();
-    } catch (ParseException e) {
+    } catch (SpecificationParseException e) {
       throw new IllegalArgumentException(e);
     }
 
@@ -139,6 +143,9 @@ public class Wrangler extends Transform<StructuredRecord, StructuredRecord> {
     } catch (IOException e) {
       throw new IllegalArgumentException("Format of output schema specified is invalid. Please check the format.");
     }
+
+    // Initialize the error counter.
+    errorCounter = 0;
   }
 
   @Override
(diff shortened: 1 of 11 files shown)


src/main/java/co/cask/wrangler/Wrangler.java
```diff
@@ -59,6 +60,9 @@
   // Output Schema associated with transform output.
   private Schema oSchema;
 
+  // Error counter.
+  private long errorCounter;
```
abhinavbansal246, line 64: Should it be a static variable?

src/main/java/co/cask/wrangler/internal/TextSpecification.java
```diff
@@ -66,8 +69,33 @@
...
   private String[] directives;
 
+  // Mapping of specification formats.
+  Map<String, String> formats = new HashMap<>();
+
   public TextSpecification(String[] directives) {
     this.directives = directives;
+
+    // Add all the usages.
+    formats.put("set format", "set format [csv|json] <delimiter> <skip empty lines>");
```
russorat, line 79: Is there a way to add a method to each directive to define this so that when someone adds a new directive, they don't have to make updates in multiple places?

src/main/java/co/cask/wrangler/internal/TextSpecification.java
```diff
@@ -138,136 +170,169 @@ public TextSpecification(String directives) {
...
+                          String directive, String field, int lineno) throws SpecificationParseException {
+    String value;
+    if (tokenizer.hasMoreTokens()) {
+      if (delimiter == null) {
+        value = tokenizer.nextToken();
+      } else {
+        value = tokenizer.nextToken(delimiter);
+      }
+    } else {
+      String d = formats.get(directive);
```
russorat, line 321: then this could delegate to the specific directive class to get the usage string?

src/test/java/co/cask/wrangler/internal/TextSpecificationTest.java
```diff
@@ -53,4 +54,38 @@ public void testBasicSpecification() throws Exception {
...
+      try {
+        List<Step> steps = specification.getSteps();
+      } catch (SpecificationParseException e) {
+        errorcount++;
+      }
+    }
+    Assert.assertEquals(12, errorcount);
+  }
+
 }
```
abhinavbansal246: add new line

abhinavbansal246 (changes requested):

abhinavbansal246:
very minor comments otherwise
LGTM

src/main/java/co/cask/wrangler/internal/TextSpecification.java
nitinmotgi: @russorat don't have a great idea right now on how to make it easy. Can I leave it for now and address it little latter ?

src/main/java/co/cask/wrangler/internal/TextSpecification.java
nitinmotgi: Totally get what you are saying, just need to think a bit more.

src/main/java/co/cask/wrangler/Wrangler.java
nitinmotgi: if I make it static them it needs to be AtomicLong -- do we need to make that. Remember multiple instance could be running in the same JVM.

PR #21, opened 2017-01-25
Sed directive implementation

nitinmotgi:
### Examples

```
  sed body s/"//g
  sed body s/\r\n//g
```


src/main/java/co/cask/wrangler/steps/Sed.java
```diff
@@ -0,0 +1,70 @@
...
+   * Sets the new column names for the {@link Row}.
+   *
+   * @param row Input {@link Row} to be wrangled by this step.
+   * @param context Specifies the context of the pipeline.
+   * @return A newly transformed {@link Row}.
+   * @throws StepException
+   */
+  @Override
+  public Row execute(Row row, PipelineContext context) throws StepException, SkipRowException {
+    Row r = new Row(row);
```
abhinavbansal246: Why cant we just modify the existing `row` object? Do we need to preserve the original row object?

src/main/java/co/cask/wrangler/steps/Sed.java
```diff
@@ -0,0 +1,70 @@
...
+    int idx = row.find(column);
+    if (idx != -1) {
+      Object v = row.getValue(idx);
+      // Operates only on String types.
+      if (v instanceof String) {
+        String value = (String) v; // Safely converts to String.
+        Unix4jCommandBuilder builder = Unix4j.echo(value).sed(pattern);
```
abhinavbansal246: what happens if sed throws exception on a particular string. Should we throw skipRow exception?

src/main/java/co/cask/wrangler/steps/Sed.java
nitinmotgi: fixed

src/main/java/co/cask/wrangler/Wrangler.java
```diff
@@ -86,7 +100,7 @@ public void configurePipeline(PipelineConfigurer configurer) throws IllegalArgum
 
     Schema iSchema = configurer.getStageConfigurer().getInputSchema();
 
-    if (!config.field.equalsIgnoreCase("all")) {
+    if (!config.field.equalsIgnoreCase("*")) {
```
abhinavbansal246, line 103: UI was having trouble in sending `*` in tracker search api. We should confirm that this will work from their end.

abhinavbansal246:
1 small question otherwise LGTM

src/main/java/co/cask/wrangler/Wrangler.java
nitinmotgi: It's no more a drop down, so it's fine. Merging now.

PR #22, opened 2017-01-26
Adding multiple parsers

nitinmotgi:
- [x] Parse JSON
- [x] Parse CSV
- [x] Parse Fixed Length
- [x] Json Path

### DIFF
--- a/docs/Wrangler-transform.md
+++ b/docs/Wrangler-transform.md
@@ -11,8 +11,77 @@ the order they are specified.
 ## Types of Directives
 Following are different types of directives that are supported by the Wrangler plugin.
 
-### Parser
+### Parsers
 
+#### CSV Parser
+
+Parses a column a comma separated value (CSV).
+
+**Specification**
+```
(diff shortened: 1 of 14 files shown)


docs/Wrangler-transform.md
```diff
@@ -11,8 +11,77 @@ the order they are specified.
...
+  drop body.deviceReference,
+  rename body.deviceReference.timestamp timestamp,
+  set column timestamp timestamp / 1000000,
+  drop body.deviceReference.alerts,
+  set columns timestamp,alerts,phone,battery,brand,type,comments,deviceId,os_name,os_version,size1,size2,size3,size4,signal
+```
+
+#### Fixed Length Parser
+
+Parses a column as fixed length record with range specifications specified.
```
russorat, line 72: is this inclusive for start and exclusive for end? Looks like it's inclusive for both and starts at 1 instead of zero.  We should mention that in the docs.

russorat:
LGTM

docs/Wrangler-transform.md
nitinmotgi: addressed.

PR #28, opened 2017-02-02
Fixes error message and tests in Expression directive

nitinmotgi:


russorat:
LGTM

wrangler-core/src/test/java/co/cask/wrangler/steps/ExpressionTest.java
```diff
@@ -0,0 +1,126 @@
...
+    );
+
+    records = PipelineTest.execute(directives, records);
+
+    Assert.assertTrue(records.size() == 1);
+    Assert.assertEquals("root", records.get(0).getValue("first"));
+    Assert.assertEquals(99, records.get(0).getValue("number"));
+  }
+
+}
```
sreevatsanraman: Add a new line to the end of the file - otherwise will break checkstyle.

wrangler-core/src/main/java/co/cask/wrangler/steps/Expression.java
```diff
@@ -95,6 +97,8 @@ public Expression(int lineno, String detail, String column, String expression) {
     functions.put(null, Convertors.class);
     functions.put("math", Math.class);
     functions.put("string", StringUtils.class);
+    functions.put("bytes", Bytes.class);
+    functions.put("arrays", Arrays.class);
```
sreevatsanraman, line 101: add test case for arrays.

sreevatsanraman:
Couple minor comments, fix and merge.

wrangler-core/src/main/java/co/cask/wrangler/steps/Expression.java
nitinmotgi: The idea is to test for directives to access the functionality. Don't need to test each functions within arrays.

wrangler-core/src/test/java/co/cask/wrangler/steps/ExpressionTest.java
nitinmotgi: done!

PR #30, opened 2017-02-04
WRITE-TO-JSON Directive and Usage Annotation

nitinmotgi:
This PR has two parts

* WRITE-TO-JSON directive &
* Cleanup of Usage for Text Directives through Annotations. 

For Usage clean up, we have added a annotation called ```Usage``` that would be used when building new steps as follows:

```
@Usage(directive = "write-to-xml", usage = "write-to-xml <column>")
public class WriteToXML extends AbstractStep {
  ...
}
```


docs/directives/write-to-json.md
```diff
@@ -0,0 +1,43 @@
+# Write to JSON
+
+WRITE-TO-JSON directive converts the record into JSON.
+
+## Syntax
+```
+  write-to-json <column>
```
russorat: this in my opinion is not write-to-json, this is write-to-json-map or something. write-to-json on the example below should create a body = "{\"int\":1,\"string\":\"this is a string\"}"

docs/directives/write-to-json.md
```diff
@@ -0,0 +1,43 @@
...
+```
+
+```column``` will contain the JSON of the fields in the record.
+
+## Usage Notes
+
+The WRITE-TO-JSON directive provides an easy way to convert the record
+into a JSON. If the ```column``` already exists, it will override it.
```
russorat: override or overwrite

wrangler-core/src/main/java/co/cask/wrangler/steps/MaskNumber.java
```diff
@@ -0,0 +1,129 @@
...
+import co.cask.wrangler.api.PipelineContext;
+import co.cask.wrangler.api.Record;
+import co.cask.wrangler.api.StepException;
+import co.cask.wrangler.api.Usage;
+
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * A Wrangler plugin that applies substitution and shuffling masking on the column.
```
russorat: i think this should say shuffle masking

wrangler-core/src/main/java/co/cask/wrangler/steps/MaskNumber.java
```diff
@@ -0,0 +1,129 @@
...
+ *
+ * <p>
+ *  Substitution masking us generally used for masking credit card or SSN numbers.
+ *  This type of masking is fixed masking, where the pattern is applied on the
+ *  fixed length string.
+ *
+ *  <ul>
+ *    <li>Use of # will include the digit from the position.</li>
+ *    <li>Use x to mask the digit at that position</li>
+ *    <li>Any other characters will be inserted as-is.</li>
```
russorat, line 39: how do you include the character "x" in the final result?

wrangler-core/src/main/java/co/cask/wrangler/steps/ParseLog.java
```diff
@@ -28,6 +29,7 @@
 /**
  * A Step for parsing Apache HTTPD and NGINX log files.
```
russorat: Since you can specify the format, this could be applied to any log files correct?

wrangler-core/src/main/java/co/cask/wrangler/steps/RecordRegexFilter.java
```diff
@@ -28,6 +29,7 @@
 /**
  * A Wrangle step for filtering rows that match the pattern specified on the column.
  */
+@Usage(directive = "filter-row-if-matched", usage = "filter-row-if-matched <column> <regex>")
```
russorat: is there a filter-column-if-matched? if not, can these just be filter-if-matched?

PR #39, opened 2017-02-16
Adding columns-format-case directive

russorat:


docs/directives/columns-format-case.md
```diff
@@ -0,0 +1,53 @@
...
+
+COLUMNS-FORMAT-CASE directive provides the ability to clean up the column names in bulk.
+
+## Syntax
+
+```
+ columns-format-case current-format desired-format
+```
+
+```current-format``` or ```desired-format``` specifies the current or desired format and can accept any of the following options:
```
nitinmotgi, line 11: Why do we need to specify current-format ? Why not just tell what you need it to be. I found it confusing to understand what my current-format is.

wrangler-core/src/main/java/co/cask/wrangler/internal/TextDirectives.java
```diff
@@ -696,6 +698,14 @@ public TextDirectives(String directives) {
         }
         break;
 
+        // columns-format-case <current-case> <desired-case>
+        case "columns-format-case" : {
+          String currentCase = getNextToken(tokenizer, command, "current-case", lineno);
+          String desiredCase = getNextToken(tokenizer, command, "desired-case", lineno);
```
nitinmotgi, line 704: What happens if I mis-type the desired-case. There is no validation on the supported types. Could we add check for that and in case it doesn't match the things that we want it to be we throw a ```DirectiveParseException```

wrangler-core/src/test/java/co/cask/wrangler/steps/column/ColumnsFormatCaseTest.java
```diff
@@ -0,0 +1,49 @@
...
+
+import java.util.Arrays;
+import java.util.List;
+
+/**
+ * Tests {@link ColumnsFormatCaseTest}
+ */
+public class ColumnsFormatCaseTest {
+
+  @Test
```
nitinmotgi, line 32: Add a test case that tests negative case.

nitinmotgi (changes requested):
Please take one more pass over the implementation and the idea.

russorat:
closing. can be accomplished using:
```
columns-replace s/([A-Z])/_$1/g
cleanse-column-names
```

PR #45, opened 2017-02-28
Fix json generation script

rohitsinha54:
same as https://github.com/hydrator/naive-bayes-analytics/pull/5

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -85,7 +85,7 @@
     <!-- properties for script build step that creates the config files for the artifacts -->
     <widgets.dir>widgets</widgets.dir>
     <docs.dir>docs</docs.dir>
-    <etl.versionRange>[4.0.0,10.0.0-SNAPSHOT)</etl.versionRange>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -85,7 +85,7 @@
     <!-- properties for script build step that creates the config files for the artifacts -->
     <widgets.dir>widgets</widgets.dir>
     <docs.dir>docs</docs.dir>
-    <etl.versionRange>[4.0.0,10.0.0-SNAPSHOT)</etl.versionRange>
+    <app.parents>system:cdap-etl-batch[4.1.0,4.2.0-SNAPSHOT),system:cdap-etl-realtime[4.1.0,4.2.0-SNAPSHOT),system:cdap-data-pipeline[4.1.0,4.2.0-SNAPSHOT),system:cdap-data-streams[4.1.0,4.2.0-SNAPSHOT)</app.parents>
```
nitinmotgi: Why do we need batch and realtime when we have data pipeline and streams ?

nitinmotgi (changes requested):

pom.xml
rohitsinha54: Not entirely sure. It depend on it in older script version:
https://github.com/hydrator/wrangler/pull/45/files#diff-600376dffeb79835ede4a0b285078036L234

@albertshau Might be able to clarify if its needed or not.

albertshau: its still needed if we want it to be available with those deprecated apps. Whether it should be there I guess is a product decision?

bdmogal: Unless somethimg breaks (I know there was some plugin function that required this) without this, lets not add it

PR #48, opened 2017-03-01
Applying filtering before directives are applied

nitinmotgi:
This feature adds a pre-condition expression based filter that can be applied. 

precondition

screen shot 2017-03-01 at 12 09 45 am


transform/src/main/java/co/cask/wrangler/PreconditionException.java
```diff
@@ -0,0 +1,10 @@
+package co.cask.wrangler;
+
+/**
+ * Thrown when there are issues with pre-conditions.
+ */
+public class PreconditionException extends Exception {
```
bdmogal, line 22: if this extends IllegalArgumentException instead, will it simplify exception handling?

transform/src/test/java/co.cask.wrangler/PreconditionTest.java
```diff
@@ -0,0 +1,26 @@
+package co.cask.wrangler;
```
bdmogal, line 17: license

transform/src/main/java/co/cask/wrangler/PreconditionException.java
```diff
@@ -0,0 +1,10 @@
+package co.cask.wrangler;
```
bdmogal, line 17: license

bdmogal (commented):
1 question abt exception  handling. rest lgtm

transform/src/main/java/co/cask/wrangler/PreconditionException.java
nitinmotgi: No runtime exception **EVER** also I need that during evaluation too.

transform/src/main/java/co/cask/wrangler/PreconditionException.java
nitinmotgi: done!

transform/src/test/java/co.cask.wrangler/PreconditionTest.java
nitinmotgi: done!

anwar6953 (approved):

transform/src/test/java/co.cask.wrangler/PreconditionTest.java
```diff
@@ -0,0 +1,42 @@
...
+    Assert.assertEquals(true, new Precondition("true").apply(record));
+    Assert.assertEquals(false, new Precondition("false").apply(record));
+  }
+
+  @Test(expected = PreconditionException.class)
+  public void testBadCondition() throws Exception {
+    Record record = new Record("a", 1).add("b", "x").add("c", 2.06);
+    Assert.assertEquals(true, new Precondition("c").apply(record));
+  }
+}
```
anwar6953: nit - missing newline at EOF

nitinmotgi: addressed

PR #50, opened 2017-03-02
Fixing delimiter -- during upload

nitinmotgi:


service/src/main/java/co/cask/wrangler/service/WranglerService.java
```diff
@@ -144,13 +144,17 @@ public void upload(HttpServiceRequest request, HttpServiceResponder responder,
...
-      delimiter = "\\u001A";
-    }
-    delimiter = StringEscapeUtils.unescapeJava(delimiter);
-    for (String line : body.split(delimiter)) {
-      records.add(new Record(ws, line));
-      ++i;
+    if(delimiter == null ) {
+      records.add(new Record(ws, content));
+    } else {
+      if (delimiter.isEmpty()) {
```
anwar6953, line 150: Its not clear to me why the behavior should be different if delimiter is `null`, versus when it is empty.
Doesn't the ui always send a non-null value?

nitinmotgi: Fair, fixing it.

PR #58, opened 2017-03-14
Update the script for child pom too

rohitsinha54:
In this PR https://github.com/hydrator/wrangler/pull/45 we updated the script and the property to app.parent. 
Changing this child pom script to the same.


anwar6953:
LGTM

PR #63, opened 2017-03-18
WRANGLER-84 Implement UnreachableOptimizer

Denton-L:
* Upgrade pom files
  * This upgrades Guava to 20.0. In addition, JUnit is explicitly included.
Finally, dependencies are reordered to stop JAR hell from happening.

* Implement UnreachableOptimizer
  * This optimiser works by removing any directives that are not used to
produce the final outputs.


nitinmotgi:
The ```AbstractStep``` should not change, but each step can delegate information. We want to have only one interface for people to extend. This implementation creates a lot of interface. So it's not user friendly. Think of an alternative design and please document and discuss design before implementation. Thanks.

nitinmotgi (changes requested):
This is doing major re-writing without thinking about user in mind. So, please take a step back / design and discuss this before second attempt for implementation.

Denton-L:
I could quickly rewrite this such that the end user only needs to extend the `AbstractStepClass` in such a way that it would work as it did previously. The only caveat for this is that the user-implemented step would not be optimised but this should be fine because it would be impossible to optimise without providing any sort of metadata, anyway.

nitinmotgi:
@Denton-L there is a way. Could we discuss this on OSS thread or f-2-f please.

Denton-L:
@nitinmotgi, let's discuss this whenever you're free to.

PR #64, opened 2017-03-21
Better error message.

anwar6953:
Better error message in case there's a type of object in Record that isn't supported.


abhinavbansal246:
LGTM

PR #65, opened 2017-03-21
Proper error message when schema is complex

nitinmotgi:


core/src/main/java/co/cask/wrangler/internal/RecordConvertor.java
```diff
@@ -136,9 +136,9 @@ public Schema toSchema(String id, List<Record> records) throws RecordConvertorEx
           Schema arraySchema = null;
           try {
             arraySchema = generateJSONArraySchema(array);
-          } catch (UnsupportedTypeException e) {
+          } catch (UnsupportedOperationException | UnsupportedTypeException e) {
```
anwar6953: Instead of handling UnsupportedOperationException, isn't it better to throw the appropriate exception in the first place?
https://github.com/hydrator/wrangler/pull/64/files

service/src/main/java/co/cask/wrangler/service/WranglerService.java
```diff
@@ -333,6 +333,10 @@ public void schema(HttpServiceRequest request, HttpServiceResponder responder,
       } catch (RecordConvertorException e) {
         error(responder, "There was a problem in generating schema for the record. " + e.getMessage());
         return;
+      } catch (UnsupportedOperationException e) {
```
anwar6953, line 336: I dont think we need to handle this anymore, right?

anwar6953:
Reopening; see the PR for details:
https://github.com/hydrator/wrangler/pull/66

PR #66, opened 2017-03-22
Feature/complex schema generation

anwar6953:
Same PR as https://github.com/hydrator/wrangler/pull/65, but against release/1.2 branch, instead of develop.
Also added a commit to `Remove handling of UnsupportedOperationException.`.
Also added a commit to `Bump version to 1.2.1 from 1.2.0`.


bdmogal:
LGTM

PR #77, opened 2017-04-17
Revisions

awholegunch:
Revisions to README, license, and pom.xml.


README.md
```diff
@@ -1,20 +1,23 @@
...
+![cm-available](https://cdap-users.herokuapp.com/assets/cm-available.svg)
+![cdap-transform](https://cdap-users.herokuapp.com/assets/cdap-transform.svg)
+[![Join CDAP community](https://cdap-users.herokuapp.com/badge.svg?t=wrangler)](https://cdap-users.herokuapp.com)
+[![Build Status](https://travis-ci.org/hydrator/wrangler.svg?branch=develop)](https://travis-ci.org/hydrator/wrangler)
+[![Coverity Scan Build Status](https://scan.coverity.com/projects/11434/badge.svg)](https://scan.coverity.com/projects/hydrator-wrangler-transform)
 [![License](https://img.shields.io/badge/License-Apache%202.0-blue.svg)](https://opensource.org/licenses/Apache-2.0)
 
-Collection of libraries, pipeline plugin and CDAP service for performing data cleansing, transformation and filtering using a set of instructions. Instructions to manipulate data are either generated using an interative visual tool or could be manually entered.
+A collection of libraries, a pipeline plugin, and a CDAP service for performing data
+cleansing, transformation, and filtering using a set of data manipulation instructions.
```
russorat: do you want to call them "directives" instead of "data manipulation instructions"? Or maybe put directives in () afterwards.

README.md
```diff
@@ -23,71 +26,81 @@ Also, check the documentation of Wrangler Transform [here](transform/docs/Wrangl
...
 
-A Column is a data value of any supported java type, one for each Record.
+A *Column* is a data value of any of the supported Java types, one for each record.
 
 ### Directive
 
-A Directive is a single data manipulation instruction specified to either transform, filter or pivot a single record into zero or more records. A directive can generate one or more Steps to be executed by the Pipeline. 
+A *Directive* is a single data manipulation instruction, specified to either transform,
+filter, or pivot a single record into zero or more records. A directive can generate one
+or more *steps* to be executed by a Pipeline.
```
russorat, line 49: I think there are also directives that don't do anything in a pipeline. I don't know if they are in this release or not. @nitinmotgi ?

README.md
```diff
@@ -23,71 +26,81 @@ Also, check the documentation of Wrangler Transform [here](transform/docs/Wrangl
...
-A Pipeline is a collection of Steps to be applied on a Record. Record(s) outputed from each Step is passed to the next Step in the pipeline. 
+A *Pipeline* is a collection of steps to be applied on a record. The record(s) outputed
+from a step are passed to the next step in the pipeline.
 
 ## Notations
 
 ### Directives
 
-A directive is represented as simple text in the format as specified below
+A directive is represented by simple text in this format:
```
russorat: This is technically a TextDirective. For example, If a user wants to create a JSONDirective, it could be specified in the format { "action":"something","field":"else"}

README.md
```diff
@@ -146,101 +159,112 @@ Following are different directives currently available.
...
+| Medium | 18 | 13,499,973 | 4,499,534,313 | 64,998.50 | 64,921.29 | 46,866.70 | 36,149.86 |
+| Medium | 18 | 80,999,838 | 26,997,205,878 | 62,465.93 | 62,706.39 | 60,755.41 | 56,673.32 |
+
 
 ## Wrangler Service
 
-Wrangler is integrated as a CDAP Service to support REST based interactive ways to wrangle data. The main objective of having this service is to make it easy for interactively generating directives required for parsing data. This service does not support full scale big data processing.
+Wrangler is integrated as a CDAP Service to support HTTP RESTful-based interactive
+wrangling of data. The main objective of this service is to make it simple and easy to
+interactively generate the directives required for parsing a given data set. This service
```
russorat: technically, the service doesn't generate the directives. It interactively **applies** the directives.

README.md
```diff
@@ -146,101 +159,112 @@ Following are different directives currently available.
...
+does not support full-scale big data processing.
```
russorat: Is it better to say something like: This service is not intended to replace full-scale big data processing. It is primarily used to interactively apply directives on a sample of your data.

It's strange to say it doesn't support it, since that's not what it is used for. like saying a spoon doesn't support cutting steak :)

README.md
```diff
@@ -146,101 +159,112 @@ Following are different directives currently available.
...
 
-The base endpoint is defined below :
+The base endpoint is:
 
 ```
-  http://<hostname>:11015/v3/namespaces/<namespace>/apps/wrangler/services/service/methods
+http://<hostname>:11015/v3/namespaces/<namespace>/apps/wrangler/services/service/methods
```
russorat: I think the application name has been changed to dataprep

README.md
```diff
@@ -146,101 +159,112 @@ Following are different directives currently available.
...
+* [cdap-user@googlegroups.com](https://groups.google.com/d/forum/cdap-user)
 
 The *cdap-user* mailing list is primarily for users using the product to develop
-applications or building plugins for appplications. You can expect questions from 
-users, release announcements, and any other discussions that we think will be helpful 
+applications or building plugins for appplications. You can expect questions from
+users, release announcements, and any other discussions that we think will be helpful
 to the users.
 
 ## IRC Channel
```
russorat: Should we add anything about a slack channel here?

russorat:
Reviewed. minor changes and suggestions.

README.md
awholegunch: Are there other directive types? Is there a list or source for the types? I don't see `JSONDirective` in the code anywhere.

README.md
awholegunch: In parentheses, afterwards.

README.md
awholegunch: Fixed.

README.md
awholegunch: That's fine. What I don't understand (at this point) is how this is intended to be used in practice. How does one generate the directives, how do you move from a sample of your data to using what you have written in a production setting?

README.md
awholegunch: Yes, or maybe replace the IRC with the Slack channel? @nitinmotgi , what do you think? Both, or  replace IRC with the Slack channel? And do this everywhere?

README.md
awholegunch: Hmm. Does *wrangler* disappear then?

README.md
russorat: It isn't in there today, but technically, a directive can be represented by anything. What we have implemented is a TextDirective which is what is represented by simple text in this format. I think it would be more correct to say "We have implemented a set of directives that are represented using simple text in this format:"

README.md
russorat: Both the service and the transform use the same common library for applying directives to data: wrangler-core.jar. The service is just a tool for users to interactively apply directives to sample data and see what it does to the data. Since the directives are interpreted by a common library, when the user executes those same directives in the transform plugin on millions of rows (we usually call this "operationalizing" them), those directives should be executed the exact same way. So, you are correct in saying the "service does not support full scale big data processing", but the library that is common to both the service and the transform supports it.

README.md
russorat: yeah, wrangler = dataprep

README.md
awholegunch: Added a Slack Team. Changed the IRC Channel to include a link to the webchat.

russorat (approved):
LGTM

PR #88, opened 2017-05-03
CDAP-11391 Upload doesn't require workspace as pre-requisite. If the …

nitinmotgi:
…workspace doesn't exist, then API will automatically create it

### DIFF
--- a/service/src/main/java/co/cask/wrangler/dataset/workspace/WorkspaceDataset.java
+++ b/service/src/main/java/co/cask/wrangler/dataset/workspace/WorkspaceDataset.java
@@ -143,8 +143,8 @@ public class WorkspaceDataset extends AbstractDataset {
       return false;
     } catch (DataSetException e) {
       throw new WorkspaceException(
-        String.format("Unable to check if workspace exists '%s'",
-                      e.getMessage())
+        String.format("Unable to check if workspace '%s' exists '%s'",
+                      id, e.getMessage())
       );
     }
   }
--- a/service/src/main/java/co/cask/wrangler/dataset/workspace/WorkspaceDataset.java
+++ b/service/src/main/java/co/cask/wrangler/dataset/workspace/WorkspaceDataset.java
@@ -137,7 +137,7 @@ public class WorkspaceDataset extends AbstractDataset {
   public boolean hasWorkspace(String id) throws WorkspaceException {
     try {
       Row row = table.get(toKey(id));
-      if (row != null) {
+      if (!row.isEmpty()) {
         return true;
       }
       return false;
--- a/service/src/main/java/co/cask/wrangler/dataset/workspace/WorkspaceDataset.java
+++ b/service/src/main/java/co/cask/wrangler/dataset/workspace/WorkspaceDataset.java
@@ -128,6 +128,28 @@ public class WorkspaceDataset extends AbstractDataset {
   }
 
   /**
+   * Checks if a workspace exists.
+   *
+   * @param id of the workspace to be checked for.
+   * @return true if workspace exists, false otherwise.
+   * @throws WorkspaceException thrown if there are any issues with workspace.
+   */
+  public boolean hasWorkspace(String id) throws WorkspaceException {
+    try {
+      Row row = table.get(toKey(id));
+      if (row != null) {
+        return true;
+      }
+      return false;
+    } catch (DataSetException e) {
+      throw new WorkspaceException(
+        String.format("Unable to check if workspace exists '%s'",
+                      e.getMessage())
+      );
+    }
+  }
+
+  /**
    * Lists all the workspaces registered.
    *
    * @return List of workspaces.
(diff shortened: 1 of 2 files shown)


service/src/main/java/co/cask/wrangler/dataset/workspace/WorkspaceDataset.java
```diff
@@ -128,6 +128,28 @@ public void createWorkspaceMeta(String id, String name, DataType type,
...
+   * Checks if a workspace exists.
+   *
+   * @param id of the workspace to be checked for.
+   * @return true if workspace exists, false otherwise.
+   * @throws WorkspaceException thrown if there are any issues with workspace.
+   */
+  public boolean hasWorkspace(String id) throws WorkspaceException {
+    try {
+      Row row = table.get(toKey(id));
+      if (row != null) {
```
russorat: table.get never returns null. I think you need to check isEmpty() or something.

nitinmotgi: Addressed comment -- not sure why it worked. But using isEmpty is right way to do it.

service/src/main/java/co/cask/wrangler/dataset/workspace/WorkspaceDataset.java
```diff
@@ -128,6 +128,28 @@ public void createWorkspaceMeta(String id, String name, DataType type,
...
+  public boolean hasWorkspace(String id) throws WorkspaceException {
+    try {
+      Row row = table.get(toKey(id));
+      if (!row.isEmpty()) {
+        return true;
+      }
+      return false;
+    } catch (DataSetException e) {
+      throw new WorkspaceException(
+        String.format("Unable to check if workspace exists '%s'",
```
awholegunch: What about using:
```
String.format("Unable to check if workspace '%s' exists '%s'", id,

File: service/src/main/java/co/cask/wrangler/service/directive/DirectivesService.java [hunk 3]
```diff
@@ -265,6 +267,106 @@ public void get(HttpServiceRequest request, HttpServiceResponder responder,
...
+        }
+
+        case BINARY: {
+          table.writeToWorkspace(id, WorkspaceDataset.DATA_COL, DataType.BINARY, content);
+          break;
+        }
+
+        default: {
+          error(responder, "Invalid content type. Supports text/plain, application/octet-stream " +
+            "and application/data-prep");
```
awholegunch, line 347: What about:
```
          error(responder, "Invalid content type. Supports text/plain, application/octet-stream, " +
             "or application/data-prep.");

File: service/src/main/java/co/cask/wrangler/service/directive/DirectivesService.java [hunk 4]
```diff
@@ -265,6 +267,106 @@ public void get(HttpServiceRequest request, HttpServiceResponder responder,
...
+      // Extract charset, if not specified, default it to UTF-8.
+      String charset = handler.getHeader(RequestExtractor.CHARSET_HEADER, "UTF-8");
+
+      // Get content type - application/data-prep, application/octet-stream or text/plain.
+      String contentType = handler.getHeader(RequestExtractor.CONTENT_TYPE_HEADER, "application/data-prep");
+
+      // Extract content.
+      byte[] content = handler.getContent();
+      if (content == null) {
+        error(responder, "Body not present, please post the file containing the records to be wrangle.");
```
awholegunch: What about:
```
"Body not present; missing a file containing records to be processed."

File: service/src/main/java/co/cask/wrangler/service/directive/DirectivesService.java [hunk 5]
```diff
@@ -265,6 +267,106 @@ public void get(HttpServiceRequest request, HttpServiceResponder responder,
...
+   * @param request Handler for incoming request.
+   * @param responder Responder for data going out.
+   */
+  @POST
+  @Path("workspaces")
+  public void upload(HttpServiceRequest request, HttpServiceResponder responder) {
+
+    try {
+      String name = request.getHeader(PropertyIds.FILE_NAME);
+      String id = ServiceUtils.generateMD5(request.getHeader(name));
```
elfenheart: should this be `ServiceUtils.generateMD5(name)`?

service/src/main/java/co/cask/wrangler/service/directive/DirectivesService.java
```diff
@@ -265,6 +267,106 @@ public void get(HttpServiceRequest request, HttpServiceResponder responder,
...
+
+      // if workspace doesn't exist, then we create the workspace before
+      // adding data to the workspace.
+      if (!table.hasWorkspace(id)) {
+        table.createWorkspaceMeta(id, id);
```
elfenheart: can the displayName of the workspace be the file name?

elfenheart: And also perhaps the first column should be `body` instead of the workspace id as well

bdmogal:
@nitinmotgi @elfenheart I've pushed a couple of commits. please review and merge if it looks ok.

elfenheart (approved):
Tested the change, and it works. 👍  LGTM

PR #98, opened 2017-05-11
Updating json to work with 4.1.1

russorat:
just updating artifact versions and fixing an issue with double quoting the \r\n


anwar6953:
LGTM

PR #99, opened 2017-05-11
[CDAP-11486] adding BYTES conversion for structured records

russorat:
https://issues.cask.co/browse/CDAP-11486


nitinmotgi (approved):
LGTM 👍

PR #109, opened 2017-05-26
Correct username property for database source

bdmogal:
This is to have the correct property per what the database source expects at https://github.com/caskdata/hydrator-plugins/blob/release/1.7/database-plugins/src/main/java/co/cask/hydrator/plugin/ConnectionConfig.java#L31


elfenheart (approved):
👍  LGTM

PR #115, opened 2017-06-02
[CDAP-11711] Rename Postgres driver mapping to postgres65 to avoid na…

sameetandpotatoes:
Fixes https://issues.cask.co/browse/CDAP-11771

Makes drivers mapping more consistent, so that PostgreSQL v6.5 and earlier and PostgreSQL v7.0 and later map to shorter names.


nitinmotgi (approved):
LGTM 👍

PR #118, opened 2017-06-03
[CPAP-11786] Ensures that null values are filtered correct

archilmedes:
I believe this is the correct fix to CDAP-11786. `idx` represents the column number that is being affected by the filtering and the -1 check is checking if the column name is in the table. `object` represents the actual value in the column for the particular entry and this is where we have to ensure whether or not null values are being properly handled. I have also included a test. 

@nitinmotgi does this seem right to you?

### DIFF
--- a/core/src/test/java/co/cask/wrangler/steps/row/RecordRegexFilterTest.java
+++ b/core/src/test/java/co/cask/wrangler/steps/row/RecordRegexFilterTest.java
@@ -1,0 +1,68 @@
+/*
+ * Copyright © 2017 Cask Data, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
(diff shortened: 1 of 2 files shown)


core/src/test/java/co/cask/wrangler/steps/row/RecordRegexFilterTest.java
```diff
@@ -0,0 +1,68 @@
...
+
+import java.util.Arrays;
+import java.util.List;
+
+/**
+ * Created by archilshah on 6/2/17.
+ */
+public class RecordRegexFilterTest {
+
+    @Test
```
anwar6953: nit - indentation seems inconsistent w/ rest of repository (2 spaces vs 4)

nitinmotgi:
LGTM 👍

PR #119, opened 2017-06-03
[CDAP-11730] Parse JSON as DataType.Text to fix JSParser

sameetandpotatoes:
Parses JSON files as a `DataType.TEXT` through the filebrowser so that the `parse-as-json` directive works as expected.

### DIFF
--- a/core/src/main/java/co/cask/wrangler/steps/parser/JsParser.java
+++ b/core/src/main/java/co/cask/wrangler/steps/parser/JsParser.java
@@ -90,7 +90,7 @@ public class JsParser extends AbstractStep {
           JsonElement element = null;
           if(value instanceof String) {
             String document = (String) value;
-            element = parser.parse(document);
+            element = parser.parse(document.trim());
           } else if (value instanceof JsonObject || value instanceof JsonArray) {
             element = (JsonElement) value;
           } else {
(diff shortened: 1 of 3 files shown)


core/src/main/java/co/cask/wrangler/steps/parser/JsParser.java
```diff
@@ -90,7 +90,7 @@ public JsParser(int lineno, String detail, String col, int maxDepth) {
           JsonElement element = null;
           if(value instanceof String) {
             String document = (String) value;
-            element = parser.parse(document);
+            element = parser.parse(document.trim());
```
sameetandpotatoes, line 93: When treating a `json` file as a `DataType.TEXT`, extra spaces are consistently added at the end of the input string, so the `JsonParser` here throws an error, unless you trim the input string before parsing it.

core/src/test/java/co/cask/wrangler/steps/parser/JsParserTest.java
```diff
@@ -142,4 +142,18 @@ public void testFlattenToDepth() throws Exception {
...
   }
+
+  @Test
+  public void testParsingExtraCharacters() throws Exception {
+    String[] directives = new String[] {
+      "parse-as-json body"
+    };
+
+    List<Record> records = Arrays.asList(
+      new Record("body", "[1,2,3,4,5]             ")
```
sameetandpotatoes, line 153: Because we don't currently have tests for the `FilebrowserExplorer` (or a way to open a file through the `FilebrowserExplorer` through a test, it's difficult to simulate the issue that is seen when parsing the JSON since we just have these unit tests, but this is the sort of input that gets sent to the `JsParser`.

nitinmotgi:
LGTM +1

PR #121, opened 2017-06-05
[WIP] Support for User Defined Directives (UDD)

nitinmotgi:
- [x] Change all steps to use CDAP Plugin Annotations
- [x] Create a separate module for API and move all API classes to that module.
- [x] Dynamically load Modules and Register with Usage Registry
- [x] Delegating parsing and detection of directive to the library. 
- [x] Convert previously written directives to use new API
- [x] Migrate all tests to use new API
- [x] Create a third party module that includes modules that are not packaged. 
- [x] Create a example-directive repository and seed it with a working directive project. 
- [x] Ensure that the ```load-directive``` is not executed every time, it's more of compile time. 
- [x] Separate Wrangler API into a separate module.
- [x] Create a separate Wrangler Testing Module
- [x] Add endpoint for new Usage API
- [x] Automatically generate usage information
- [x] Add API documentation
- [x] Add UI mapping element to `UsageDefintion`
- [x] Add Label to `UsageDefinition`
- [x] Add service endpoints for retrieving artifact and plugin information
- [x] Modify Service to use new API
- [x] Modify Wrangler Transform to use new API
- [x] Map new Usage API to return only result format for now
- [x] Improve compiler error messages
- [x] Add documentation for how to develop, package, deploy and manage custom directives
- [ ] Add `destroy()` method to the Directive API


core/pom.xml
```diff
@@ -11,13 +11,24 @@
...
+    </dependency>
+    <dependency>
       <groupId>org.apache.httpcomponents</groupId>
       <artifactId>httpclient</artifactId>
       <version>${httpclient.version}</version>
     </dependency>
     <dependency>
+      <groupId>org.reflections</groupId>
+      <artifactId>reflections</artifactId>
+      <version>0.9.11</version>
```
ken-cask: Is it OK if you downgrade the version to 0.9.10? I tried using 0.9.11 version, and it gave method not found exception because of Guava version conflict, when I ran an unit test that create Reflections. 0.9.10 version works for me.

PR #123, opened 2017-06-08
Feature/cdap-11435

ken-cask:
Allow a user to specify an unlimited number of acceptable errors, by setting error threshold to -1.


russorat (approved):
lgtm

nitinmotgi:
What is the use case for this ?

ken-cask:
By setting the threshold to -1, the pipe pile will never stop because there are too many errors. Otherwise for example setting it to 10, pipeline will stop and fail after more than 10 error occurs in wrangler.

PR #124, opened 2017-06-08
Feature/cdap-11868

ken-cask:
Added a directive "set-type" to convert data type of a column.
 Accepted types to covert from / convert to are: int, double, string, boolean (https://issues.cask.co/browse/CDAP-11868)


nitinmotgi (approved):
LGTM

PR #127, opened 2017-06-08
bump wrangler version and fix cdap dependency version

albertshau:
1.3.x versions of wrangler are supposed to be compatible with
4.1.x versions of cdap. It should not have 4.2.0-SNAPSHOT as the
cdap dependency.


nitinmotgi:
LGTM 👍

PR #130, opened 2017-06-15
Updated the set-column doc to indicate that:

bdmogal:
1. if the column exists already, its value will be overwritten with the result of the specified expression
2. if the column doesn't exist, a new column will be created with the value as the result of the specified expression


nitinmotgi:
LGTM 👍

PR #137, opened 2017-06-23
Downgrade guava version

CuriousVini:
Issue: https://issues.cask.co/browse/CDAP-11980
Build: https://builds.cask.co/browse/HYP-WT118-2

### DIFF
--- a/core/src/main/java/co/cask/wrangler/steps/transformation/Quantization.java
+++ b/core/src/main/java/co/cask/wrangler/steps/transformation/Quantization.java
@@ -21,12 +21,10 @@ import co.cask.wrangler.api.PipelineContext;
 import co.cask.wrangler.api.Record;
 import co.cask.wrangler.api.StepException;
 import co.cask.wrangler.api.Usage;
-import com.google.common.collect.Range;
-import com.google.common.collect.RangeMap;
-import com.google.common.collect.TreeRangeMap;
 
 import java.util.ArrayList;
 import java.util.List;
+import java.util.TreeMap;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
@@ -41,7 +39,7 @@ import java.util.regex.Pattern;
 )
 public class Quantization extends AbstractStep {
   private static final String RANGE_PATTERN="([+-]?\\d+(?:\\.\\d+)?):([+-]?\\d+(?:\\.\\d+)?)=(.[^,]*)";
-  private final RangeMap<Double, String> rangeMap = TreeRangeMap.create();
+  private final TreeMap<Double, String> rangeMap = new TreeMap<>();
 
   private String col1;
   private String col2;
@@ -55,7 +53,9 @@ public class Quantization extends AbstractStep {
     while(matcher.find()) {
       double lower = Double.parseDouble(matcher.group(1));
       double upper = Double.parseDouble(matcher.group(2));
-      rangeMap.put(Range.closed(lower, upper), matcher.group(3));
+      rangeMap.put(lower, matcher.group(3));
+      // store null in place of upper bound
+      rangeMap.put(upper, null);
     }
   }
 
--- a/core/src/test/java/co/cask/wrangler/steps/PipelineTest.java
+++ b/core/src/test/java/co/cask/wrangler/steps/PipelineTest.java
@@ -24,9 +24,6 @@ import co.cask.wrangler.api.StepException;
 import co.cask.wrangler.parser.TextDirectives;
 import co.cask.wrangler.steps.transformation.MaskShuffle;
 import co.cask.wrangler.steps.transformation.Split;
-import com.google.common.collect.Range;
-import com.google.common.collect.RangeMap;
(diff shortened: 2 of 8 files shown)


anew:
Do you have a Jira for this, explaining why this is needed?

core/src/test/java/co/cask/wrangler/steps/PipelineTest.java
```diff
@@ -101,22 +100,35 @@ public void testMaskSuffle() throws Exception {
...
+    rangeMap.put(0.9, null);
+    rangeMap.put(2.0, "B");
+    rangeMap.put(3.9, null);
+    rangeMap.put(4.0, "C");
+    rangeMap.put(5.9, null);
+
+    String s = mappedValue(rangeMap, 2.2);
     Assert.assertEquals("B", s);
 
     Matcher m = Pattern.compile("([+-]?\\d+(?:\\.\\d+)?):([+-]?\\d+(?:\\.\\d+)?)=(.[^,]*)").matcher("0.9:2.1=Foo,2.2:3.4=9.2");
```
anew, line 114: Is this the same pattern as in the Quantizer? It seems that this should be a public (or visible to the test case is enough) constant then

core/src/main/java/co/cask/wrangler/steps/transformation/Quantization.java
```diff
@@ -55,7 +53,9 @@ public Quantization(int lineno, String detail, String col1, String col2, String
     while(matcher.find()) {
       double lower = Double.parseDouble(matcher.group(1));
       double upper = Double.parseDouble(matcher.group(2));
-      rangeMap.put(Range.closed(lower, upper), matcher.group(3));
+      rangeMap.put(lower, matcher.group(3));
+      // store null in place of upper bound
+      rangeMap.put(upper, null);
```
anew, line 58: don't you need to change the way the map is searched? (in line 92)
I feel that it would be better to introduce a helper class that replaces RangeMap.

core/src/test/java/co/cask/wrangler/steps/PipelineTest.java
```diff
@@ -101,22 +100,35 @@ public void testMaskSuffle() throws Exception {
...
-    rangeMap.put(Range.closed(4.0, 5.9), "C");
-    String s = rangeMap.get(2.2);
+    TreeMap<Double, String> rangeMap = new TreeMap<>();
+    rangeMap.put(0.1, "A");
+    rangeMap.put(0.9, null);
+    rangeMap.put(2.0, "B");
+    rangeMap.put(3.9, null);
+    rangeMap.put(4.0, "C");
+    rangeMap.put(5.9, null);
+
```
anew, line 110: Do you need a test where the ranges overlap? I don't think the code prevents that. I am also not sure what the semantics should be in that case... and what RangeMap would have done?

core/src/test/java/co/cask/wrangler/steps/PipelineTest.java
```diff
@@ -101,22 +100,35 @@ public void testMaskSuffle() throws Exception {
...
-      rm.put(Range.closed(lower, upper), value);
+      stringRangeMap.put(lower, value);
+      stringRangeMap.put(upper, null);
     }
-    Assert.assertEquals("[[0.9‥2.1]=Foo, [2.2‥3.4]=9.2]", rm.toString());
+    Assert.assertEquals("{0.9=Foo, 2.1=null, 2.2=9.2, 3.4=null}", stringRangeMap.toString());
+  }
+
+  private static <K, V> V mappedValue(TreeMap<K, V> rangeMap, K key) {
+    Map.Entry<K, V> e = rangeMap.floorEntry(key);
```
anew, line 127: I think this needs to be ceilingEntry

core/src/test/java/co/cask/wrangler/steps/PipelineTest.java
```diff
@@ -101,22 +100,35 @@ public void testMaskSuffle() throws Exception {
...
+    if (e != null && e.getValue() == null) {
+      e = rangeMap.lowerEntry(key);
```
anew, line 129: and this needs to be floorEntry?

anew (commented):
I think it would be best to write a helper class that replaces RangeMap, with its own set of complete test cases.

pom.xml
```diff
@@ -82,7 +82,7 @@
     <commons-jexl.version>3.0</commons-jexl.version>
     <commons-csv.version>1.4</commons-csv.version>
     <commons-lang.version>3.5</commons-lang.version>
-    <guava.version>19.0</guava.version>
+    <guava.version>13.0.1</guava.version>
```
nitinmotgi, line 85: Should the guava dependency become ```provided```

core/src/test/java/co/cask/wrangler/steps/PipelineTest.java
CuriousVini: Confirmed with @nitinmotgi : the ranges will not overlap

PR #138, opened 2017-06-30
Fix text messages

bdmogal:


nitinmotgi (approved):
lgtm

bdmogal:
will merge after a few UI changes, just to make sure that end-to-end works.

service/src/main/java/co/cask/wrangler/service/kafka/KafkaService.java
```diff
@@ -281,12 +282,11 @@ public void specification(HttpServiceRequest request, final HttpServiceResponder
 
       Map<String, String> properties = new HashMap<>();
       properties.put("topic", topic);
-      properties.put("referenceName", topic);
+      properties.put("referenceName", topic + "_" + UUID.randomUUID());
```
nitinmotgi: Why unique id -- it should be just topic ?

bdmogal: it clashes with the kafka offset table name. It defaults to the topic name.

We can't use a hard coded offset table name, because then every time you create a pipeline it uses the same offset table, so offsets consumed by a pipeline affect those of the other pipeline. 

One thing we could do is keep reference name as topic only, and make the offset table name unique. Would that work?

bdmogal: On second thoughts, it might actually be better to keep the reference name as topic only, and offset name unique, because the reference name shows up in lineage diagrams, etc. I can make that change.

PR #144, opened 2017-07-31
Fixing guava conflicts in wrangler

CuriousVini:
The wrangler Integration tests were failing on MapR clusters because of conflicting guava dependencies. This could happen on any cluster. The fix is to exclude conflicting dependency pulled in transitively from `sisu-guava`

Tested it on MapR 5.2 ITN cluster

Build: https://builds.cask.co/browse/HYP-WT126-3

### DIFF
--- a/wrangler-core/pom.xml
+++ b/wrangler-core/pom.xml
@@ -23,13 +23,7 @@
     </dependency>
     <dependency>
       <groupId>org.antlr</groupId>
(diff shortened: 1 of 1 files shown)


anwar6953 (approved):

CuriousVini:
Build is green. Tested fix on MapR 5.2 cluster. Merging it.

PR #149, opened 2017-08-18
CDAP-12130 Send-to-error does not work on derived columns

nitinmotgi:


wrangler-service/src/main/java/co/cask/wrangler/service/explorer/Explorer.java
```diff
@@ -118,8 +119,12 @@ public Explorer(DatasetProvider provider) {
       response.put("count", values.size());
       response.put("values", values);
       return response;
+    } catch(AccessControlException e) {
+      throw new ExplorerException(String.format(
```
anwar6953, line 123: The original `AccessControlException` exception very likely has information about what the path's permissions are and what the requesting user is. Throwing an ExplorerException like this means you're losing out on that valuable information.
Usually, the `AccessControlException`'s message will be sufficient.

PR #159, opened 2017-09-12
Support for Numeric type for row filter

CuriousVini:
Issue: https://issues.cask.co/browse/CDAP-12428

### DIFF
--- a/wrangler-core/src/main/java/co/cask/directives/row/RecordRegexFilter.java
+++ b/wrangler-core/src/main/java/co/cask/directives/row/RecordRegexFilter.java
@@ -101,11 +101,15 @@ public class RecordRegexFilter implements Directive {
           }
         } else if (object instanceof String) {
           String value = (String) row.getValue(idx);
-          if (isMatching(value)) continue;
+          if (isMatching(value)) {
+            continue;
+          }
         } else if (object instanceof Number) {
           Number number = (Number) row.getValue(idx);
           String value = number.toString();
-          if (isMatching(value)) continue;
+          if (isMatching(value)) {
+            continue;
+          }
         } else {
           throw new DirectiveExecutionException(
             String.format("%s : Invalid value type '%s' of column '%s'. Should be of type String.",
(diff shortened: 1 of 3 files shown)


wrangler-core/src/main/java/co/cask/directives/row/RecordRegexFilter.java
```diff
@@ -105,13 +101,11 @@ public void destroy() {
...
-            matches = !matches;
-          }
-          if (matches) {
-            continue;
-          }
+          if (isMatching(value)) continue;
+        } else if (object instanceof Number) {
+          Number number = (Number) row.getValue(idx);
+          String value = number.toString();
+          if (isMatching(value)) continue;
```
bdmogal: break this for consistency?

bdmogal (approved):
lgtm, but can you get another set of eyes on this?

nitinmotgi:
Don't see test cases. Please add test cases for int,long,short,string,double,float,Integer,Short,Long,Double,Float please.

wrangler-core/src/main/java/co/cask/directives/row/RecordRegexFilter.java
```diff
@@ -105,13 +101,11 @@ public void destroy() {
...
         } else if (object instanceof String) {
           String value = (String) row.getValue(idx);
-          boolean matches = pattern.matcher(value).matches(); // pattern.matcher(value).matches();
-          if(!matched) {
-            matches = !matches;
-          }
-          if (matches) {
-            continue;
-          }
+          if (isMatching(value)) continue;
```
nitinmotgi: put { } before and after continue.

nitinmotgi: Optimize code were you put isMatching in only one place. If so, then you don't need isMatching function.

CuriousVini:
@nitinmotgi Thanks for the review. I have added the test and optimized the code. However, I could only add test for Integer, Double and String types because of https://issues.cask.co/browse/CDAP-12605.

CuriousVini:
Thanks @nitinmotgi I have added the tests and optimized the code. Please take a quick look when you get a chance, thanks

nitinmotgi:
LGTM

PR #166, opened 2017-09-20
Handle oracle's ROWID datatype

CuriousVini:
Issue: https://issues.cask.co/browse/CDAP-12639

image

Tested the fix. Here RID is a `RowId` type.


bdmogal:
@CuriousVini does this work when you create a pipeline/ingest data?

CuriousVini:
@bdmogal Does it throw same exception when used in pipeline as well?

bdmogal:
@CuriousVini I am not sure, but we should test the end-to-end flow, since we've discovered this issue in Data Prep.

CuriousVini:
Agreed. I will test that. But I think the fix will go in Database source/hydrator.

bdmogal:
lgtm, presuming further testing of database source done separately.

PR #181, opened 2017-11-03
GCS service

nitinmotgi:


bdmogal:
I think we should revert this PR to resolve conflicts between release and develop

PR #198, opened 2018-01-06
CDAP-13038 - Handling empty string

nitinmotgi:
... converting empty string to null when destination schema is nullable.


sreevatsanraman:
Changes look good. Can we add a test?

nitinmotgi:
@sreevatsanraman added test cases (+ & -).

sreevatsanraman (approved):
LGTM

PR #207, opened 2018-02-27
Changes to include group as optional parameter for creating workspace

nitinmotgi:


nitinmotgi:
@sreevatsanraman @ajainarayanan can this be merged?

wrangler-service/src/main/java/co/cask/wrangler/dataset/workspace/WorkspaceDataset.java
```diff
@@ -204,6 +235,39 @@ public void deleteWorkspace(String id) throws WorkspaceException {
...
+   *
+   * @param group to be deleted.
+   * @return number of workspaces deleted within a group.
+   * @throws WorkspaceException thrown if there are issues with deleting workspace within group.
+   */
+  @WriteOnly
+  public int deleteGroup(String group) throws WorkspaceException {
+    int count = 0;
+    co.cask.cdap.api.dataset.table.Row row;
+    try (Scanner scanner = table.scan(null, null)) {
```
sreevatsanraman, line 249: How big can the table get? Since it is doing a full table scan, this might never succeed on a slow cluster with a lot of entries since it executes in one transaction. Should batch deletes in separate transaction.

sreevatsanraman:
One comment about the delete. Okay to address the specific comment in a follow up PR.

wrangler-service/src/main/java/co/cask/wrangler/dataset/workspace/WorkspaceDataset.java
nitinmotgi: @sreevatsanraman Workspaces are scoped within a namespace and within a namespace they are also grouped by a group-id. The groups are controlled by system. Some examples of groups are 'default', 'mmds'  and 'dataprep'. So there are only limited number.

sreevatsanraman (approved):

PR #215, opened 2018-07-26
update mssql jdbc driver mapping name

elfenheart:


bdmogal (approved):
lgtm so long as its tested.

PR #218, opened 2018-08-14
configure to make project and credentials optional for GCS and BigQuery

gsps1:
Making credentials and project informational as optional settings when constructing `Storage` or `BigQuery`.


albertshau (approved):
lgtm

PR #220, opened 2018-08-17
surfacing available connection types and creating default connections

gsps1:
JIRA : https://issues.cask.co/browse/CDAP-14017

Adding support in dataprep for disabling certain connection-types and creating default connections.

Updated the properties map from `<String, Object>` to `<String, String>`, Object causes artifact inspection to fail when we are adding an artifact and it didn't really seem necessary.

### DIFF
--- a/wrangler-service/src/main/java/co/cask/wrangler/service/bigquery/BigQueryService.java
+++ b/wrangler-service/src/main/java/co/cask/wrangler/service/bigquery/BigQueryService.java
@@ -210,10 +210,10 @@ public class BigQueryService extends AbstractWranglerService {
       scope = WorkspaceDataset.DEFAULT_SCOPE;
     }
 
-    Map<String, Object> connectionProperties = connection.getAllProps();
-    String projectId = (String) connectionProperties.get(GCPUtils.PROJECT_ID);
-    String path = (String) connectionProperties.get(GCPUtils.SERVICE_ACCOUNT_KEYFILE);
-    String bucket = (String) connectionProperties.get(BUCKET);
+    Map<String, String> connectionProperties = connection.getAllProps();
+    String projectId = connectionProperties.get(GCPUtils.PROJECT_ID);
+    String path = connectionProperties.get(GCPUtils.SERVICE_ACCOUNT_KEYFILE);
+    String bucket = connectionProperties.get(BUCKET);
 
     TableId tableIdObject = TableId.of(projectId, datasetId, tableId);
 
--- a/wrangler-service/src/main/java/co/cask/wrangler/service/connections/ConnectionService.java
+++ b/wrangler-service/src/main/java/co/cask/wrangler/service/connections/ConnectionService.java
@@ -31,9 +31,15 @@ import com.google.gson.Gson;
 import com.google.gson.JsonArray;
 import com.google.gson.JsonObject;
 import com.google.gson.JsonPrimitive;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import java.net.HttpURLConnection;
+import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
+import java.util.Set;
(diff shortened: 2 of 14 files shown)


wrangler-service/src/main/java/co/cask/wrangler/service/bigquery/BigQueryService.java
```diff
@@ -210,7 +210,7 @@ public void readTable(HttpServiceRequest request, HttpServiceResponder responder
       scope = WorkspaceDataset.DEFAULT_SCOPE;
     }
 
-    Map<String, Object> connectionProperties = connection.getAllProps();
+    Map<String, String> connectionProperties = connection.getAllProps();
     String projectId = (String) connectionProperties.get(GCPUtils.PROJECT_ID);
```
albertshau: none of the casting is needed anymore, I imagine it's all over the place

wrangler-service/src/main/java/co/cask/wrangler/service/connections/ConnectionService.java
```diff
@@ -68,6 +89,38 @@
...
+      if (defaultConnection.getType() == null || defaultConnection.getType() == ConnectionType.UNDEFINED) {
+        LOG.warn("Skipping the default connection {}, type is missing or un-recognized", defaultConnection.getName());
+        continue;
+      }
+      try {
+        store.create(defaultConnection);
+      } catch (Exception e) {
+        // we don't want to disrupt data-prep if we are not able to create a default connection,
+        // log the error and continue
+        LOG.debug("Exception while creating default connection ", e);
```
albertshau: should be a warning, not debug. Mention the name of the connection in the exception message.

wrangler-service/src/main/java/co/cask/wrangler/service/s3/S3Configuration.java
```diff
@@ -33,7 +33,7 @@
   private final String region;
 
   S3Configuration(Connection connection) {
-    Map<String, Object> properties = connection.getAllProps();
+    Map<String, String> properties = connection.getAllProps();
```
albertshau, line 36: there should be corresponding changes to remove parsing/casting

wrangler-service/src/test/java/co/cask/wrangler/dataset/connections/ConnectionTest.java
```diff
@@ -41,7 +41,7 @@ public void testConnectionSerDe() throws Exception {
     connection.setCreated(System.currentTimeMillis()/1000);
     connection.setUpdated(System.currentTimeMillis()/1000);
     connection.putProp("hostname", "localhost");
-    connection.putProp("port", 3306);
+    connection.putProp("port", "3306");
```
albertshau, line 44: does this mean database connection requires an integer here? Does that mean existing db connections will break?

albertshau (commented):
I'm not sure the change from Object to String is safe. Are we sure every existing connection uses String?

wrangler-service/src/main/java/co/cask/wrangler/service/bigquery/BigQueryService.java
gsps1: yeah, have removed all the casting now.

wrangler-service/src/test/java/co/cask/wrangler/dataset/connections/ConnectionTest.java
gsps1: no, this test isn't really accurate as this doesn't use the DatabaseService, rather just tests connection serialization. DatabaseService expects connectionString as connection property.

gsps1:
@albertshau i have tested this change for all connections except S3 and they work, for S3 i have checked all property values are string.

albertshau:
ok good. It would have been a headache to worry about backwards compatibility

albertshau (approved):
lgtm

PR #221, opened 2018-08-17
Skip project id if its null, can be optional if the instance is on GC…

gsps1:
…P. 

Fix for NPE we saw when we read big query table in Dataprep in GCP.


albertshau (approved):
lgtm

PR #222, opened 2018-08-20
Convert String to Integer using Integer.parseInt wrapper method.

sagarkapare:
Converting String to Number fails because they are not part of same inheritance hierarchy. Using `Integer.parseInt` instead.


chtyim (approved):
LGTM

PR #224, opened 2018-08-27
Feature/cdap 14021 logical type

CuriousVini:
**Summary**

- Adding support for logical types in wrangler to convert Row to StructuredRecord and vice versa
- Adding support for logical types in Database service and BigQuery Service
- Adding missing license files to several files and RAT plugin
- Testcase to test logical types in Database Service 

**Remaining**
- WIP on support for other date/time related directives.
    - parse-as-date
    - parse-as-simple-date
     - diff-date
     - format-date


pom.xml
```diff
@@ -183,6 +184,31 @@
           <artifactId>maven-antrun-plugin</artifactId>
           <version>1.7</version>
         </plugin>
+        <plugin>
+          <groupId>org.apache.rat</groupId>
```
albertshau, line 188: thank you for adding this, it is much appreciated and needed. Next time though, open a separate PR for this and the relevant changes.

wrangler-core/src/main/java/co/cask/directives/column/SetType.java
```diff
@@ -256,11 +263,52 @@ private Object convertType(String toType, Object object) throws Exception {
...
+      case "DATE": {
+        if (object instanceof Integer) {
+          object = Long.valueOf((Integer) object);
+        }
+
+        if (object instanceof Long) {
+          return LocalDate.ofEpochDay((Long) object);
+        }
+
+        throw new DirectiveExecutionException(String.format("Only integer or long representing number days from epoch" +
```
albertshau: I think throwing an exception is the correct thing to do, but the other methods seem to just return the object if it's not one of the expected type. We should make the behavior consistent with the other types.

Please open a jira about this.

wrangler-core/src/main/java/co/cask/directives/column/SetType.java
```diff
@@ -256,11 +263,52 @@ private Object convertType(String toType, Object object) throws Exception {
         }
       }
 
+      case "DATE": {
+        if (object instanceof Integer) {
+          object = Long.valueOf((Integer) object);
+        }
+
+        if (object instanceof Long) {
```
albertshau: the other cases use if - else if - else. Let's make these new ones consistent.

wrangler-core/src/main/java/co/cask/directives/column/SetType.java
```diff
@@ -256,11 +263,52 @@ private Object convertType(String toType, Object object) throws Exception {
...
+        throw new DirectiveExecutionException(String.format("Only long representing timestamp in microseconds " +
+                                                              "can be converted to %s. ",
+                                                            toType));
+      }
+
       default:
         throw new DirectiveExecutionException(
           String.format("Unknown data type '%s' found in the directive. " +
-                  "Accepted types are: int, short, long, double, boolean, string, bytes", toType)
+                  "Accepted types are: int, short, long, double, boolean, string, bytes, date, time, timestamp", toType)
```
albertshau: update javadoc for this class too.

albertshau: what happens if somebody sets the type of a column to a timestamp, then sets the type to a long? It will act very strangely. Also, I think most users would try to do set-type on a string, and then would be surprised that it doesn't work.

wrangler-core/src/main/java/co/cask/directives/column/SetType.java
```diff
@@ -256,11 +263,52 @@ private Object convertType(String toType, Object object) throws Exception {
...
         );
     }
   }
+
+  private ZonedDateTime getZonedDateTime(long microTs) {
```
albertshau: I don't think it makes sense to use micros here. Most of the time, users will not have micros in their data.

wrangler-core/src/main/java/co/cask/wrangler/utils/Json2Schema.java
```diff
@@ -89,6 +92,22 @@ public Schema toSchema(String id, Row row) throws RecordConvertorException {
         fields.add(Schema.Field.of(name, schema));
       }
 
+      if (value instanceof LocalDate) {
```
albertshau, line 95: this looks really messed up that schema is determined by a series of instanceof checks. FYI though you have to follow the existing pattern, it's not a good pattern to follow in future classes.

wrangler-core/src/test/java/co/cask/wrangler/utils/ObjectSerDeTest.java
```diff
@@ -54,4 +58,31 @@ public void testNull() throws Exception {
...
+    Row firstRow = new Row();
+    firstRow.add("id", 1);
+    firstRow.add("name", "abc");
+    firstRow.add("date", LocalDate.of(2018, 11, 11));
+    firstRow.add("time", LocalTime.of(11, 11, 11));
+    firstRow.add("timestamp", ZonedDateTime.of(2018, 11 , 11 , 11, 11, 11, 0, ZoneId.of("UTC")));
+    expectedRows.add(firstRow);
+    byte[] bytes = objectSerDe.toByteArray(expectedRows);
+    List<Row> actualRows = objectSerDe.toObject(bytes);
+    Assert.assertEquals(expectedRows.size(), actualRows.size());
```
albertshau, line 75: The existing tests in this class that only check size equality are not good, and we should not continue that pattern in new tests. The actual objects should be equal, especially for something that is supposedly checking serialization and deserialization. Does the Row class not have equals and hash code?

wrangler-service/src/main/java/co/cask/wrangler/service/bigquery/BigQueryService.java
```diff
@@ -334,14 +341,21 @@ public void specification(HttpServiceRequest request, final HttpServiceResponder
...
           case DATETIME:
           case TIMESTAMP:
-            row.add(fieldName, fieldValue.getStringValue());
+            long tsMicroValue = fieldValue.getTimestampValue();
+            row.add(fieldName, getZonedDateTime(tsMicroValue));
             break;
 
           case STRING:
-            row.add(fieldName, fieldValue.getStringValue());
+            row.add(fieldName, fieldValue.getStringValue ());
```
albertshau: extra space. Checkstyle does not catch this?

wrangler-service/src/main/java/co/cask/wrangler/service/bigquery/BigQueryService.java
```diff
@@ -334,14 +341,21 @@ public void specification(HttpServiceRequest request, final HttpServiceResponder
...
+            break;
+
           case DATE:
+            row.add(fieldName, LocalDate.parse(fieldValue.getStringValue()));
+            break;
+
           case DATETIME:
           case TIMESTAMP:
-            row.add(fieldName, fieldValue.getStringValue());
+            long tsMicroValue = fieldValue.getTimestampValue();
```
albertshau, line 352: big query returns micros here?

wrangler-core/src/main/java/co/cask/directives/column/SetType.java
albertshau: I think we should support changing a date/time/timestamp to some other type through set-type, but not the other way around. Going from a primitive type to date/time/timestamp should be done by a connection, or by the time parsing directives.

albertshau: Let's do the set-type work in a separate PR though. There are too many things going on in this PR. In the future, break work up into multiple PRs when possible. One for each connection or directive.

wrangler-service/src/test/java/co/cask/wrangler/service/database/DatabaseServiceTest.java
```diff
@@ -82,4 +126,141 @@ public void testReadingDriverConfiguration() throws Exception {
...
+    }
+
+    String getConnectionUrl() {
+      return this.connectionUrl;
+    }
+
+    String getDatabase() {
+      return this.database;
+    }
+  }
```
albertshau: existing file has a problem with no newline at end of file.

wrangler-service/src/test/java/co/cask/wrangler/service/database/DatabaseServiceTest.java
```diff
@@ -82,4 +126,141 @@ public void testReadingDriverConfiguration() throws Exception {
     }
     Assert.assertEquals(9, drivers.size());
   }
+
+  @Test
+  public void databaseWithLogicalTypes() throws Exception {
+    String hsqlDBDir = temporaryFolder.newFolder("hsqldb").getAbsolutePath();
+    TimeZone.setDefault(TimeZone.getTimeZone("UTC"));
+    hsqlDBServer = new HSQLDBServer(hsqlDBDir, "testdb");
```
albertshau: starting up the test db should happen in a `@BeforeClass` method, so that all tests can use it.

wrangler-service/src/test/java/co/cask/wrangler/service/database/DatabaseServiceTest.java
```diff
@@ -82,4 +126,141 @@ public void testReadingDriverConfiguration() throws Exception {
...
+
+  protected String getDatabase() {
+    return hsqlDBServer == null ? null : hsqlDBServer.getDatabase();
+  }
+
+  protected Connection getConnection() {
+    return hsqlDBServer == null ? null : hsqlDBServer.getConnection();
+  }
+
+  private static class HSQLDBServer {
```
albertshau: thanks for adding a test that actually tests the connection.

wrangler-transform/src/main/java/co/cask/wrangler/Wrangler.java
```diff
@@ -440,6 +446,44 @@ public void transform(StructuredRecord input, Emitter<StructuredRecord> emitter)
...
   }
 
+  private Object getValue(StructuredRecord input, String fieldName, Schema.LogicalType logicalType) {
+    if (logicalType != null) {
+      switch (logicalType) {
+        case DATE:
+          return input.getDate(fieldName);
+        case TIME_MILLIS:
+        case TIME_MICROS:
+          return  input.getTime(fieldName);
```
albertshau: extra space

wrangler-transform/src/main/java/co/cask/wrangler/Wrangler.java
```diff
@@ -440,6 +446,44 @@ public void transform(StructuredRecord input, Emitter<StructuredRecord> emitter)
...
+        case TIMESTAMP_MILLIS:
+        case TIMESTAMP_MICROS:
+          return  input.getTimestamp(fieldName);
```
albertshau: extra space

wrangler-transform/src/main/java/co/cask/wrangler/Wrangler.java
```diff
@@ -440,6 +446,44 @@ public void transform(StructuredRecord input, Emitter<StructuredRecord> emitter)
...
+
+  @Nullable
+  private Schema.LogicalType getLogicalType(Schema schema) {
+    Schema.LogicalType logicalType = schema.getLogicalType();
+
+    if (logicalType != null) {
+      return logicalType;
+    }
+
+    if (schema.getType() == Schema.Type.UNION) {
```
albertshau: a union represents a field that can be a bunch of different types. For example, [date, string, double, null] means it can be any one of those four fields. So this is not the correct thing to do for the general union case, as we don't want to treat this as if it must be a date.

We do, however, want this behavior for the [date, null] case, which represents a nullable date. So use the isNullable() and getNonNullable() methods instead of checking for any union.

wrangler-transform/src/main/java/co/cask/wrangler/Wrangler.java
```diff
@@ -440,6 +446,44 @@ public void transform(StructuredRecord input, Emitter<StructuredRecord> emitter)
     }
   }
 
+  private Object getValue(StructuredRecord input, String fieldName, Schema.LogicalType logicalType) {
```
albertshau: it is cleaner if this method only takes the record and field name as input and have this method derive the logical type for the field. Otherwise it is possible to call this method in an invalid way.

wrangler-transform/src/main/java/co/cask/wrangler/Wrangler.java
```diff
@@ -432,6 +436,8 @@ public void transform(StructuredRecord input, Emitter<StructuredRecord> emitter)
           if (wObject instanceof String) {
             builder.convertAndSet(field.getName(), (String) wObject);
           } else {
+            // No need to use specific methods for fields of logical type - timestamp, date and time. This is because
```
albertshau, line 436: good comment

wrangler-core/src/test/java/co/cask/wrangler/utils/ObjectSerDeTest.java
CuriousVini: `toString` is used because `Row` class does not have `equals` and `hash` code implemented.

wrangler-core/src/main/java/co/cask/directives/column/SetType.java
CuriousVini: Filed https://issues.cask.co/browse/CDAP-14217

wrangler-core/src/main/java/co/cask/directives/column/SetType.java
CuriousVini: this change is not needed since we have decided not to support date/time in `set-type` directive.

wrangler-service/src/main/java/co/cask/wrangler/service/bigquery/BigQueryService.java
CuriousVini: wrangler does not have checkstyle plugin. I will add that in next PR.

wrangler-service/src/main/java/co/cask/wrangler/service/bigquery/BigQueryService.java
CuriousVini: yea its in micro seconds.

image

wrangler-transform/src/main/java/co/cask/wrangler/Wrangler.java
```diff
@@ -440,6 +443,37 @@ public void transform(StructuredRecord input, Emitter<StructuredRecord> emitter)
     }
   }
 
+  private Object getValue(StructuredRecord input, String fieldName) {
+    Schema fieldSchema = input.getSchema().getField(fieldName).getSchema();
+    fieldSchema = fieldSchema.isNullable() ? fieldSchema.getNonNullable() : fieldSchema;
+    Schema.Type fieldType = fieldSchema.getType();
+
+    switch (fieldType) {
```
albertshau: don't need this switch. Just switch on the logical type, and put the `return input.get(fieldName)` after that.

albertshau (commented):
one minor comment

albertshau (approved):
lgtm

PR #225, opened 2018-08-31
CDAP-14058 Spanner dataprep connection service

gsps1:
Adding Spanner data prep connection. 

Spanner service Endpoints to 

- Test connection
- List spanner instances
- List spanner databases in an instance
- List spanner tables in a database
- Read spanner table data into a workspace
- Get specification for spanner source plugin for creating a pipeline.

JIRA : https://issues.cask.co/browse/CDAP-14058

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -104,7 +104,7 @@
     <avro.version>1.8.1</avro.version>
     <slf4j.version>1.7.5</slf4j.version>
     <google.http.client.version>1.22.0</google.http.client.version>
-    <protobuf.version>3.3.0</protobuf.version>
+    <protobuf.version>3.4.0</protobuf.version>
     <guava.retrying.version>2.0.0</guava.retrying.version>
     <geogson.version>1.1.97</geogson.version>
     <kafka.clients.version>0.10.2.1</kafka.clients.version>
--- a/wrangler-service/src/main/java/co/cask/wrangler/service/spanner/SpannerInstance.java
+++ b/wrangler-service/src/main/java/co/cask/wrangler/service/spanner/SpannerInstance.java
@@ -1,0 +1,28 @@
+/*
+ * Copyright © 2018 Cask Data, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package co.cask.wrangler.service.spanner;
+
+/**
+ * Spanner instance information
+ */
+public class SpannerInstance {
+  private final String name;
+
+  SpannerInstance(String name) {
+    this.name = name;
+  }
+}
--- a/wrangler-service/src/main/java/co/cask/wrangler/service/spanner/SpannerService.java
+++ b/wrangler-service/src/main/java/co/cask/wrangler/service/spanner/SpannerService.java
@@ -1,0 +1,402 @@
+/*
+ * Copyright © 2018 Cask Data, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package co.cask.wrangler.service.spanner;
+
+import co.cask.cdap.api.data.schema.Schema;
+import co.cask.cdap.api.data.schema.UnsupportedTypeException;
+import co.cask.cdap.api.service.http.HttpServiceRequest;
+import co.cask.cdap.api.service.http.HttpServiceResponder;
+import co.cask.wrangler.PropertyIds;
+import co.cask.wrangler.RequestExtractor;
+import co.cask.wrangler.ServiceUtils;
+import co.cask.wrangler.api.Row;
+import co.cask.wrangler.dataset.connections.Connection;
+import co.cask.wrangler.dataset.workspace.DataType;
+import co.cask.wrangler.dataset.workspace.WorkspaceDataset;
+import co.cask.wrangler.service.PluginConfiguration;
+import co.cask.wrangler.service.ServiceResponse;
+import co.cask.wrangler.service.common.AbstractWranglerService;
+import co.cask.wrangler.service.connections.ConnectionType;
+import co.cask.wrangler.service.gcp.GCPUtils;
+import co.cask.wrangler.utils.ObjectSerDe;
+import com.google.cloud.ByteArray;
+import com.google.cloud.spanner.DatabaseId;
+import com.google.cloud.spanner.KeySet;
+import com.google.cloud.spanner.Options;
+import com.google.cloud.spanner.ResultSet;
+import com.google.cloud.spanner.Spanner;
+import com.google.cloud.spanner.Statement;
+import com.google.cloud.spanner.Type;
+import com.google.common.base.Charsets;
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableMap;
+import com.google.gson.Gson;
+import com.google.gson.JsonSyntaxException;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.net.HttpURLConnection;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
+import javax.ws.rs.BadRequestException;
+import javax.ws.rs.DefaultValue;
+import javax.ws.rs.GET;
+import javax.ws.rs.POST;
+import javax.ws.rs.Path;
+import javax.ws.rs.PathParam;
+import javax.ws.rs.QueryParam;
+
+/**
+ * Spanner data prep connection service
+ */
+public class SpannerService extends AbstractWranglerService {
+  private static final Logger LOG = LoggerFactory.getLogger(SpannerService.class);
+  private static final String TABLE_NAME = "TableName";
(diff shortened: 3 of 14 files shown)


pom.xml
```diff
@@ -104,7 +104,7 @@
     <avro.version>1.8.1</avro.version>
     <slf4j.version>1.7.5</slf4j.version>
     <google.http.client.version>1.22.0</google.http.client.version>
-    <protobuf.version>3.3.0</protobuf.version>
+    <protobuf.version>3.4.0</protobuf.version>
```
gsps1, line 107: updated to 3.4.0 as i was noticing the issue with NoSuchMethodError for `AbstractMessageLite#addAll(...)` as specified in this issue https://github.com/protocolbuffers/protobuf/issues/4521

wrangler-service/src/main/java/co/cask/wrangler/service/spanner/SpannerResponse.java
```diff
@@ -0,0 +1,38 @@
...
+  private final int httpStatusCode;
+  private final String statusMessage;
+  private final int count;
+  private final List<T> values;
+
+  public SpannerResponse(List<T> values) {
+    this.httpStatusCode = HttpURLConnection.HTTP_OK;
+    this.statusMessage = "Success";
+    this.count = values.size();
+    this.values = values;
```
gsps1: I tried to follow the response format similar to other services

wrangler-service/src/main/java/co/cask/wrangler/service/spanner/SpannerResponse.java
```diff
@@ -0,0 +1,38 @@
...
+package co.cask.wrangler.service.spanner;
+
+import java.net.HttpURLConnection;
+import java.util.List;
+
+/**
+ * Response sent by spanner service methods
+ * @param <T>
+ */
+public class SpannerResponse<T> {
```
albertshau: this doesn't seem specific to Spanner. Could it be used as the response for other API calls? If so, lets rename it.

wrangler-service/src/main/java/co/cask/wrangler/service/spanner/SpannerService.java
```diff
@@ -0,0 +1,400 @@
...
+public class SpannerService extends AbstractWranglerService {
+  private static final Logger LOG = LoggerFactory.getLogger(SpannerService.class);
+  private static final String TABLE_NAME = "TableName";
+  // Spanner queries for listing tables and listing schema of table are documented at
+  // https://cloud.google.com/spanner/docs/information-schema
+  private static final Statement LIST_TABLES_STATEMENT =
+    Statement.of("SELECT t.table_name FROM " +
+                   "information_schema.tables AS t WHERE t.table_catalog = '' and t.table_schema = ''");
+
+  private static final Statement.Builder SCHEMA_STATEMENT_BUILDER =  Statement.newBuilder(
```
albertshau: extra space before Statement.newBuilder

wrangler-service/src/main/java/co/cask/wrangler/service/spanner/SpannerService.java
```diff
@@ -0,0 +1,400 @@
...
+    String.format("SELECT  t.column_name,t.spanner_type, t.is_nullable FROM information_schema.columns AS t WHERE " +
```
albertshau: extra space after SELECT

wrangler-service/src/main/java/co/cask/wrangler/service/spanner/SpannerService.java
```diff
@@ -0,0 +1,400 @@
...
+                    "  t.table_catalog = ''  AND  t.table_schema = '' AND t.table_name = @%s", TABLE_NAME));
```
albertshau: extra spaces around first AND

wrangler-service/src/main/java/co/cask/wrangler/service/spanner/SpannerService.java
```diff
@@ -0,0 +1,400 @@
...
+  // Default Number of rows read from Spanner table by data-prep
+  private static final String DEFAULT_ROW_LIMIT = "1000";
+  private static final Gson GSON = new Gson();
+
+  /**
+   * Tests Spanner Connection.
+   */
+  @POST
+  @Path("/connections/spanner/test")
```
albertshau, line 89: is this the pattern followed by the other endpoints? If so, it's fine to follow the pattern, but this is not how paths should be. This would normally refer to a 'test' action on a connection with id 'spanner'.

albertshau: If this is how everything works, please open a jira.

wrangler-service/src/main/java/co/cask/wrangler/service/spanner/SpannerService.java
```diff
@@ -0,0 +1,400 @@
...
+  @Path("/spanner/workspaces/{workspace-id}/specification")
+  @GET
+  public void specification(HttpServiceRequest request, final HttpServiceResponder responder,
+                            @PathParam("workspace-id") String workspaceId) {
+    try {
+      Map<String, String> config = ws.getProperties(workspaceId);
+
+      // deserialize and send spanner source specification
+      SpannerSpecification specification =
+        GSON.fromJson(config.get(PropertyIds.PLUGIN_SPECIFICATION), SpannerSpecification.class);
```
albertshau, line 234: this can throw JsonSyntaxException, which should result in a 400.

wrangler-service/src/main/java/co/cask/wrangler/service/spanner/SpannerSpecification.java
```diff
@@ -0,0 +1,43 @@
...
+ */
+public class SpannerSpecification {
+  // gcp properties
+  private final String serviceFilePath;
+  private final String project;
+  // spanner properties
+  private final String instance;
+  private final String database;
+  private final String table;
+  private final Schema schema;
```
albertshau: I assume this means the other ones use the actual Schema class instead of the string version? It's really weird, since the plugin actually expects a string, which means the UI is doing some conversion?

wrangler-service/src/main/java/co/cask/wrangler/service/spanner/SpannerService.java
```diff
@@ -0,0 +1,400 @@
...
+        return Schema.recordOf("tableSchema", schemaFields);
+      }
+    } finally {
+      spanner.close();
+    }
+  }
+
+  private Schema parseSchemaFromSpannerTypeString(String spannerType) throws UnsupportedTypeException {
+    if (spannerType.startsWith("STRING")) {
+      // STRING and BYTES also have size at the end in the format, example : STRING(1024)
```
albertshau, line 277: good comment

wrangler-service/src/main/java/co/cask/wrangler/service/spanner/SpannerInstance.java
```diff
@@ -0,0 +1,28 @@
...
+ * the License.
+ */
+
+package co.cask.wrangler.service.spanner;
+
+/**
+ * Spanner instance information
+ */
+public class SpannerInstance {
+  private final String instanceName;
```
albertshau: is the existing pattern to call these xyzName? Seems redundant to me, since a GET is being made on instances. I would just call it 'name', but let's follow whatever pattern is established.

wrangler-service/src/main/java/co/cask/wrangler/service/spanner/SpannerService.java
```diff
@@ -0,0 +1,400 @@
...
+      List<Row> data = getTableData(connection, instanceId, databaseId, tableId, schema, Long.parseLong(limit));
+
+      // create workspace id
+      String identifier = ServiceUtils.generateMD5(String.format("%s:%s", scope, tableId));
+      ws.createWorkspaceMeta(identifier, scope, tableId);
+
+      // write data to workspace
+      ObjectSerDe<List<Row>> serDe = new ObjectSerDe<>();
+      byte[] dataBytes = serDe.toByteArray(data);
+      ws.writeToWorkspace(identifier, WorkspaceDataset.DATA_COL, DataType.RECORDS, dataBytes);
```
albertshau, line 192: this means all the records are stored into a single cell doesn't it... I think we'll need to create a new backwards incompatible version of dataprep at some point.

albertshau (commented):
logic looks correct. Most comments are questions about existing conventions

wrangler-service/src/main/java/co/cask/wrangler/service/spanner/SpannerSpecification.java
gsps1: Other classes, construct the json object manually, however they convert schema to string and send it, i can store it as string too.

wrangler-service/src/main/java/co/cask/wrangler/service/spanner/SpannerService.java
gsps1: yes, also the data lifecycle is not clearly defined.

wrangler-service/src/main/java/co/cask/wrangler/service/spanner/SpannerInstance.java
gsps1: i will update  to `name`. The naming changes as with constructing JSON manually, prone to error, but i looked at few places and `name` is used.

wrangler-service/src/main/java/co/cask/wrangler/service/spanner/SpannerResponse.java
gsps1: yes, its not specific to Spanner.

wrangler-service/src/main/java/co/cask/wrangler/service/spanner/SpannerService.java
gsps1: yes, there should be a single endpoint with connection-type however that is not the case currently. created JIRA https://issues.cask.co/browse/CDAP-14250

wrangler-service/src/main/java/co/cask/wrangler/service/spanner/SpannerResponse.java
gsps1: Renamed it to `ServiceResponse`.

wrangler-service/src/main/java/co/cask/wrangler/service/spanner/SpannerSpecification.java
```diff
@@ -29,10 +29,10 @@
   private final String instance;
   private final String database;
   private final String table;
-  private final Schema schema;
+  private final String schema;
 
   SpannerSpecification(String serviceFilePath, String project, String instance,
-                       String database, String table, Schema schema) {
+                       String database, String table, String schema) {
```
albertshau: lets still pass in a Schema, but set the field using toString(). This is to make sure it's always set to a valid value.

albertshau (approved):
one small comment, otherwise lgtm.

gsps1:
rebased commits, merging.

PR #227, opened 2018-09-01
Feature/directive date time

CuriousVini:
**Summary**

Adding support for Date/time in existing apis. Each commit modifies one directive.

### DIFF
--- a/wrangler-core/src/main/java/co/cask/directives/date/DiffDate.java
+++ b/wrangler-core/src/main/java/co/cask/directives/date/DiffDate.java
@@ -1,5 +1,5 @@
 /*
- *  Copyright © 2017 Cask Data, Inc.
+ *  Copyright © 2017-2018 Cask Data, Inc.
  *
  *  Licensed under the Apache License, Version 2.0 (the "License"); you may not
  *  use this file except in compliance with the License. You may obtain a copy of
@@ -30,7 +30,9 @@ import co.cask.wrangler.api.parser.ColumnName;
 import co.cask.wrangler.api.parser.TokenType;
 import co.cask.wrangler.api.parser.UsageDefinition;
 
-import java.util.Date;
+import java.time.ZoneId;
+import java.time.ZoneOffset;
+import java.time.ZonedDateTime;
 import java.util.List;
 
 /**
@@ -46,7 +48,8 @@ public class DiffDate implements Directive {
   private String column1;
   private String column2;
   private String destCol;
-  private final Date date = new Date();
+  // Timestamp now UTC
+  private final ZonedDateTime date = ZonedDateTime.now(ZoneId.ofOffset("UTC", ZoneOffset.UTC));
 
   @Override
   public UsageDefinition define() {
@@ -72,10 +75,10 @@ public class DiffDate implements Directive {
   @Override
   public List<Row> execute(List<Row> rows, ExecutorContext context) throws DirectiveExecutionException {
     for (Row row : rows) {
-      Date date1 = getDate(row, column1);
-      Date date2 = getDate(row, column2);
+      ZonedDateTime date1 = getDate(row, column1);
+      ZonedDateTime date2 = getDate(row, column2);
       if (date1 != null && date2 != null) {
-        row.addOrSet(destCol, date1.getTime() - date2.getTime());
+        row.addOrSet(destCol, date1.toInstant().toEpochMilli() - date2.toInstant().toEpochMilli());
       } else {
         row.addOrSet(destCol, null);
       }
@@ -83,7 +86,7 @@ public class DiffDate implements Directive {
     return rows;
   }
 
-  private Date getDate(Row row, String colName) throws DirectiveExecutionException {
+  private ZonedDateTime getDate(Row row, String colName) throws DirectiveExecutionException {
     // If one of the column contains now, then we return
     // the current date.
     if (colName.equalsIgnoreCase("now")) {
@@ -94,12 +97,12 @@ public class DiffDate implements Directive {
     int idx = row.find(colName);
     if (idx == -1) {
       throw new DirectiveExecutionException(toString() + " : '" +
-                                colName + "' column is not defined in the row.");
+                                              colName + "' column is not defined in the row.");
     }
     Object o = row.getValue(idx);
-    if (o == null || !(o instanceof Date)) {
+    if (o == null || !(o instanceof ZonedDateTime)) {
       return null;
     }
-    return (Date) o;
+    return (ZonedDateTime) o;
   }
 }
--- a/wrangler-core/src/main/java/co/cask/directives/date/FormatDate.java
+++ b/wrangler-core/src/main/java/co/cask/directives/date/FormatDate.java
@@ -1,5 +1,5 @@
 /*
- *  Copyright © 2017 Cask Data, Inc.
+ *  Copyright © 2017-2018 Cask Data, Inc.
  *
  *  Licensed under the Apache License, Version 2.0 (the "License"); you may not
  *  use this file except in compliance with the License. You may obtain a copy of
@@ -31,10 +31,12 @@ import co.cask.wrangler.api.parser.Text;
 import co.cask.wrangler.api.parser.TokenType;
 import co.cask.wrangler.api.parser.UsageDefinition;
 
-import java.text.DateFormat;
-import java.text.SimpleDateFormat;
+import java.time.LocalDate;
+import java.time.ZoneId;
+import java.time.ZoneOffset;
+import java.time.ZonedDateTime;
+import java.time.format.DateTimeFormatter;
 import java.util.ArrayList;
-import java.util.Date;
 import java.util.List;
 
 /**
@@ -48,7 +50,7 @@ public class FormatDate implements Directive {
   public static final String NAME = "format-date";
   private String format;
   private String column;
-  private DateFormat destinationFmt;
+  private DateTimeFormatter destinationFmt;
 
   @Override
   public UsageDefinition define() {
@@ -62,7 +64,7 @@ public class FormatDate implements Directive {
   public void initialize(Arguments args) throws DirectiveParseException {
     this.column = ((ColumnName) args.value("column")).value();
     this.format = ((Text) args.value("format")).value();
-    this.destinationFmt = new SimpleDateFormat(this.format);
+    this.destinationFmt = DateTimeFormatter.ofPattern(this.format);
   }
 
   @Override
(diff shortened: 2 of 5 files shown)


wrangler-core/src/main/java/co/cask/directives/date/DiffDate.java
```diff
@@ -46,7 +46,8 @@
   private String column1;
   private String column2;
   private String destCol;
-  private final Date date = new Date();
+  // Timestamp now in system default timezone
+  private final ZonedDateTime date = ZonedDateTime.now();
```
albertshau: it doesn't look like this is used

wrangler-core/src/main/java/co/cask/directives/date/FormatDate.java
```diff
@@ -76,21 +78,32 @@ public void destroy() {
...
+                                                "Please check the wrangling step."
+        );
+      }
+
+      Object object = row.getValue(idx);
+
+      if (object != null) {
+        try {
+          if (object instanceof LocalDate) {
+            object = ((LocalDate) object).atStartOfDay(ZoneId.systemDefault());
```
albertshau: instead of writing over the value of object, it's cleaner and more readable to have a separate ZonedDateTime variable.

wrangler-core/src/main/java/co/cask/directives/date/FormatDate.java
```diff
@@ -76,21 +78,32 @@ public void destroy() {
...
+          } else if (object instanceof LocalTime) {
+            object = ZonedDateTime.of(LocalDate.of(1970, 1, 1), (LocalTime) object, ZoneId.systemDefault());
+          }
+
+          dt.setValue(idx, destinationFmt.format((ZonedDateTime) object));
+        } catch (ClassCastException e) {
```
albertshau: instead of catching a ClassCastException, remove the try-catch and throw this exception in the else clause where you are checking the object class. Using exceptions to control the flow of a program is brittle and makes everything difficult to read.

wrangler-core/src/main/java/co/cask/directives/date/FormatDate.java
```diff
@@ -76,21 +78,32 @@ public void destroy() {
...
+        throw new DirectiveExecutionException(toString() + " : '" + column + "' column is not defined in the row. " +
+                                                "Please check the wrangling step."
+        );
+      }
+
+      Object object = row.getValue(idx);
+
+      if (object != null) {
+        try {
+          if (object instanceof LocalDate) {
```
albertshau: when is it LocalDate? I thought we were using ZonedDateTime everywhere?

wrangler-core/src/main/java/co/cask/directives/date/FormatDate.java
```diff
@@ -76,21 +78,32 @@ public void destroy() {
...
+            object = ((LocalDate) object).atStartOfDay(ZoneId.systemDefault());
+          } else if (object instanceof LocalTime) {
```
albertshau: it does not make sense to format a time of day as a date. This should result in an exception

wrangler-core/src/main/java/co/cask/directives/parser/ParseDate.java
```diff
@@ -84,13 +83,14 @@ public void destroy() {
       if (idx != -1) {
         Object object = row.getValue(idx);
         if (object instanceof String) {
-          Parser parser = new Parser();
+          TimeZone timeZone = TimeZone.getTimeZone(timezone);
```
albertshau: this should be done in initialize() since it never changes

wrangler-core/src/main/java/co/cask/directives/parser/ParseSimpleDate.java
```diff
@@ -47,7 +49,7 @@
 public class ParseSimpleDate implements Directive {
   public static final String NAME = "parse-as-simple-date";
   private String column;
-  private SimpleDateFormat format;
+  private SimpleDateFormat formatter;
```
albertshau, line 54: why not use the DateTimeFormatter you used in the other directive?

wrangler-core/src/main/java/co/cask/directives/parser/ParseSimpleDate.java
```diff
@@ -78,16 +80,23 @@ public void destroy() {
...
-        if (object == null || object instanceof Date) {
+        if (object == null || object instanceof ZonedDateTime) {
           continue;
         }
         if (object instanceof String) {
           try {
-            Date date = format.parse((String) object);
-            row.setValue(idx, date);
+            // This implementation first creates Date object and then converts it into ZonedDateTime.
+            // DateTimeFormat java 8 api to apply pattern on a string to get ZonedDateTime, throws DateTimeException
```
albertshau: I don't understand what this comment means.

wrangler-core/src/main/java/co/cask/directives/parser/ParseSimpleDate.java
```diff
@@ -78,16 +80,23 @@ public void destroy() {
...
+            // when Zone and Time is not provided in the pattern.
+            Date date = formatter.parse(object.toString());
+
+            // java.util.Date.toString() uses default system timezone, so ZonedDateTime instance is created with the
+            // system default ZoneId
+            ZonedDateTime zonedDateTime = ZonedDateTime.from(date.toInstant().atZone(ZoneId.systemDefault()));
```
albertshau: should use UTC, not system default.

wrangler-core/src/main/java/co/cask/directives/date/DiffDate.java
CuriousVini: Its used here: https://github.com/data-integrations/wrangler/blob/develop/wrangler-core/src/main/java/co/cask/directives/date/DiffDate.java#L89-L91

wrangler-core/src/main/java/co/cask/directives/date/FormatDate.java
albertshau: I got my objects mixed up. LocalDate is for date, ZonedDateTime for timestamp. Shouldn't this directive support formatting timestamps? (java.util.Date is basically a timestamp).

CuriousVini: `FormatDate` can be applied on data read directly from connections like sql database, bigquery, spanner etc. If user wants to format `LocalDate` or `LocalTime` read from the connection, he/she can still do it using this directive.

wrangler-core/src/main/java/co/cask/directives/date/FormatDate.java
albertshau: should always use UTC. From the avro spec, a date is the number of days since the unix epoch.

albertshau: seems like this should be UTC, not the system default. In general, it makes things more predictable if the behavior doesn't change depending on the environment it runs on.

wrangler-core/src/main/java/co/cask/directives/date/FormatDate.java
albertshau: It doesn't matter where the data came from. 

LocalTime doesn't make any sense, it's not tied to any day of the year. ZonedDateTime (timestamp logical type) should be supported.

wrangler-core/src/main/java/co/cask/directives/parser/ParseSimpleDate.java
CuriousVini: Basically to create `ZonedDateTime`,  the pattern/object to be parsed must have time and timezone components. If they are absent in case of pattern `yyyy-mm-dd`, `DateTimeParseException` is thrown.
For example:
```java
@Test
  public void testDate() {
    DateTimeFormatter destinationFmt = DateTimeFormatter.ofPattern("yyyy-mm-dd");
    ZonedDateTime parse = ZonedDateTime.parse("2018-12-21", destinationFmt);
  }
```
Above code throws `DateTimeParseException`. 

I have rephrased the comment with an example.

wrangler-core/src/main/java/co/cask/directives/parser/ParseSimpleDate.java
CuriousVini: This is because `FormatDate` formats `ZonedDateTime` to `String` which does not throw `DateTimeParseException` because the timestamp is already in expected format (has date, time and timezone).

albertshau (approved):
lgtm

CuriousVini:
Squashed the commits.

wrangler-core/src/test/java/co/cask/directives/transformation/ParseDateTest.java
```diff
@@ -121,7 +121,7 @@ public void testSimpleDateWithPatterns() throws Exception {
 
     List<Row> rows = TestingRig.execute(directives, Arrays.asList(row1));
     LocalDate localDate = LocalDate.of(2016, 12, 10);
-    LocalTime zeroTime = LocalTime.of(0, 0);
+    LocalTime zeroTime = LocalTime.of(8, 0);
```
albertshau: this looks like a bug in the directive. Why isn't the date getting parsed as a UTC date?

CuriousVini: This is because `Date` object uses system default zone. https://github.com/data-integrations/wrangler/pull/227/commits/c8ff45182010463411f4e733b0e0e231a0ada298#diff-a337b38e7ac7122bc0e77bc272a912bfR93

CuriousVini: We can set timezone using `formatter.setTimeZone(TimeZone.getTimeZone("UTC"));`. but that would mean whenever there is no timezone specified in the data, the directive will assume it to be UTC. However, existing behavior considers date without timezone to be system default.

albertshau: It has to be consistent across the directives. This unit test will break on any system that is not in the PST timezone.

I think parse-as-simple-date should use UTC. Even though the existing behavior is to use the system default, I don't think that's correct. Does the directive have a way to specify which timezone to use?

albertshau: For example, imagine if the output changed between pipeline runs because the cluster happened to be on the east coast for one run and the west coast for another.

CuriousVini: Right, we can use UTC.`parse-as-simple-date` directive does not have a way to provide timezone. However `parse-as-date` does.

albertshau: yeah I think that makes sense

albertshau (approved):
lgtm.
Can you open a jira to add timezone as an optional argument to the relevant directives?

CuriousVini:
Created https://issues.cask.co/browse/CDAP-14275. Squashed last 2 commits. Merging.

PR #228, opened 2018-09-04
add reference name to spanner plugin specification

gsps1:
Reference name is required by UI while creating pipeline plugin.

### DIFF
--- a/wrangler-service/src/main/java/co/cask/wrangler/service/spanner/SpannerSpecification.java
+++ b/wrangler-service/src/main/java/co/cask/wrangler/service/spanner/SpannerSpecification.java
@@ -34,7 +34,7 @@ public class SpannerSpecification {
 
   SpannerSpecification(String serviceFilePath, String project, String instance,
                        String database, String table, Schema schema) {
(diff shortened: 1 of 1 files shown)


wrangler-service/src/main/java/co/cask/wrangler/service/spanner/SpannerSpecification.java
```diff
@@ -33,6 +34,7 @@
 
   SpannerSpecification(String serviceFilePath, String project, String instance,
                        String database, String table, Schema schema) {
+    this.referenceName = "Spanner_Source";
```
albertshau: this is a bad reference name as it would be the same for any spanner connection created by dataprep. Is this what the other connections do?

A better one would be something like `<database>:<table>`, though you would have to make sure whatever special characters CDAP doesn't allow are replaced.

nitinmotgi: @albertshau comment addressed.

nitinmotgi:
@albertshau please review when you have a chance.

PR #230, opened 2018-09-07
CDAP-13977 change wrangler-core to not be a bundled jar

albertshau:
There is no reason for wrangler-core to be a bundle jar as it is
just a common module for the service and transform. This saves
about 100mb in space, as both the service and transform jars
were bundling all of wrangler-core's dependencies as well as a
fat wrangler-core.


CuriousVini (approved):
LGTM 👍

PR #232, opened 2018-09-07
Fix/expression test

CuriousVini:
**Summary**

- Changes to support date/time in expressions
- Added actual test to test values in `ExpressionTest`
- Added a directive `parse-timestamp-as-date` to convert unix timestamp in millis to ZonedDateTime.
- Added corresponding test for `parse-timestamp-as-date`

https://builds.cask.co/browse/HYP-WT205-9

### DIFF
--- a/wrangler-core/src/main/java/co/cask/directives/parser/ParseTimestamp.java
+++ b/wrangler-core/src/main/java/co/cask/directives/parser/ParseTimestamp.java
@@ -1,0 +1,148 @@
+/*
+ *  Copyright © 2018 Cask Data, Inc.
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ *  use this file except in compliance with the License. You may obtain a copy of
+ *  the License at
+ *
+ *  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ *  WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ *  License for the specific language governing permissions and limitations under
+ *  the License.
+ */
+
+package co.cask.directives.parser;
+
+import co.cask.cdap.api.annotation.Description;
+import co.cask.cdap.api.annotation.Name;
+import co.cask.cdap.api.annotation.Plugin;
+import co.cask.wrangler.api.Arguments;
+import co.cask.wrangler.api.Directive;
+import co.cask.wrangler.api.DirectiveExecutionException;
+import co.cask.wrangler.api.DirectiveParseException;
+import co.cask.wrangler.api.ErrorRowException;
+import co.cask.wrangler.api.ExecutorContext;
+import co.cask.wrangler.api.Optional;
+import co.cask.wrangler.api.Row;
+import co.cask.wrangler.api.annotations.Categories;
+import co.cask.wrangler.api.parser.ColumnName;
+import co.cask.wrangler.api.parser.Text;
+import co.cask.wrangler.api.parser.TokenType;
+import co.cask.wrangler.api.parser.UsageDefinition;
+
+import java.time.Instant;
+import java.time.ZoneId;
+import java.time.ZoneOffset;
+import java.time.ZonedDateTime;
+import java.util.EnumSet;
+import java.util.List;
+import java.util.Set;
+import java.util.concurrent.TimeUnit;
+
+/**
+ * A Executor to parse timestamp as date.
+ */
+@Plugin(type = Directive.Type)
+@Name("parse-timestamp")
+@Categories(categories = {"parser", "date"})
+@Description("Parses column values representing unix timestamp as date.")
+public class ParseTimestamp implements Directive {
+  public static final String NAME = "parse-timestamp";
+  private static final Set<TimeUnit> SUPPORTED_TIME_UNITS = EnumSet.of(TimeUnit.SECONDS, TimeUnit.MILLISECONDS,
+                                                                       TimeUnit.MICROSECONDS);
+  private String column;
+  private TimeUnit timeUnit;
+
+  @Override
+  public UsageDefinition define() {
+    UsageDefinition.Builder builder = UsageDefinition.builder(NAME);
+    builder.define("column", TokenType.COLUMN_NAME);
+    builder.define("timeunit", TokenType.TEXT, Optional.TRUE);
+    return builder.build();
+  }
+
+  @Override
+  public void initialize(Arguments args) throws DirectiveParseException {
+    this.column = ((ColumnName) args.value("column")).value();
+    this.timeUnit = TimeUnit.MILLISECONDS;
+
+    if (args.contains("timeunit")) {
+      String unitValue = ((Text) args.value("timeunit")).value();
+      this.timeUnit = getTimeUnit(unitValue);
+    }
+  }
+
+  @Override
+  public void destroy() {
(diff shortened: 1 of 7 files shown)


wrangler-core/src/test/java/co/cask/wrangler/registry/CompositeDirectiveRegistryTest.java
```diff
@@ -140,7 +140,7 @@ public void testIteratorUsage() throws Exception {
       iterator.next();
       count++;
     }
-    Assert.assertEquals(78, count);
+    Assert.assertEquals(79, count);
```
CuriousVini, line 143: This change is needed because we have added another directive.

wrangler-core/src/main/java/co/cask/directives/parser/ParseTimestamp.java
```diff
@@ -0,0 +1,94 @@
...
+import java.time.ZoneId;
+import java.time.ZoneOffset;
+import java.time.ZonedDateTime;
+import java.util.List;
+
+/**
+ * A Executor to parse timestamp as date.
+ */
+@Plugin(type = Directive.Type)
+@Name("parse-timestamp-as-date")
```
albertshau: I think a better name would just be 'parse-timestamp'

wrangler-core/src/main/java/co/cask/directives/parser/ParseTimestamp.java
```diff
@@ -0,0 +1,94 @@
...
+@Categories(categories = {"parser", "date"})
+@Description("Parses column values representing unix timestamp in millis as date.")
+public class ParseTimestamp implements Directive {
+  public static final String NAME = "parse-timestamp-as-date";
+  private String column;
+
+  @Override
+  public UsageDefinition define() {
+    UsageDefinition.Builder builder = UsageDefinition.builder(NAME);
+    builder.define("column", TokenType.COLUMN_NAME);
```
albertshau, line 62: should we have an optional argument for the unit (seconds, millis, micros)?

wrangler-core/src/test/java/co/cask/directives/transformation/ParseDateTest.java
```diff
@@ -262,4 +262,22 @@ public void testFormatDate() throws Exception {
     Assert.assertEquals("Sat, Dec 10, '16", rows.get(0).getValue("date12"));
     Assert.assertEquals("2016.12.10 AD at 14:45:11 UTC", rows.get(0).getValue("date15"));
   }
+
+  @Test
+  public void testParseTimestamp() throws Exception {
```
albertshau: should be in a separate class ParseTimestampTest

wrangler-core/src/main/java/co/cask/directives/parser/ParseTimestamp.java
```diff
@@ -0,0 +1,94 @@
...
+        // If the data in the cell is null or is already of date format, then
+        // continue to next row.
+        if (object == null || object instanceof ZonedDateTime) {
+          continue;
+        }
+        if (object instanceof Long) {
+          Instant i = Instant.ofEpochMilli((Long) object);
+          ZonedDateTime zonedDateTime = ZonedDateTime.ofInstant(i, ZoneId.ofOffset("UTC", ZoneOffset.UTC));
+          row.setValue(idx, zonedDateTime);
+        } else {
```
albertshau: 'parse' kind of implies it would work on strings too. I think it makes sense to try and parse a string as a long.

wrangler-core/src/main/java/co/cask/directives/parser/ParseTimestamp.java
```diff
@@ -0,0 +1,94 @@
...
+          throw new ErrorRowException(
+            String.format("%s : Invalid type '%s' of column '%s'. Should be of type Long.", toString(),
+                          object != null ? object.getClass().getName() : "null", column), 2
```
albertshau: it's not possible to be null right?

wrangler-core/src/main/java/co/cask/directives/parser/ParseTimestamp.java
CuriousVini: yea, thats a better option.

wrangler-core/src/main/java/co/cask/directives/parser/ParseTimestamp.java
```diff
@@ -0,0 +1,137 @@
...
+    return rows;
+  }
+
+  private static TimeUnit getTimeUnit(String unitValue) throws DirectiveParseException {
+    if (EnumUtils.isValidEnum(TimeUnit.class, unitValue) &&
+      SUPPORTED_TIME_UNITS.contains(TimeUnit.valueOf(unitValue))) {
+      return TimeUnit.valueOf(unitValue);
+    }
+
+    throw new DirectiveParseException(String.format("Provided time unit %s is not supported. Supported time " +
```
albertshau: nit: '%s' is not a supported time unit.

wrangler-core/src/main/java/co/cask/directives/parser/ParseTimestamp.java
```diff
@@ -0,0 +1,137 @@
...
+                                                      "units are %s", unitValue, SUPPORTED_TIME_UNITS));
+  }
+
+  private long getLongValue(Object object) throws ErrorRowException {
+    long longValue;
+    if (object instanceof Long) {
+      longValue = (long) object;
+    } else if (object instanceof String) {
+      longValue = Long.parseLong((String) object);
```
albertshau: if this throws an exception, should we catch it and throw an ErrorRowException?

Is the contract that ErrorRowExceptions are emitted as error records?

wrangler-core/src/main/java/co/cask/directives/parser/ParseTimestamp.java
```diff
@@ -0,0 +1,137 @@
...
+        long longValue = getLongValue(object);
+        ZonedDateTime zonedDateTime = getZonedDateTime(longValue, timeUnit, ZoneId.ofOffset("UTC", ZoneOffset.UTC));
+        row.setValue(idx, zonedDateTime);
+      }
+    }
+    return rows;
+  }
+
+  private static TimeUnit getTimeUnit(String unitValue) throws DirectiveParseException {
+    if (EnumUtils.isValidEnum(TimeUnit.class, unitValue) &&
```
albertshau: simpler do to something like:

```
TimeUnit unit;
try {
  unit = TimeUnit.valueOf(unitValue.toUpperCase());
} catch (IllegalArgumentException e) {
  throw new DirectiveParseException(...);
}

if (SUPPORTED_TIME_UNITS.contains(unit)) {
  throw new DirectiveParseException(...);
}
```

albertshau: nit: I would uppercase unitValue in this method instead of before you pass it to this method. That way, if there is an error, the user sees the exact value that they specified and not the upper-cased version.

albertshau (commented):
minor comments.

For future reference, these should be separated into two PRs. One for adding the parse timestamp directive and one for fixing existing logic.

wrangler-core/src/main/java/co/cask/directives/parser/ParseTimestamp.java
CuriousVini: yes `ErrorRowException` is thrown to emit records to error collector.

albertshau (approved):
lgtm

CuriousVini:
Latest build is green. Squashing and merging.

PR #233, opened 2018-09-13
Emit FLL from wrangler transform

rohitsinha54:
- Emit basic FLL from Wrangler Transform
- See details [here](https://issues.cask.co/browse/CDAP-14005?focusedCommentId=33734&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-33734)

image

[Issue](https://issues.cask.co/browse/CDAP-14005)

### DIFF
--- a/wrangler-transform/src/main/java/co/cask/wrangler/Wrangler.java
+++ b/wrangler-transform/src/main/java/co/cask/wrangler/Wrangler.java
@@ -28,8 +28,11 @@ import co.cask.cdap.api.plugin.PluginProperties;
 import co.cask.cdap.etl.api.Emitter;
 import co.cask.cdap.etl.api.InvalidEntry;
 import co.cask.cdap.etl.api.PipelineConfigurer;
+import co.cask.cdap.etl.api.StageSubmitterContext;
 import co.cask.cdap.etl.api.Transform;
 import co.cask.cdap.etl.api.TransformContext;
+import co.cask.cdap.etl.api.lineage.field.FieldOperation;
+import co.cask.cdap.etl.api.lineage.field.FieldTransformOperation;
 import co.cask.directives.aggregates.DefaultTransientStore;
 import co.cask.wrangler.api.CompileException;
 import co.cask.wrangler.api.CompileStatus;
@@ -64,10 +67,13 @@ import java.net.MalformedURLException;
 import java.net.URI;
 import java.net.URISyntaxException;
 import java.net.URL;
+import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Collections;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Set;
+import java.util.stream.Collectors;
 import javax.annotation.Nullable;
 
 /**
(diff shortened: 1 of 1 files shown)


wrangler-transform/src/main/java/co/cask/wrangler/Wrangler.java
```diff
@@ -56,6 +59,7 @@
 import co.cask.wrangler.registry.SystemDirectiveRegistry;
 import co.cask.wrangler.registry.UserDirectiveRegistry;
 import com.google.common.collect.ImmutableMap;
+import edu.emory.mathcs.backport.java.util.Collections;
```
albertshau: wrong Collections

wrangler-transform/src/main/java/co/cask/wrangler/Wrangler.java
```diff
@@ -233,6 +239,38 @@ public void configurePipeline(PipelineConfigurer configurer) throws IllegalArgum
...
+      outputFields = outputSchema.getFields().stream().map(Schema.Field::getName).collect(Collectors.toList());
+    }
+    FieldOperation dataPrepOperation = new FieldTransformOperation("DataPrep", config.directives, inputFields,
+                                                                   outputFields);
+    context.record(Collections.singletonList(dataPrepOperation));
+  }
+
+  private boolean checkSchema(Schema schema, String name) {
+    if (schema == null) {
+      LOG.debug(String.format("The %s schema is null.", name));
```
albertshau: Log messages should tell somebody reading it everything they need to know without looking at any context. Better to make it:

is null. Field level lineage will not be recorded.

wrangler-transform/src/main/java/co/cask/wrangler/Wrangler.java
```diff
@@ -233,6 +239,38 @@ public void configurePipeline(PipelineConfigurer configurer) throws IllegalArgum
...
+      return false;
+    }
+    if (schema.getFields() == null) {
+      LOG.debug(String.format("The %s schema fields are null.", name));
```
albertshau: is null. Field level lineage will not be recorded.

wrangler-transform/src/main/java/co/cask/wrangler/Wrangler.java
```diff
@@ -233,6 +239,38 @@ public void configurePipeline(PipelineConfigurer configurer) throws IllegalArgum
...
+    if (checkSchema(inputSchema, "input")) {
+      //noinspection ConstantConditions
+      inputFields = inputSchema.getFields().stream().map(Schema.Field::getName).collect(Collectors.toList());
+    }
+    Schema outputSchema = context.getOutputSchema();
+    if (checkSchema(outputSchema, "output")) {
+      //noinspection ConstantConditions
+      outputFields = outputSchema.getFields().stream().map(Schema.Field::getName).collect(Collectors.toList());
+    }
+    FieldOperation dataPrepOperation = new FieldTransformOperation("DataPrep", config.directives, inputFields,
```
albertshau: The other names seem to be verbs, whereas this is a noun. I would call it 'Prepare' or 'Prepare Data' (are spaces allowed?)

albertshau (commented):
from the screenshot I don't see any separators between the directives, which makes it hard to read. Are they stripped somehow?

elfenheart:
Would it be possible to join the list of operations with some kind of a delimiter? Perhaps a `;`? The description is not really readable for wrangler. @bdmogal any input?

rohitsinha54:
@albertshau Addressed comments. 
The separators are added in V2 format I believe so I have to migrate the directive.
image

rohitsinha54:
@elfenheart Please have another look. Added delimiters.

wrangler-transform/src/main/java/co/cask/wrangler/Wrangler.java
```diff
@@ -233,6 +239,40 @@ public void configurePipeline(PipelineConfigurer configurer) throws IllegalArgum
...
+      //noinspection ConstantConditions
+      inputFields = inputSchema.getFields().stream().map(Schema.Field::getName).collect(Collectors.toList());
+    }
+    Schema outputSchema = context.getOutputSchema();
+    if (checkSchema(outputSchema, "output")) {
+      //noinspection ConstantConditions
+      outputFields = outputSchema.getFields().stream().map(Schema.Field::getName).collect(Collectors.toList());
+    }
+    FieldOperation dataPrepOperation = new FieldTransformOperation("Prepare Data",
+                                                                   new MigrateToV2(config.directives).migrate(),
```
albertshau, line 258: I assume this doesn't cause problems if the directives are already in "v2" format? What format is "v1"? It couldn't have been splitting on spaces.

IMO, newlines are the most readable.

rohitsinha54: There is MigrateToV2Test which tests that new format can be passed in so it can handle that.
It does not split on whitespace it sees the directive name and then parses the required thing so it has logic for each and every directive 

https://github.com/data-integrations/wrangler/blob/develop/wrangler-core/src/main/java/co/cask/wrangler/parser/MigrateToV2.java#L67

rohitsinha54: The directives are separated by new line but the UI does not show it.

@elfenheart Will work on fixing that.

albertshau (approved):
lgtm

PR #234, opened 2018-09-14
fixing default connection behavior

gsps1:
When listing connections, check if the configured default connection for ui landing exists in connection list. 

When deleting a connection, check if its configured as a default connection and disallow deleting a connection if it is a default connection.

Reasoning behind this : 

1) We create default connections during service handler initialize, however the initialize gets called on each handler thread initialization, which can happen quite often, hence even if a default connection is deleted, it gets immediately recreated. Also we don't want to store the state of default connection creation/deletion in a table to workaround that, as the admin can make changes to the config to make certain default connections as non-default and at that time we want (and will) to allow those connections to be deletable. 

2) If there was a lifecycle phase such as prepareRun in service that runs only once, it would have been ideal to create the default connections during that phase. the default connection would gets recreated on DataPrep re-initialization if a user has deleted it would have been ideal. However we don't have that option currently.

This trade-off of disallowing users to delete a default connection isn't too bad as if the users doesn't want a default connection, they can contact the administrators about changing the app config for DataPrep and it can be removed from the config.

JIRA : https://issues.cask.co/browse/CDAP-14270

### DIFF
--- a/wrangler-service/src/main/java/co/cask/wrangler/dataset/AbstractTableStore.java
+++ b/wrangler-service/src/main/java/co/cask/wrangler/dataset/AbstractTableStore.java
@@ -165,12 +165,12 @@ public abstract class AbstractTableStore<T> {
   }
 
   /**
-   * Mangles the name.
+   * Get connection id for the given connection name
    *
-   * @param name to be mangled.
-   * @return mangled name.
+   * @param name name of the connection.
+   * @return connection id.
    */
-  protected String mangle(String name) {
+  public String getConnectionId(String name) {
     name = name.trim();
     // Lower case columns
     name = name.toLowerCase();
--- a/wrangler-service/src/main/java/co/cask/wrangler/service/connections/ConnectionService.java
+++ b/wrangler-service/src/main/java/co/cask/wrangler/service/connections/ConnectionService.java
@@ -32,11 +32,14 @@ import com.google.gson.Gson;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import java.net.HttpURLConnection;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 import java.util.Set;
+import javax.annotation.Nullable;
 import javax.ws.rs.DELETE;
 import javax.ws.rs.DefaultValue;
 import javax.ws.rs.GET;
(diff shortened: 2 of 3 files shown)


wrangler-service/src/main/java/co/cask/wrangler/dataset/AbstractTableStore.java
```diff
@@ -170,7 +170,7 @@ protected void putObject(byte[] key, T object) {
    * @param name to be mangled.
    * @return mangled name.
    */
-  protected String mangle(String name) {
```
albertshau: update the javadoc for this too

wrangler-service/src/main/java/co/cask/wrangler/service/connections/ConnectionService.java
```diff
@@ -234,12 +237,33 @@ public void list(HttpServiceRequest request, HttpServiceResponder responder,
...
+  private boolean isDefaultConnection(String connectionId) {
+    List<Connection> defaultConnections = connectionTypeConfig.getConnections();
+    // default connection provided as part of config only contains name and type of connection
+    // since connection id is derived from connection name, we get the connection-id for the default connection name
+    // and check with the input connection id.
+    return defaultConnections.stream()
+      .filter(c -> connectionId.equals(store.getConnectionId(c.getName()))).findFirst().isPresent();
+  }
+
+  private Optional<Connection> getDefaultConnection(List<Connection> connections,
```
albertshau: in this case, I think it's easier to just return null instead of using an Optional, as you end up converting to null anyway.

wrangler-service/src/main/java/co/cask/wrangler/service/connections/ConnectionService.java
```diff
@@ -252,6 +276,11 @@ public void list(HttpServiceRequest request, HttpServiceResponder responder,
   public void delete(HttpServiceRequest request, HttpServiceResponder responder,
                      @PathParam("id") final String id) {
     try {
+      if (isDefaultConnection(id)) {
+        responder.sendError(HttpURLConnection.HTTP_BAD_REQUEST,
```
albertshau: I think it makes more sense for this to be a not authorized error

wrangler-service/src/main/java/co/cask/wrangler/service/connections/ConnectionService.java
```diff
@@ -234,12 +237,33 @@ public void list(HttpServiceRequest request, HttpServiceResponder responder,
...
+      Optional<Connection> defaultConnection = getDefaultConnection(connections, connectionTypeConfig);
+      responder.sendJson(
+        new ConnectionResponse<>(connections,
+                                 defaultConnection.isPresent() ? defaultConnection.get().getName() : null));
     } catch (Exception e) {
       error(responder, e.getMessage());
     }
   }
 
+  private boolean isDefaultConnection(String connectionId) {
```
albertshau: the use of 'default' is confusing because there is an actual single default connection. A better name could be 'canBeDeleted', as that's really what it's being used for.

I also think it's more clear if you just put the logic in the delete method instead of having a separate method, since it's basically a single line and is only called from one place.

wrangler-service/src/main/java/co/cask/wrangler/service/connections/ConnectionService.java
```diff
@@ -252,6 +276,11 @@ public void list(HttpServiceRequest request, HttpServiceResponder responder,
   public void delete(HttpServiceRequest request, HttpServiceResponder responder,
                      @PathParam("id") final String id) {
     try {
+      if (isDefaultConnection(id)) {
+        responder.sendError(HttpURLConnection.HTTP_BAD_REQUEST,
+                            String.format("Cannot delete default connection %s", id));
```
albertshau: default is an ambiguous term here. I think we can say 'Cannot delete admin controlled connection %s.'

albertshau (approved):
lgtm

PR #237, opened 2018-09-20
CDAP-14155 handle bigquery exception while listing datasets

CuriousVini:
Issue: https://issues.cask.co/browse/CDAP-14155
**Summary**
Bigquery error message is too large for UI to display since it returns a json. Instead just extract reason of the exception.

image

### DIFF
--- a/wrangler-service/src/main/java/co/cask/wrangler/service/bigquery/BigQueryService.java
+++ b/wrangler-service/src/main/java/co/cask/wrangler/service/bigquery/BigQueryService.java
@@ -34,6 +34,7 @@ import co.cask.wrangler.service.gcp.GCPUtils;
 import co.cask.wrangler.utils.ObjectSerDe;
 import com.google.api.gax.paging.Page;
 import com.google.cloud.bigquery.BigQuery;
+import com.google.cloud.bigquery.BigQueryException;
 import com.google.cloud.bigquery.Dataset;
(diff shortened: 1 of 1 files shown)


wrangler-service/src/main/java/co/cask/wrangler/service/bigquery/BigQueryService.java
```diff
@@ -162,30 +163,38 @@ public void listTables(HttpServiceRequest request, HttpServiceResponder responde
...
+      JsonObject response = new JsonObject();
+      response.addProperty("status", HttpURLConnection.HTTP_OK);
+      response.addProperty("message", "Success");
+      response.addProperty("count", values.size());
+      response.add("values", values);
+      sendJson(responder, HttpURLConnection.HTTP_OK, response.toString());
+    } catch (BigQueryException e) {
+      if (e.getReason() != null) {
+        // CDAP-14155 - BigQueryException message is too large. Instead just throw reason of the exception
+        throw new Exception(e.getReason());
```
albertshau: nit: I would throw a RuntimeException instead of a checked Exception. Doesn't matter much here in the handler endpoint, but would matter more if we refactor this logic into some other class where it actually belongs.

albertshau (approved):
minor comment, otherwise lgtm

CuriousVini:
Compilation passes locally. So not waiting for the build since there are no test changes.

PR #238, opened 2018-09-24
CDAP-14215 fix error message for set-type directive

CuriousVini:
Make error message more descriptive for `set-type` directive.

image

### DIFF
--- a/wrangler-core/src/main/java/co/cask/directives/column/SetType.java
+++ b/wrangler-core/src/main/java/co/cask/directives/column/SetType.java
@@ -77,10 +77,11 @@ public final class SetType implements Directive {
         }
         try {
           row.setValue(idx, convertType(type, object));
+        } catch (DirectiveExecutionException e) {
+          throw e;
         } catch (Exception e) {
-          throw new DirectiveExecutionException(
-            String.format(toString() + ":" + e.getMessage())
-          );
+          throw new DirectiveExecutionException(String.format("Invalid data: Column '%s' can not be converted to '%s'",
+                                                              col, type));
         }
       }
     }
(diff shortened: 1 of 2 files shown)


albertshau:
this is a better error message but I thought we want to remove the exception name from the message too?

wrangler-service/src/main/java/co/cask/wrangler/service/directive/DirectivesService.java
```diff
@@ -672,6 +673,9 @@ public void execute(HttpServiceRequest request, HttpServiceResponder responder,
     } catch (DirectiveParseException e) {
       LOG.warn(e.getMessage(), e);
       error(responder, e.getMessage());
+    } catch (RecipeException e) {
+      LOG.error(e.getMessage(), e);
```
albertshau: can you open a jira to improve the logging in wrangler? Users errors should not be logged as service errors or warnings. A user running directives on invalid data is a normal operation.

wrangler-service/src/main/java/co/cask/wrangler/service/directive/DirectivesService.java
```diff
@@ -672,6 +673,9 @@ public void execute(HttpServiceRequest request, HttpServiceResponder responder,
...
+      error(responder, e.getCause().getMessage());
```
albertshau: In general, we should avoid unwrapping exceptions like this, as context is lost. For example, if somebody threw a new RecipeException(msg, cause), this would lose the actual message they explicitly set. This also assumes a cause is always present, which does not have to be the case.

If the exception class is getting into the message, it probably because something is throwing new RecipeException(cause) without specifying a message. In that scenario, Throwable will toString() the cause to generate the message. 

All that to say, I suspect the actual problem is that someplace `throw new RecipeException(e)` is being called. If you change it to `throw new RecipeException(e.getMessage(), e)`, I believe it will fix the exception class in the message.

wrangler-service/src/main/java/co/cask/wrangler/service/directive/DirectivesService.java
CuriousVini: Filed: https://issues.cask.co/browse/CDAP-14393

albertshau (approved):
lgtm

CuriousVini:
build: https://builds.cask.co/browse/HYP-WT216-1

CuriousVini:
Build is green. Merging.

PR #240, opened 2018-09-26
CDAP-14372 make directives nullable

albertshau:
If no directives are given, the transform is a no-op.
The main use case for this is if somebody creates a pipeline from
a connection that doesn't actually transform any data. For example,
if somebody wants to write from one dataset table to another
without transformation.


CuriousVini (approved):
lgtm 👍

PR #241, opened 2018-09-27
CDAP-14390 fix gcs json and xml files

albertshau:
The service is using a really half baked system for file types.
There is an enum for Json and XML types, but not way to read
either of those. So anything tagged with those types result in
blank screens.

Removing those types so nobody tries to use them, and switching
their use to TEXT.


CuriousVini (approved):
lgtm 👍

PR #242, opened 2018-09-28
CDAP-14387 fix default connection

albertshau:
Fixed a bug where the default connection would not show up if
the name differed from the id.


CuriousVini (approved):
lgtm 👍

PR #244, opened 2018-10-01
CDAP-14397 Use text content type for files without extension

CuriousVini:
**Summary**
- Uses text content type for files without extension
- Uses tab instead of space in `file.extensions`.

image

image

Build: https://builds.cask.co/browse/HYP-WT219-1

### DIFF
--- a/wrangler-service/src/main/java/co/cask/wrangler/service/FileTypeDetector.java
+++ b/wrangler-service/src/main/java/co/cask/wrangler/service/FileTypeDetector.java
@@ -16,14 +16,11 @@
 
 package co.cask.wrangler.service;
 
-import co.cask.wrangler.service.explorer.Explorer;
 import org.apache.commons.io.FilenameUtils;
 import org.apache.twill.filesystem.Location;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import java.io.File;
-import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.util.HashMap;
 import java.util.Map;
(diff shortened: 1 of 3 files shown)


wrangler-service/src/main/java/co/cask/wrangler/service/FileTypeDetector.java
```diff
@@ -46,7 +46,7 @@ public FileTypeDetector() {
       Scanner scanner = new Scanner(file);
```
chtyim: This is bad. It uses the default charset to decode the file. Please use the one with charset `public Scanner(File source, String charsetName)`

chtyim: Also, can't comment on line 44, but shouldn't do `new File(Explorer.class.getClassLoader().getResource("file.extensions").getFile())`, as that assume the `file.extensions` resource is available in file system, which is not always the case (it is ClassLoader implementation specific). Should use the `Scanner(InputStream source, String charsetName)` constructor instead, by opening the stream from the `URL` returned by the `getResource` method.

```java
try (InputStream is = getClass().getClassLoader().getResource("file.extensions").openStream()) {
  Scanner scanner = new Scanner(is, "UTF-8");
  ...
}
```

wrangler-service/src/main/java/co/cask/wrangler/service/FileTypeDetector.java
```diff
@@ -46,7 +46,7 @@ public FileTypeDetector() {
...
       while(scanner.hasNext()) {
         String line = scanner.nextLine();
-        String[] parts = line.split("[\t+\\s+]");
+        String[] parts = line.split("\\t+");
```
chtyim: Why the change?

albertshau: This is from a comment I made in the JIRA that the file.extensions mapping file uses tabs almost everywhere but spaces in a couple places. Wasn't sure if the ones with spaces were getting ignored or not. 

Maybe just use `\\s+` to be more safe? The previous regex is silly, a tab counts as whitespace.

chtyim: Split by whitespaces `\\s+` seems make more sense.

CuriousVini:
@chtyim Thanks for the review. Addressed comments. Please take another pass when you get a chance.

wrangler-service/src/main/java/co/cask/wrangler/service/FileTypeDetector.java
```diff
@@ -41,19 +39,18 @@
...
+      Scanner scanner = new Scanner(is, "UTF-8");
+      while (scanner.hasNext()) {
         String line = scanner.nextLine();
-        String[] parts = line.split("[\t+\\s+]");
+        String[] parts = line.split("\\s+");
         if (parts.length == 2) {
           extensions.put(parts[0], parts[1]);
         }
       }
       scanner.close();
```
chtyim: The `scanner.close()` will close the `InputStream`. So you can actually do:

```java
try (Scanner scanner = new Scanner(getClass().getClassLoader().....)) {
}
```

Sorry didn't catch that earlier.

chtyim (approved):
LGTM

CuriousVini:
Squashing and merging.

PR #245, opened 2018-10-02
Fix S3Service to limit number of files in the bucket to 10k

CuriousVini:
image

image

### DIFF
--- a/wrangler-service/src/main/java/co/cask/wrangler/service/s3/S3Service.java
+++ b/wrangler-service/src/main/java/co/cask/wrangler/service/s3/S3Service.java
@@ -77,6 +77,7 @@ import java.util.HashMap;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
+import javax.ws.rs.DefaultValue;
 import javax.ws.rs.GET;
 import javax.ws.rs.POST;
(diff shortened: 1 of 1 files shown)


wrangler-service/src/main/java/co/cask/wrangler/service/s3/S3Service.java
```diff
@@ -239,18 +241,26 @@ public void run(DatasetContext datasetContext) throws Exception {
       listObjectsRequest.setDelimiter("/");
       ObjectListing result;
       DirectoryListing listing = new DirectoryListing();
+      boolean limitExceeded = false;
       do {
+        if (listing.size() == BUCKET_LIMIT) {
```
albertshau: I think it's better to make this a query param with a default value instead of hardcoded. It's not much more work and offers a possibility of the UI tweaking it if needed.

albertshau (approved):
minor comment otherwise lgtm. fix and merge

wrangler-service/src/main/java/co/cask/wrangler/service/s3/S3Service.java
CuriousVini: Yea I agree. fixed.

PR #246, opened 2018-10-02
CDAP-14355 Improve reference name

rohitsinha54:
Issue: https://issues.cask.co/browse/CDAP-14355

## Summary
- Improve the referenceName which gets filled in by Wrangler. 
- Ideally we will like these to be unique although gurnating uniqueness in all cases is hard. For example in case of GCS the complete path to a file should be used as reference but the path can be long leading to a long reference name which is not user friendly so we try the best. The platform should fail the deployment if same referenceName is used for different sources.
- The referenceNames might consists of special characters which is not supported in DatasetId. Wrangler does not do any filtering for it. CDAP will fail the deployment if the referenceName consists of invalid chars.(This will come in another PR).

### DIFF
--- a/wrangler-service/src/main/java/co/cask/wrangler/service/bigquery/BigQueryService.java
+++ b/wrangler-service/src/main/java/co/cask/wrangler/service/bigquery/BigQueryService.java
@@ -309,8 +309,10 @@ public class BigQueryService extends AbstractWranglerService {
 
       Map<String, String> properties = new HashMap<>();
       JsonObject value = new JsonObject();
-
+      String externalDatasetName =
+        new StringJoiner(".").add(config.get(DATASET_ID)).add(config.get(TABLE_ID)).toString();
       JsonObject bigQuery = new JsonObject();
+      properties.put("referenceName", externalDatasetName);
       properties.put("serviceFilePath", config.get(GCPUtils.SERVICE_ACCOUNT_KEYFILE));
       properties.put("bucket", config.get(BUCKET));
       properties.put("project", config.get(GCPUtils.PROJECT_ID));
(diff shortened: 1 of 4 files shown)


wrangler-service/src/main/java/co/cask/wrangler/service/bigquery/BigQueryService.java
```diff
@@ -52,6 +52,7 @@
 import com.google.cloud.bigquery.TableId;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Charsets;
+import com.google.common.base.Joiner;
```
albertshau: unused?

wrangler-service/src/main/java/co/cask/wrangler/service/gcs/GCSService.java
```diff
@@ -450,19 +452,18 @@ public void specification(HttpServiceRequest request, final HttpServiceResponder
 
       Map<String, String> config = ws.getProperties(workspaceId);
       String pluginType = config.get(PropertyIds.PLUGIN_TYPE);
-      String bucket = config.get("bucket");
       String uri = config.get(PropertyIds.URI);
-      String file = config.get(PropertyIds.FILE_NAME);
+      String filename = uri.substring(uri.lastIndexOf("/") + 1);
```
albertshau: should handle the case when the uri ends in '/'.

wrangler-service/src/main/java/co/cask/wrangler/service/spanner/SpannerService.java
```diff
@@ -202,8 +203,10 @@ public void readTable(HttpServiceRequest request, HttpServiceResponder responder
       String projectId = connectionProperties.get(GCPUtils.PROJECT_ID);
       String path = connectionProperties.get(GCPUtils.SERVICE_ACCOUNT_KEYFILE);
 
+      String externalDsName = new StringJoiner(".").add(databaseId).add(tableId).toString();
```
albertshau: should the instance id be in there as well?

albertshau (approved):
lgtm

PR #248, opened 2018-10-03
CDAP-12852 hack to copy headers when parse-as-csv is used

albertshau:
Added a hacky workaround that tells pipeline sources to copy
headers if the parse-as-csv directive is used with that directive.

### DIFF
--- a/wrangler-service/src/main/java/co/cask/wrangler/service/common/AbstractWranglerService.java
+++ b/wrangler-service/src/main/java/co/cask/wrangler/service/common/AbstractWranglerService.java
@@ -22,12 +22,20 @@ import co.cask.cdap.api.service.http.AbstractHttpServiceHandler;
 import co.cask.cdap.api.service.http.HttpServiceContext;
 import co.cask.wrangler.DataPrep;
 import co.cask.wrangler.dataset.connections.ConnectionStore;
+import co.cask.wrangler.dataset.workspace.DataType;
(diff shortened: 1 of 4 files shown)


wrangler-service/src/main/java/co/cask/wrangler/service/common/AbstractWranglerService.java
```diff
@@ -41,4 +51,28 @@ public void initialize(HttpServiceContext context) throws Exception {
...
+    }
+    String data = ws.getData(workspaceId, WorkspaceDataset.REQUEST_COL, DataType.TEXT);
+    if (data != null) {
+      RequestV1 workspaceReq = GSON.fromJson(data, RequestV1.class);
+      Recipe recipe = workspaceReq.getRecipe();
+      List<String> directives = recipe.getDirectives();
+      // yes this is really hacky, but there doesn't seem to be a good way to get the actual directive classes
+      for (String directive : directives) {
+        directive = directive.trim();
+        if (directive.startsWith("parse-as-csv") && directive.endsWith("true")) {
```
CuriousVini: would lamda be more readable? 
```java
boolean copyHeader = false;
`copyHeader = list.stream().anyMatch(str -> str.trim().startsWith("parse-as-csv") && str.trim().endsWith("true"));`

return copyHeader;```

CuriousVini (approved):
just one minor comment. Rest LGTM 👍

PR #249, opened 2018-10-03
CDAP-14370 reduce log level when dataprep service is not available

albertshau:
This is normal in cloud environments and should not be a warning.
Also making the message more clear, though I'm not sure what
aliasing and restrictions are.

### DIFF
--- a/wrangler-transform/src/main/java/co/cask/wrangler/Wrangler.java
+++ b/wrangler-transform/src/main/java/co/cask/wrangler/Wrangler.java
@@ -342,10 +342,9 @@ public class Wrangler extends Transform<StructuredRecord, StructuredRecord> {
         ConfigDirectiveContext dContext = new ConfigDirectiveContext(url);
         recipe.initialize(dContext);
       } else {
-        LOG.warn(
-          String.format("Stage:%s - Context is set to default, no aliasing and restriction would be applied.",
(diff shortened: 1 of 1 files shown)


wrangler-transform/src/main/java/co/cask/wrangler/Wrangler.java
```diff
@@ -342,10 +342,9 @@ public void initialize(TransformContext context) throws Exception {
         ConfigDirectiveContext dContext = new ConfigDirectiveContext(url);
         recipe.initialize(dContext);
       } else {
-        LOG.warn(
-          String.format("Stage:%s - Context is set to default, no aliasing and restriction would be applied.",
-                        getContext().getStageName())
-          );
+        // this is normal in cloud environments
```
CuriousVini, line 345: If it is normal then why do we need to log it?

albertshau: because it's not normal if you run with the native profile. I'm not actually sure what aliasing and restriction are, but I would imagine it's good to know if you're using those features.

CuriousVini: I see.

CuriousVini (approved):
lgtm

PR #252, opened 2018-10-03
Add bucket limit to GCS service

CuriousVini:
image

### DIFF
--- a/wrangler-service/src/main/java/co/cask/wrangler/service/gcs/GCSService.java
+++ b/wrangler-service/src/main/java/co/cask/wrangler/service/gcs/GCSService.java
@@ -38,7 +38,6 @@ import co.cask.wrangler.service.common.AbstractWranglerService;
 import co.cask.wrangler.service.connections.ConnectionType;
 import co.cask.wrangler.service.gcp.GCPUtils;
 import co.cask.wrangler.utils.ObjectSerDe;
-import com.google.api.client.googleapis.auth.oauth2.GoogleCredential;
 import com.google.api.gax.paging.Page;
 import com.google.cloud.ReadChannel;
 import com.google.cloud.storage.Acl;
@@ -60,14 +59,12 @@ import java.io.ByteArrayOutputStream;
 import java.io.File;
 import java.io.IOException;
 import java.net.HttpURLConnection;
-import java.net.URI;
 import java.nio.ByteBuffer;
 import java.nio.channels.Channels;
 import java.nio.channels.WritableByteChannel;
 import java.security.Security;
 import java.util.ArrayList;
 import java.util.Arrays;
-import java.util.Collection;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Iterator;
@@ -76,6 +73,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.StringJoiner;
+import javax.ws.rs.DefaultValue;
 import javax.ws.rs.GET;
 import javax.ws.rs.POST;
 import javax.ws.rs.Path;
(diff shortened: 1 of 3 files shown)


wrangler-service/src/main/java/co/cask/wrangler/service/gcs/GCSService.java
```diff
@@ -154,8 +153,9 @@ private boolean validateConnection(String connectionId, Connection connection,
   @GET
   @Path("/connections/{connection-id}/gcs/explore")
   public void exploreGCS(HttpServiceRequest request, HttpServiceResponder responder,
-                              @PathParam("connection-id") final String connectionId,
-                              @QueryParam("path") String path) {
+                         @PathParam("connection-id") final String connectionId,
+                         @QueryParam("path") String path,
+                         @QueryParam("bucket-limit") @DefaultValue("1000") int bucketLimit) {
```
albertshau: didn't notice before, but I think this should be named 'limit' and not 'bucket-limit' for both endpoints

wrangler-service/src/main/java/co/cask/wrangler/service/gcs/GCSService.java
```diff
@@ -510,12 +511,12 @@ public void specification(HttpServiceRequest request, final HttpServiceResponder
...
   }
 
-  private Collection<Bucket> getBuckets(Storage storage, Set<String> whitelist) {
+  private Collection<Bucket> getBuckets(Storage storage, Set<String> whitelist, int bucketLimit) {
     // this will include buckets that can be listed by the service account, but may not include all buckets
     // in the whitelist, if the whitelist contains publicly accessible buckets from other projects.
     // do some post-processing to filter out anything not in the whitelist and also try and lookup buckets
     // that are in the whitelist but not in the returned list
-    Page<Bucket> list = storage.list();
+    Page<Bucket> list = storage.list(Storage.BucketListOption.pageSize(bucketLimit));
```
albertshau: this sounds like it's a page size limit, not a limit for the entire response. list.iterateAll() will iterate through every page. There is another method that iterates through just the elements in the page.

Is there a limit for how big the requested page size can be? If not, we just need to change the iterateAll() to the other one. If there is, we need to go page by page and limit it ourselves.

wrangler-service/src/main/java/co/cask/wrangler/service/gcs/GCSService.java
```diff
@@ -275,12 +280,19 @@ public void run(DatasetContext datasetContext) throws Exception {
           object.addProperty("wrangle", isWrangeable);
         }
         values.add(object);
+        if (values.size() >= bucketLimit) {
```
CuriousVini: @albertshau should we use a separate query param to limit number of objects in a bucket?

wrangler-service/src/main/java/co/cask/wrangler/service/gcs/GCSService.java
```diff
@@ -154,8 +152,9 @@ private boolean validateConnection(String connectionId, Connection connection,
   @GET
   @Path("/connections/{connection-id}/gcs/explore")
   public void exploreGCS(HttpServiceRequest request, HttpServiceResponder responder,
-                              @PathParam("connection-id") final String connectionId,
-                              @QueryParam("path") String path) {
+                         @PathParam("connection-id") final String connectionId,
+                         @QueryParam("path") String path,
+                         @QueryParam("limit") @DefaultValue("1000") int bucketLimit) {
```
albertshau: nit: this is more like an objectLimit, as it returns more than just buckets.

albertshau (approved):
one nit about renaming a variable. fix and merge

wrangler-service/src/main/java/co/cask/wrangler/service/gcs/GCSService.java
albertshau: no i think its fine to have a single limit, as it's a single API.

PR #253, opened 2018-10-04
Combination of send-to-error-and-continue and send-to-error 

nitinmotgi:
Now correctly handles the use case of subjecting a record to multiple checks before excluding it from the processing. 

```
send-to-error-and-continue exp:{ x == 1 } 'Bad value of x'
send-to-error-and-continue exp:{ name == 'XYZ'} 'Incorrect name'
send-to-error exp:{ dq_failure >= 1}
```

Would check for conditions before emitting it on error port when at-least one of the send-to-error-and-continue condition is hit.

### DIFF
--- a/wrangler-core/src/main/java/co/cask/directives/aggregates/DefaultTransientStore.java
+++ b/wrangler-core/src/main/java/co/cask/directives/aggregates/DefaultTransientStore.java
@@ -16,6 +16,7 @@
 
 package co.cask.directives.aggregates;
 
+import co.cask.wrangler.api.TransientVariableScope;
 import co.cask.wrangler.api.TransientStore;
 
 import java.util.HashMap;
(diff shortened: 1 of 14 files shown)


wrangler-core/src/main/java/co/cask/directives/aggregates/DefaultTransientStore.java
```diff
@@ -41,12 +43,23 @@
    * @param value associated with the variable.
    */
   @Override
-  public void increment(String name, long value) {
-    Long count = get(name);
+  public void increment(TransientVariableScope scope, String name, long value) {
+    if(scope == TransientVariableScope.GLOBAL) {
```
sreevatsanraman: space after if. otherwise will fail checkstyle. Would be good to add checkstyle from cdap repo if you don't have it configured already.

sreevatsanraman (approved):

PR #254, opened 2018-10-05
CDAP-14470 remove offset from schema of file based sources

albertshau:
Directives like the set columns directive break if there is more
than a single input field. That's an indication of bad design,
but in the meantime, changing the sources to only emit a single
field to workaround the issue.


nitinmotgi (approved):
LGTM 👍

PR #255, opened 2018-10-05
[CDAP-14478] Fixes creating workspace when uploaded directly to explorer service

ajainarayanan:
JIRA: https://issues.cask.co/browse/CDAP-14478


albertshau (approved):
lgtm.

PR #262, opened 2018-11-01
CDAP-14542 set the dataset project property for bq source

albertshau:
Properly handle the case when creating a pipeline where the
bigquery dataset is not in the same project as the connection


wrangler-service/src/main/java/co/cask/wrangler/service/bigquery/BigQueryService.java
```diff
@@ -316,6 +319,7 @@ public void specification(HttpServiceRequest request, final HttpServiceResponder
       properties.put("serviceFilePath", config.get(GCPUtils.SERVICE_ACCOUNT_KEYFILE));
       properties.put("bucket", config.get(BUCKET));
       properties.put("project", config.get(GCPUtils.PROJECT_ID));
+      properties.put("datasetProject", config.get(DATASET_PROJECT));
```
rohitsinha54: we can use the constant here

rohitsinha54 (approved):

PR #280, opened 2018-12-11
Feature/merge rel32 to dev

rohitsinha54:


albertshau:
There are no differences, does this need to be done?

rohitsinha54:
It is nice to have. Since it will create the commit history for the release version on develop. So later if we merge something on release or bump to snapshot on release again we will not have conflict due to this commit.

rohitsinha54:
Also it keeps the release and develop commits history tree in sync which I think we should have.

albertshau:
As we discussed in person, please open another PR to bump the version on release after this is merged.

albertshau (approved):

albertshau (approved):

PR #285, opened 2018-12-17
Fix BigQuery service to use DATETIME type as string

CuriousVini:
BigQuery `DATETIME` field type is time without timezone. BigQuery apis returns this value as string in ISO 8601 format. However, Bigquery service assumes it to be timestamp instead of string.
Also fixing Natural Date directive to not throw error when column value is null (similar to Simple Date directive).

### DIFF
--- a/wrangler-service/src/main/java/co/cask/wrangler/service/bigquery/BigQueryService.java
+++ b/wrangler-service/src/main/java/co/cask/wrangler/service/bigquery/BigQueryService.java
@@ -387,12 +387,12 @@ public class BigQueryService extends AbstractWranglerService {
             row.add(fieldName, LocalDate.parse(fieldValue.getStringValue()));
             break;
 
-          case DATETIME:
           case TIMESTAMP:
             long tsMicroValue = fieldValue.getTimestampValue();
             row.add(fieldName, getZonedDateTime(tsMicroValue));
             break;
 
+          case DATETIME:
           case STRING:
             row.add(fieldName, fieldValue.getStringValue());
             break;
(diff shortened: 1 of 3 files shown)


wrangler-service/src/main/java/co/cask/wrangler/service/bigquery/BigQueryService.java
```diff
@@ -387,12 +387,12 @@ public void specification(HttpServiceRequest request, final HttpServiceResponder
             row.add(fieldName, LocalDate.parse(fieldValue.getStringValue()));
             break;
 
-          case DATETIME:
           case TIMESTAMP:
             long tsMicroValue = fieldValue.getTimestampValue();
             row.add(fieldName, getZonedDateTime(tsMicroValue));
             break;
 
+          case DATETIME:
```
albertshau, line 395: If it's always ISO 8601, should we convert to a timestamp automatically?

If somebody wants to write to a datetime column using the bigquery sink, how does it work?

CuriousVini: We can convert string to ZonedDateTime. However, `DATETIME` does not have timezone information. So we will have to assume that the value is in UTC or ask user to enter timezone information probably in bigquery connection config.

Another option is to keep the type of the value as is and let the user provide timezone information using Natural Date directive. 
image

In plugins, we do not have a notion of timestamp without time zone. Thats why sink only supports DATE, TIME and TIMESTAMP bigquery types.
https://github.com/data-integrations/google-cloud/blob/release/0.11/src/main/java/co/cask/gcp/bigquery/BigQuerySink.java#L189

albertshau: I see. The approach in this PR seems fine to me then

albertshau (approved):
lgtm

PR #287, opened 2018-12-18
Fix BigQuery service to use DATETIME type as string

CuriousVini:
cherry pick from https://github.com/data-integrations/wrangler/pull/285


albertshau (approved):
lgtm

PR #288, opened 2018-12-19
CDAP-14619 refactor project to allow running dataset tests

albertshau:
Created two new modules that have minimal dependencies.
The first is wrangler-proto, which contains all the proto classes
used in requests and responses in the REST endpoints.
The second is wrangler-storage, which contains all storage related
classes.

Besides having a cleaner logical separation, this allows the
storage classes to be tested with CDAP TestBase since that module
no longer depends on a clashing version of guava.

Added a couple basic test cases for the ConnectionStore and
SchemaRegistry.


CuriousVini (approved):
Thanks for adding tests for ConnectionStore and SchemaRegistry. Code changes look good, lgtm 👍

PR #290, opened 2018-12-21
CDAP-14619 use a separate table for schema registry

albertshau:
Instead of sharing the same table between the connection store and
schema registry, use separate tables. This is to prepare for the
switch to the system table SPI, where each table has its own
set schema.


CuriousVini (approved):
lgtm

PR #291, opened 2019-01-02
CDAP-14709 fix read logic for GCS connections

albertshau:
Fixed an issue where GCS objects larger than 2gb would cause an
error due to casting that length to an integer and using it to
allocate a byte array, resulting in a byte array of negative size.

Also fixed a bug in reading GCS blob content that caused data to
be jumbled.

### DIFF
--- a/wrangler-service/src/main/java/co/cask/wrangler/service/gcs/GCSService.java
+++ b/wrangler-service/src/main/java/co/cask/wrangler/service/gcs/GCSService.java
@@ -89,7 +89,7 @@ import static co.cask.wrangler.ServiceUtils.sendJson;
  */
 public class GCSService extends AbstractWranglerService {
   private static final Logger LOG = LoggerFactory.getLogger(GCSService.class);
-  static final int FILE_SIZE = 10 * 1024 * 1024;
+  static final long FILE_SIZE = 10 * 1024 * 1024;
   private FileTypeDetector detector;
 
   @Override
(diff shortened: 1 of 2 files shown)


wrangler-service/src/main/java/co/cask/wrangler/service/gcs/GCSService.java
```diff
@@ -387,8 +388,10 @@ public void loadObject(HttpServiceRequest request, HttpServiceResponder responde
           }
 
           List<Row> rows = new ArrayList<>();
-          for (String line : lines) {
-            rows.add(new Row("body", line));
+          // if the content was truncated, ignore the last line because it's probably not complete.
```
elfenheart, line 391: do we need to worry if the file actually has large number of columns, that the partial data in the blob is actually still within the same row?

albertshau: this only truncates if there is more than 1 row. I suppose this may look funny if there were one and a half rows or something, though I think we'll have other issues if somebody has a 5mb row.

CuriousVini (commented):
Shouldn't UI receive a flag if data is truncated in the response? Also is it possible to add a unit test for this case?

albertshau:
I think the intent was everything is assumed to be a sample so there's nothing indicating something has been "truncated". It's not really truncated, we just decided to sample one fewer row.

Unit testing this is difficult without reading from GCS or doing some more refactoring around how wrangler reads data. I didn't want to consolidate that logic in this PR, as wrangler is actually inconsistent in how it handles data between GCS, S3, and the filesystem.

CuriousVini:
I see, I thought `GCSServiceTest` has unit tests for `GCSService` related functionality. But seems like all the tests there are ignored.

CuriousVini (approved):
lgtm 👍

PR #292, opened 2019-01-03
CDAP-14709 fix read logic for GCS connections

albertshau:
Fixed an issue where GCS objects larger than 2gb would cause an
error due to casting that length to an integer and using it to
allocate a byte array, resulting in a byte array of negative size.

Also fixed a bug in reading GCS blob content that caused data to
be jumbled.


albertshau:
cherry-pick of #291

CuriousVini (approved):
lgtm 👍

PR #293, opened 2019-01-05
CDAP-14619 clean up directives service

albertshau:
Replaced usages of JsonObject with proper java classes to make it
clearer what the API is, make the code less error prone, and make
it easier to maintain.


wrangler-core/src/main/java/co/cask/wrangler/registry/DirectiveInfo.java
```diff
@@ -1,28 +1,27 @@
 /*
- *  Copyright © 2017 Cask Data, Inc.
+ * Copyright © 2018 Cask Data, Inc.
```
CuriousVini: license year should be 2019 for all new classes

wrangler-proto/src/main/java/co/cask/wrangler/proto/workspace/ColumnStatistics.java
```diff
@@ -0,0 +1,41 @@
...
+public class ColumnStatistics {
+  private final Map<String, Float> general;
+  private final Map<String, Float> types;
+
+  public ColumnStatistics(@Nullable Map<String, Float> general, @Nullable Map<String, Float> types) {
+    this.general = general;
+    this.types = types;
+  }
+
+  public Map<String, Float> getGeneral() {
```
CuriousVini, line 35: `@Nullable`

wrangler-proto/src/main/java/co/cask/wrangler/proto/workspace/ColumnStatistics.java
```diff
@@ -0,0 +1,41 @@
...
+    return general;
+  }
+
+  public Map<String, Float> getTypes() {
```
CuriousVini, line 40: `@Nullable`

wrangler-core/src/main/java/co/cask/wrangler/registry/DirectiveInfo.java
```diff
@@ -111,21 +114,4 @@ public UsageDefinition definition() {
...
     return (Directive) directive.newInstance();
   }
-
-  public JsonObject toJson() {
-    JsonObject response = new JsonObject();
-    response.addProperty("plugin", name);
-    response.addProperty("usage", usage);
-    response.addProperty("description", description);
-    response.addProperty("class", directive.getCanonicalName());
-    response.add("definition", definition.toJson());
```
CuriousVini, line 121: This did not include `optionalCnt` before. Should it be transient?

wrangler-service/src/main/java/co/cask/wrangler/service/directive/DirectivesService.java
```diff
@@ -594,38 +594,34 @@ public void execute(HttpServiceRequest request, HttpServiceResponder responder,
         }
       });
 
-      JsonArray values = new JsonArray();
-      JsonArray headers = new JsonArray();
-      JsonObject types = new JsonObject();
-      Set<String> header = new HashSet<>();
+      List<Map<String, Object>> values = new ArrayList<>(rows.size());
+      Map<String, String> types = new HashMap<>();
+      Set<String> headers = new LinkedHashSet<>();
```
CuriousVini, line 599: why is it changed to LinkedHashSet?

wrangler-service/src/main/java/co/cask/wrangler/service/directive/DirectivesService.java
```diff
@@ -742,68 +728,50 @@ public void summary(HttpServiceRequest request, HttpServiceResponder responder,
...
       for (String name : uniqueColumns) {
-        JsonObject columnResult = new JsonObject();
         try {
           validator.validate(name);
-          columnResult.addProperty("valid", true);
+          columnValidationResults.put(name, new ColumnValidationResult(true, null));
         } catch (ValidatorException e) {
-          columnResult.addProperty("valid", false);
-          columnResult.addProperty("message", e.getMessage());
+          columnValidationResults.put(name, new ColumnValidationResult(false, e.getMessage()));
```
CuriousVini: does the constructor need 2 args? can `valid` be derived from `message`?

wrangler-service/src/main/java/co/cask/wrangler/service/directive/DirectivesService.java
```diff
@@ -742,68 +728,50 @@ public void summary(HttpServiceRequest request, HttpServiceResponder responder,
...
-        JsonObject v = new JsonObject();
-        JsonObject o = new JsonObject();
+        Map<String, Float> generalStats = new HashMap<>();
         for (Pair<String, Double> value : values) {
-          o.addProperty(value.getFirst(), value.getSecond().floatValue() * 100);
+          generalStats.put(value.getFirst(), value.getSecond().floatValue() * 100);
         }
-        v.add("general", o);
-        statistics.add(field.getFirst(), v);
+        ColumnStatistics columnStatistics = new ColumnStatistics(generalStats, null);
```
CuriousVini, line 756: can we have another constructor which just takes `generalStats`?

wrangler-service/src/main/java/co/cask/wrangler/service/directive/DirectivesService.java
```diff
@@ -1002,37 +934,26 @@ public void artifacts(HttpServiceRequest request, HttpServiceResponder responder
...
-        object.addProperty("scope", artifact.getScope().name());
+        DirectiveArtifact directiveArtifact = new DirectiveArtifact(artifact);
         for (PluginClass plugin : plugins) {
           if (Directive.TYPE.equalsIgnoreCase(plugin.getType())) {
-            JsonObject directive = new JsonObject();
-            directive.addProperty("name", plugin.getName());
-            directive.addProperty("description", plugin.getDescription());
-            directive.addProperty("type", plugin.getType());
-            directive.addProperty("class", plugin.getClassName());
-            directive.add("artifact", object);
```
CuriousVini, line 1026: this does not include `properties` right?

wrangler-service/src/main/java/co/cask/wrangler/service/directive/DirectivesService.java
```diff
@@ -914,48 +872,31 @@ public void usage(HttpServiceRequest request, HttpServiceResponder responder) {
...
-        values.add(usage);
+        DirectiveUsage directiveUsage = new DirectiveUsage(directive.name(), directive.usage(), directive.description(),
+                                                           config.isExcluded(directive.name()), false,
+                                                           directive.scope().name(), directive.definition(),
+                                                           directive.categories());
+        values.add(directiveUsage);
 
         // For this directive we find all aliases and add them to the
         // description.
         if (aliases.containsKey(directive)) {
```
CuriousVini: Is this comparison correct? `aliases` is a map of `String, List<String>`

CuriousVini (commented):
few comments

wrangler-service/src/main/java/co/cask/wrangler/service/directive/DirectivesService.java
albertshau: The column headers need to preserve order, so using a LinkedHashSet. The previous implementation using a Set in combination with a JsonArray was essentially implementing the relevant LinkedHashSet behavior.

wrangler-core/src/main/java/co/cask/wrangler/registry/DirectiveInfo.java
albertshau: ah yes it's 2019 now.

wrangler-core/src/main/java/co/cask/wrangler/registry/DirectiveInfo.java
albertshau: done

wrangler-proto/src/main/java/co/cask/wrangler/proto/workspace/ColumnStatistics.java
albertshau: done

wrangler-proto/src/main/java/co/cask/wrangler/proto/workspace/ColumnStatistics.java
albertshau: done

wrangler-service/src/main/java/co/cask/wrangler/service/directive/DirectivesService.java
albertshau: done

wrangler-service/src/main/java/co/cask/wrangler/service/directive/DirectivesService.java
albertshau: can't because they would have the same signature. Could have a builder but that seemed like overkill for 2 fields.

wrangler-service/src/main/java/co/cask/wrangler/service/directive/DirectivesService.java
albertshau: Good catch, this will always return false. This is an existing bug but I can fix it in this PR

wrangler-service/src/main/java/co/cask/wrangler/service/directive/DirectivesService.java
albertshau: good catch, added another constructor that doesn't set properties.

CuriousVini (approved):
lgtm 👍

PR #294, opened 2019-01-09
CDAP-14619 refactor workspace dataset

albertshau:
Refactored the workspace dataset so that logic is encapsulated
inside the class instead of leaked outside. This mostly amounts
to removing places where the DirectivesService would directly
access the underlying Table instead of having an abstraction layer
for updating and reading various workspace information.

Also removed useless exception wrapping and added better error
handling to avoid a bunch of possible null pointer exceptions
and to correctly check whether the workspace exists.

Also added unit tests for the WorkspaceDataset. This revealed a
bug where the method to delete workspaces within a scope actually
deleted every workspace that was not in that scope. Fixed the bug.

### DIFF
--- a/wrangler-proto/src/main/java/co/cask/wrangler/proto/Recipe.java
+++ b/wrangler-proto/src/main/java/co/cask/wrangler/proto/Recipe.java
@@ -1,5 +1,5 @@
 /*
- * Copyright © 2017 Cask Data, Inc.
+ * Copyright © 2017-2019 Cask Data, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not
  * use this file except in compliance with the License. You may obtain a copy of
@@ -17,19 +17,20 @@
 package co.cask.wrangler.proto;
 
 import java.util.List;
+import java.util.Objects;
 
 /**
  * Defines the recipe object that is part of the {@link Request}
  */
 public class Recipe {
   // RecipeParser specified by the user
-  private List<String> directives;
+  private final List<String> directives;
 
   // Ability to save the directives
-  private Boolean save;
+  private final Boolean save;
 
   // Name of the recipe.
-  private String name;
+  private final String name;
 
   public Recipe(List<String> directives, Boolean save, String name) {
     this.directives = directives;
@@ -70,5 +71,24 @@ public class Recipe {
    */
   public String getName() {
     return name;
+  }
+
+  @Override
+  public boolean equals(Object o) {
+    if (this == o) {
+      return true;
+    }
+    if (o == null || getClass() != o.getClass()) {
+      return false;
+    }
+    Recipe recipe = (Recipe) o;
+    return Objects.equals(directives, recipe.directives) &&
+      Objects.equals(save, recipe.save) &&
+      Objects.equals(name, recipe.name);
+  }
+
+  @Override
+  public int hashCode() {
+    return Objects.hash(directives, save, name);
   }
 }
--- a/wrangler-proto/src/main/java/co/cask/wrangler/proto/RequestV1.java
+++ b/wrangler-proto/src/main/java/co/cask/wrangler/proto/RequestV1.java
@@ -1,5 +1,5 @@
 /*
- * Copyright © 2017 Cask Data, Inc.
+ * Copyright © 2017-2019 Cask Data, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not
  * use this file except in compliance with the License. You may obtain a copy of
@@ -18,28 +18,29 @@ package co.cask.wrangler.proto;
 
 import com.google.gson.JsonObject;
 
+import java.util.Objects;
+
 /**
  * Specifies the V1 version of the {@link Request} object.
  */
 public final class RequestV1 implements Request {
   // Version of request.
-  private int version;
+  private final int version;
 
   // Workspace information associated with request.
-  private Workspace workspace;
+  private final Workspace workspace;
 
   // TestRecipe information associated with request.
-  private Recipe recipe;
+  private final Recipe recipe;
 
   // Sampling information associated with request.
-  private Sampling sampling;
+  private final Sampling sampling;
 
   // Additional properties that is of type json.
-  private JsonObject properties;
+  private final JsonObject properties;
 
-  public RequestV1(int version, Workspace workspace, Recipe recipe, Sampling sampling,
-                   JsonObject properties) {
-    this.version = version;
+  public RequestV1(Workspace workspace, Recipe recipe, Sampling sampling, JsonObject properties) {
+    this.version = 1;
     this.workspace = workspace;
     this.recipe = recipe;
     this.sampling = sampling;
@@ -84,5 +85,26 @@ public final class RequestV1 implements Request {
   @Override
   public JsonObject getProperties() {
     return properties;
+  }
+
+  @Override
+  public boolean equals(Object o) {
+    if (this == o) {
+      return true;
+    }
+    if (o == null || getClass() != o.getClass()) {
+      return false;
+    }
+    RequestV1 requestV1 = (RequestV1) o;
+    return version == requestV1.version &&
+      Objects.equals(workspace, requestV1.workspace) &&
+      Objects.equals(recipe, requestV1.recipe) &&
+      Objects.equals(sampling, requestV1.sampling) &&
+      Objects.equals(properties, requestV1.properties);
+  }
+
+  @Override
+  public int hashCode() {
+    return Objects.hash(version, workspace, recipe, sampling, properties);
   }
 }
--- a/wrangler-proto/src/main/java/co/cask/wrangler/proto/Sampling.java
+++ b/wrangler-proto/src/main/java/co/cask/wrangler/proto/Sampling.java
@@ -1,5 +1,5 @@
 /*
- * Copyright © 2017 Cask Data, Inc.
+ * Copyright © 2017-2019 Cask Data, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not
  * use this file except in compliance with the License. You may obtain a copy of
@@ -16,18 +16,26 @@
 
 package co.cask.wrangler.proto;
 
+import java.util.Objects;
+
 /**
  * Defines the sampling specification of the {@link Request}
  */
 public final class Sampling {
   // Sampling method.
-  private String method;
+  private final String method;
 
   // Seeding capability for sampling.
-  private Integer seed;
+  private final Integer seed;
 
   // Number of records to be read.
-  private Integer limit;
+  private final Integer limit;
+
+  public Sampling(String method, Integer seed, Integer limit) {
+    this.method = method;
+    this.seed = seed;
+    this.limit = limit;
+  }
 
   /**
    * @return Method for sampling data.
@@ -56,5 +64,24 @@ public final class Sampling {
     } else {
       return 100;
     }
+  }
+
+  @Override
+  public boolean equals(Object o) {
+    if (this == o) {
+      return true;
+    }
+    if (o == null || getClass() != o.getClass()) {
+      return false;
+    }
+    Sampling sampling = (Sampling) o;
+    return Objects.equals(method, sampling.method) &&
+      Objects.equals(seed, sampling.seed) &&
+      Objects.equals(limit, sampling.limit);
+  }
+
+  @Override
+  public int hashCode() {
+    return Objects.hash(method, seed, limit);
   }
 }
--- a/wrangler-proto/src/main/java/co/cask/wrangler/proto/Workspace.java
+++ b/wrangler-proto/src/main/java/co/cask/wrangler/proto/Workspace.java
@@ -1,5 +1,5 @@
 /*
- * Copyright © 2017 Cask Data, Inc.
+ * Copyright © 2017-2019 Cask Data, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not
  * use this file except in compliance with the License. You may obtain a copy of
@@ -16,15 +16,22 @@
 
 package co.cask.wrangler.proto;
 
+import java.util.Objects;
+
 /**
  * Specification about workspace for the {@link Request}
  */
 public final class Workspace {
   // Name of the workspace.
-  private String name;
+  private final String name;
 
   // Number of results to be returned for the workspace.
-  private Integer results;
+  private final Integer results;
+
+  public Workspace(String name, Integer results) {
+    this.name = name;
+    this.results = results;
+  }
 
   /**
    * @return Name of the workspace.
@@ -38,5 +45,23 @@ public final class Workspace {
    */
   public Integer getResults() {
     return results;
+  }
+
+  @Override
+  public boolean equals(Object o) {
+    if (this == o) {
+      return true;
+    }
+    if (o == null || getClass() != o.getClass()) {
+      return false;
+    }
+    Workspace workspace = (Workspace) o;
+    return Objects.equals(name, workspace.name) &&
+      Objects.equals(results, workspace.results);
+  }
+
+  @Override
+  public int hashCode() {
+    return Objects.hash(name, results);
   }
 }
--- a/wrangler-service/src/main/java/co/cask/wrangler/service/explorer/FilesystemExplorer.java
+++ b/wrangler-service/src/main/java/co/cask/wrangler/service/explorer/FilesystemExplorer.java
@@ -1,5 +1,5 @@
 /*
- * Copyright © 2017 Cask Data, Inc.
+ * Copyright © 2017-2019 Cask Data, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not
  * use this file except in compliance with the License. You may obtain a copy of
@@ -31,6 +31,7 @@ import co.cask.wrangler.api.Row;
 import co.cask.wrangler.dataset.connections.ConnectionType;
 import co.cask.wrangler.dataset.workspace.DataType;
 import co.cask.wrangler.dataset.workspace.WorkspaceDataset;
+import co.cask.wrangler.dataset.workspace.WorkspaceMeta;
 import co.cask.wrangler.proto.PluginSpec;
 import co.cask.wrangler.proto.ServiceResponse;
 import co.cask.wrangler.proto.file.FileConnectionSample;
@@ -38,7 +39,7 @@ import co.cask.wrangler.proto.file.FileSpec;
 import co.cask.wrangler.sampling.Bernoulli;
 import co.cask.wrangler.sampling.Poisson;
 import co.cask.wrangler.sampling.Reservoir;
-import co.cask.wrangler.service.common.AbstractWranglerService;
+import co.cask.wrangler.service.common.AbstractWranglerHandler;
 import co.cask.wrangler.service.common.Format;
 import co.cask.wrangler.utils.ObjectSerDe;
 import com.google.common.base.Charsets;
@@ -48,7 +49,6 @@ import org.apache.twill.filesystem.Location;
 import org.bouncycastle.jce.provider.BouncyCastleProvider;
 
 import java.io.BufferedInputStream;
-import java.io.IOException;
 import java.net.HttpURLConnection;
 import java.security.Security;
 import java.util.ArrayList;
@@ -56,6 +56,7 @@ import java.util.HashMap;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
+import javax.ws.rs.DefaultValue;
 import javax.ws.rs.GET;
 import javax.ws.rs.Path;
 import javax.ws.rs.QueryParam;
@@ -67,7 +68,7 @@ import static co.cask.wrangler.ServiceUtils.sendJson;
  * A {@link FilesystemExplorer} is a HTTP Service handler for exploring the filesystem.
  * It provides capabilities for listing file(s) and directories. It also provides metadata.
  */
-public class FilesystemExplorer extends AbstractWranglerService {
+public class FilesystemExplorer extends AbstractWranglerHandler {
   private static final Gson gson =
     new GsonBuilder().registerTypeAdapter(Schema.class, new SchemaTypeAdapter()).create();
   private Explorer explorer;
@@ -111,7 +112,7 @@ public class FilesystemExplorer extends AbstractWranglerService {
                    @QueryParam("path") String path, @QueryParam("lines") int lines,
                    @QueryParam("sampler") String sampler,
                    @QueryParam("fraction") double fraction,
-                   @QueryParam("scope") String scope) {
+                   @QueryParam("scope") @DefaultValue(WorkspaceDataset.DEFAULT_SCOPE) String scope) {
     String header = request.getHeader(PropertyIds.CONTENT_TYPE);
 
     if (header == null) {
@@ -119,12 +120,8 @@ public class FilesystemExplorer extends AbstractWranglerService {
       return;
     }
 
-    if (scope == null || scope.isEmpty()) {
-      scope = WorkspaceDataset.DEFAULT_SCOPE;
-    }
-
     if (header.equalsIgnoreCase("text/plain") || header.contains("text/")) {
-      loadSamplableFile(responder, scope, path, lines, fraction, sampler);
+      loadSampleableFile(responder, scope, path, lines, fraction, sampler);
     } else if (header.equalsIgnoreCase("application/xml")) {
       loadFile(responder, scope, path, DataType.RECORDS);
     } else if (header.equalsIgnoreCase("application/json")) {
@@ -153,7 +150,7 @@ public class FilesystemExplorer extends AbstractWranglerService {
     try {
       Format format = Format.TEXT;
       if (workspaceId != null) {
-        Map<String, String> config = ws.getProperties(workspaceId);
+        Map<String, String> config = ws.getWorkspace(workspaceId).getProperties();
         String formatStr = config.getOrDefault(PropertyIds.FORMAT, Format.TEXT.name());
         format = Format.valueOf(formatStr);
       }
@@ -177,7 +174,6 @@ public class FilesystemExplorer extends AbstractWranglerService {
   }
 
   private void loadFile(HttpServiceResponder responder, String scope, String path, DataType type) {
-    BufferedInputStream stream = null;
     try {
       Location location = explorer.getLocation(path);
       if (!location.exists()) {
@@ -195,13 +191,6 @@ public class FilesystemExplorer extends AbstractWranglerService {
       String id = String.format("%s:%s:%s:%d", scope, location.getName(),
                                 location.toURI().getPath(), System.nanoTime());
       id = ServiceUtils.generateMD5(id);
-      ws.createWorkspaceMeta(id, scope, name);
-
-      stream = new BufferedInputStream(location.getInputStream());
-      byte[] bytes = new byte[(int) location.length() + 1];
-      stream.read(bytes);
-
-      // Set all properties and write to workspace.
       Map<String, String> properties = new HashMap<>();
       properties.put(PropertyIds.FILE_NAME, location.getName());
       properties.put(PropertyIds.URI, location.toURI().toString());
@@ -210,7 +199,16 @@ public class FilesystemExplorer extends AbstractWranglerService {
       properties.put(PropertyIds.SAMPLER_TYPE, SamplingMethod.NONE.getMethod());
       Format format = type == DataType.BINARY ? Format.BLOB : Format.TEXT;
       properties.put(PropertyIds.FORMAT, format.name());
-      ws.writeProperties(id, properties);
+      WorkspaceMeta workspaceMeta = WorkspaceMeta.builder(id, name)
+        .setScope(scope)
+        .setProperties(properties)
+        .build();
+      ws.writeWorkspaceMeta(workspaceMeta);
+
+      byte[] bytes = new byte[(int) location.length() + 1];
+      try (BufferedInputStream stream = new BufferedInputStream(location.getInputStream())) {
+        stream.read(bytes);
+      }
 
       // Write records to workspace.
       if (type == DataType.RECORDS) {
@@ -218,9 +216,9 @@ public class FilesystemExplorer extends AbstractWranglerService {
         rows.add(new Row(COLUMN_NAME, new String(bytes, Charsets.UTF_8)));
         ObjectSerDe<List<Row>> serDe = new ObjectSerDe<>();
         byte[] data = serDe.toByteArray(rows);
-        ws.writeToWorkspace(id, WorkspaceDataset.DATA_COL, DataType.RECORDS, data);
+        ws.updateWorkspaceData(id, DataType.RECORDS, data);
       } else if (type == DataType.BINARY || type == DataType.TEXT) {
-        ws.writeToWorkspace(id, WorkspaceDataset.DATA_COL, type, bytes);
+        ws.updateWorkspaceData(id, type, bytes);
       }
 
       FileConnectionSample sample = new FileConnectionSample(id, name, ConnectionType.FILE.getType(),
@@ -231,24 +229,15 @@ public class FilesystemExplorer extends AbstractWranglerService {
       responder.sendJson(response);
     } catch (Exception e) {
       error(responder, e.getMessage());
-    } finally {
-      if (stream != null) {
-        try {
-          stream.close();
-        } catch (IOException e) {
-          // Nothing much we can do here.
-        }
-      }
     }
   }
 
-  private void loadSamplableFile(HttpServiceResponder responder,
-                                 String scope, String path, int lines, double fraction, String sampler) {
+  private void loadSampleableFile(HttpServiceResponder responder,
+                                  String scope, String path, int lines, double fraction, String sampler) {
     SamplingMethod samplingMethod = SamplingMethod.fromString(sampler);
     if (sampler == null || sampler.isEmpty() || SamplingMethod.fromString(sampler) == null) {
       samplingMethod = SamplingMethod.FIRST;
     }
-    BoundedLineInputStream stream = null;
     try {
       Location location = explorer.getLocation(path);
       if (!location.exists()) {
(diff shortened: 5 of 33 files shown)


wrangler-storage/src/main/java/co/cask/wrangler/dataset/workspace/WorkspaceDataset.java
```diff
@@ -16,166 +16,123 @@
...
+ *
+ * A workspace contains a data sample, metadata about the workspace, and a set of directives that can be used to
+ * process the data sample. A workspace is tagged with a scope, which can be used to group workspaces together.
+ * It also stores a map of properties, which are connection specific properties that are used to generate the
+ * pipeline source configuration when a pipeline is created from a workspace.
  */
-public class WorkspaceDataset extends AbstractDataset {
+public class WorkspaceDataset {
+  public static final String DATASET_NAME = "workspaces";
   private static final Logger LOG = LoggerFactory.getLogger(WorkspaceDataset.class);
```
CuriousVini: nit: unused?

wrangler-storage/src/main/java/co/cask/wrangler/dataset/workspace/Workspace.java
```diff
@@ -0,0 +1,139 @@
+/*
+ * Copyright © 2018 Cask Data, Inc.
```
CuriousVini: 2019?

wrangler-proto/src/main/java/co/cask/wrangler/proto/RequestV1.java
```diff
@@ -37,9 +39,8 @@
   // Additional properties that is of type json.
   private JsonObject properties;
 
-  public RequestV1(int version, Workspace workspace, Recipe recipe, Sampling sampling,
-                   JsonObject properties) {
-    this.version = version;
+  public RequestV1(Workspace workspace, Recipe recipe, Sampling sampling, JsonObject properties) {
```
CuriousVini, line 42: jc, why properties is of type `JsonObject`?

wrangler-proto/src/main/java/co/cask/wrangler/proto/Workspace.java
```diff
@@ -39,4 +46,23 @@ public String getName() {
...
+      return false;
+    }
+    Workspace workspace = (Workspace) o;
+    return Objects.equals(name, workspace.name) &&
+      Objects.equals(results, workspace.results);
+  }
+
+  @Override
+  public int hashCode() {
+
```
CuriousVini: nit: extra line

wrangler-storage/src/main/java/co/cask/wrangler/dataset/workspace/WorkspaceMeta.java
```diff
@@ -0,0 +1,120 @@
+/*
+ * Copyright © 2018 Cask Data, Inc.
```
CuriousVini: 2019?

wrangler-storage/src/main/java/co/cask/wrangler/dataset/workspace/WorkspaceMeta.java
```diff
@@ -0,0 +1,120 @@
...
+    protected final String name;
+    protected String scope;
+    protected DataType type;
+    protected Map<String, String> properties;
+
+    Builder(String id, String name) {
+      this.id = id;
+      this.name = name;
+      this.properties = new HashMap<>();
+      this.scope = "default";
```
CuriousVini: should be static final

wrangler-service/src/main/java/co/cask/wrangler/service/database/DatabaseService.java
```diff
@@ -493,33 +489,31 @@ public void execute(java.sql.Connection connection) throws Exception {
   @Path("connections/{id}/tables/{table}/read")
   public void read(HttpServiceRequest request, HttpServiceResponder responder,
                    @PathParam("id") String id, @PathParam("table") String table,
-                   @QueryParam("lines") int lines, @QueryParam("scope") String scope) {
+                   @QueryParam("lines") int lines, @QueryParam("scope") @DefaultValue("default") String scope) {
```
CuriousVini: can we reuse `WorkspaceDataset.DEFAULT_SCOPE`?

wrangler-service/src/main/java/co/cask/wrangler/service/directive/DirectivesService.java
```diff
@@ -237,15 +234,11 @@ public void create(HttpServiceRequest request, HttpServiceResponder responder,
   public void list(HttpServiceRequest request, HttpServiceResponder responder,
                    @QueryParam("scope") @DefaultValue("default") String scope) {
```
CuriousVini, line 227: same here: `WorkspaceDataset.DEFAULT_SCOPE`?

wrangler-storage/src/main/java/co/cask/wrangler/dataset/workspace/Workspace.java
```diff
@@ -0,0 +1,139 @@
...
+
+import java.util.Arrays;
+import java.util.Map;
+import java.util.Objects;
+import javax.annotation.Nullable;
+
+/**
+ * Contains all information about a workspace.
+ */
+public class Workspace extends WorkspaceMeta {
```
CuriousVini, line 29: `Workspace` is also `WorkspaceMeta`? Would it be better if the relation is of composition than of inheritance?

wrangler-storage/src/main/java/co/cask/wrangler/dataset/workspace/WorkspaceMeta.java
```diff
@@ -0,0 +1,120 @@
...
+
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Objects;
+
+/**
+ * Metadata about a workspace.
+ */
+public class WorkspaceMeta extends WorkspaceIdentifier {
```
CuriousVini, line 29: same  question about composition of WorkspaceIdentifier

wrangler-storage/src/main/java/co/cask/wrangler/dataset/workspace/RequestDeserializer.java
```diff
@@ -53,7 +53,7 @@ public Request deserialize(JsonElement json, Type type, JsonDeserializationConte
       Recipe recipe = context.deserialize(object.get("recipe"), Recipe.class);
       Sampling sampling = context.deserialize(object.get("sampling"), Sampling.class);
       JsonObject properties = context.deserialize(object.get("properties"), JsonObject.class);
-      return new RequestV1(version, workspace, recipe, sampling, properties);
+      return new RequestV1(workspace, recipe, sampling, properties);
```
CuriousVini, line 56: Not part of this PR, but do you think its a good practice to do validation in deserializer? Seems like this should happen before deserialize is called.

wrangler-service/src/main/java/co/cask/wrangler/service/common/AbstractWranglerService.java
```diff
@@ -44,34 +42,40 @@
   @UseDataSet(DataPrep.CONNECTIONS_DATASET)
   private Table connectionTable;
 
-  @UseDataSet(WORKSPACE_DATASET)
+  @UseDataSet(WorkspaceDataset.DATASET_NAME)
```
CuriousVini, line 43: nit: not able to comment on that line but, is `GSON` unused?

wrangler-storage/src/main/java/co/cask/wrangler/dataset/workspace/WorkspaceDataset.java
```diff
@@ -186,50 +143,83 @@ public boolean hasWorkspace(String id) throws WorkspaceException {
...
+   *
+   * @param id the workspace id
+   * @param request the directive execution request
+   * @throws WorkspaceNotFoundException if the workspace does not exist
+   */
+  public void updateWorkspaceRequest(String id, Request request) throws WorkspaceNotFoundException {
+    Workspace existing = getWorkspace(id);
+    Workspace updated = Workspace.builder(existing)
+      .setRequest(request)
+      .setUpdated(System.currentTimeMillis() / 1000)
```
CuriousVini, line 181: should we use `TimeUnit` instead?

wrangler-service/src/main/java/co/cask/wrangler/service/directive/DirectivesService.java
```diff
@@ -237,15 +234,11 @@ public void create(HttpServiceRequest request, HttpServiceResponder responder,
...
-      List<Pair<String, String>> workspaces = table.getWorkspaces(scope);
-      List<WorkspaceIdentifier> array = new ArrayList<>(workspaces.size());
-      for (Pair<String, String> workspace : workspaces) {
-        array.add(new WorkspaceIdentifier(workspace.getFirst(), workspace.getSecond()));
-      }
-      ServiceResponse<WorkspaceIdentifier> response = new ServiceResponse<>(array);
+      List<WorkspaceIdentifier> workspaces = table.listWorkspaces(scope);
+      ServiceResponse<WorkspaceIdentifier> response = new ServiceResponse<>(workspaces);
       responder.sendJson(response);
-    } catch (WorkspaceException e) {
```
CuriousVini, line 247: If we do not throw this exception any more, is it safe to remove the class?

wrangler-service/src/main/java/co/cask/wrangler/service/directive/DirectivesService.java
```diff
@@ -237,15 +234,11 @@ public void create(HttpServiceRequest request, HttpServiceResponder responder,
...
-      error(responder, e.getMessage());
+    } catch (RuntimeException e) {
```
CuriousVini, line 232: Why are we catching RuntimeException instead of checked exception?

CuriousVini (commented):
few comments. Also please update license year in modified classes.

wrangler-proto/src/main/java/co/cask/wrangler/proto/Workspace.java
```diff
@@ -26,6 +28,11 @@
   // Number of results to be returned for the workspace.
   private Integer results;
```
anew: should name and results be final?

wrangler-proto/src/main/java/co/cask/wrangler/proto/Sampling.java
```diff
@@ -29,6 +31,12 @@
   // Number of records to be read.
   private Integer limit;
```
anew: should the fields be final?

wrangler-proto/src/main/java/co/cask/wrangler/proto/RequestV1.java
```diff
@@ -37,9 +39,8 @@
   // Additional properties that is of type json.
   private JsonObject properties;
```
anew: should all the fields be final?

wrangler-proto/src/main/java/co/cask/wrangler/proto/Recipe.java
```diff
@@ -71,4 +72,23 @@ public Boolean getSave() {
   public String getName() {
     return name;
   }
+
```
anew, line 75: Since this class has no setters, should all the fields be final?

wrangler-service/src/main/java/co/cask/wrangler/service/database/DatabaseService.java
```diff
@@ -102,7 +98,9 @@
   private static final Logger LOG = LoggerFactory.getLogger(DatabaseService.class);
```
anew: BigQueryService extends AbstractWranglerService. Is this class the equivalent for dabatabases? Can it then also extend AbstractWranglerService?

wrangler-service/src/main/java/co/cask/wrangler/service/database/DatabaseService.java
```diff
@@ -493,33 +489,31 @@ public void execute(java.sql.Connection connection) throws Exception {
...
       cleanup = loadAndExecute(id, connection -> {
-        String grp = scope;
-        if (Strings.isNullOrEmpty(scope)) {
-          grp = WorkspaceDataset.DEFAULT_SCOPE;
-        }
         try (Statement statement = connection.createStatement();
              ResultSet result = statement.executeQuery(String.format("select * from %s", table))) {
           List<Row> rows = getRows(lines, result);
 
           String identifier = ServiceUtils.generateMD5(table);
```
anew, line 484: For BQ, it is

`String identifier = ServiceUtils.generateMD5(String.format("%s:%s", scope, tableId));
`

that is, including the scope. Should it be consistently the same here?

wrangler-service/src/main/java/co/cask/wrangler/service/database/DatabaseService.java
```diff
@@ -493,33 +489,31 @@ public void execute(java.sql.Connection connection) throws Exception {
...
-          byte[] data = serDe.toByteArray(rows);
-          ws.writeToWorkspace(identifier, WorkspaceDataset.DATA_COL, DataType.RECORDS, data);
-
           Map<String, String> properties = new HashMap<>();
           properties.put(PropertyIds.ID, identifier);
           properties.put(PropertyIds.NAME, table);
           properties.put(PropertyIds.CONNECTION_TYPE, ConnectionType.DATABASE.getType());
           properties.put(PropertyIds.SAMPLER_TYPE, SamplingMethod.NONE.getMethod());
           properties.put(PropertyIds.CONNECTION_ID, id);
-          ws.writeProperties(identifier, properties);
```
anew, line 521: Generally this looks like a lot of duplication with BigQueryService.

wrangler-service/src/main/java/co/cask/wrangler/service/database/DatabaseService.java
anew: Should this class be called a DatabaseHandler instead? Since it implements an Http Handler... same for the BigQuery one.

wrangler-service/src/main/java/co/cask/wrangler/service/directive/DirectivesService.java
```diff
@@ -120,45 +126,33 @@
...
 import javax.ws.rs.QueryParam;
 
-import static co.cask.wrangler.ServiceUtils.error;
-import static co.cask.wrangler.ServiceUtils.sendJson;
-import static co.cask.wrangler.ServiceUtils.success;
-
 /**
  * Service for managing workspaces and also application of directives on to the workspace.
  */
 public class DirectivesService extends AbstractHttpServiceHandler {
```
anew: DierctivesHandler?

wrangler-service/src/main/java/co/cask/wrangler/service/directive/DirectivesService.java
```diff
@@ -120,45 +126,33 @@
...
   private static final String COLUMN_NAME = "body";
   private static final String RECORD_DELIMITER_HEADER = "recorddelimiter";
   private static final String DELIMITER_HEADER = "delimiter";
 
-  private final Gson gson = new Gson();
-
-  @UseDataSet(WORKSPACE_DATASET)
+  @UseDataSet(WorkspaceDataset.DATASET_NAME)
+  private Table workspaceTable;
   private WorkspaceDataset table;
```
anew: to be consistent with other handlers, better call this "ws"

wrangler-service/src/main/java/co/cask/wrangler/service/directive/DirectivesService.java
```diff
@@ -201,14 +195,17 @@ public void create(HttpServiceRequest request, HttpServiceResponder responder,
...
 
-      table.createWorkspaceMeta(id, name, scope);
       Map<String, String> properties = new HashMap<>();
       properties.put(PropertyIds.ID, id);
       properties.put(PropertyIds.NAME, name);
-      table.writeProperties(id, properties);
-      success(responder, String.format("Successfully created workspace '%s'", id));
-    } catch (WorkspaceException e) {
-      error(responder, e.getMessage());
+      WorkspaceMeta workspaceMeta = WorkspaceMeta.builder(id, name)
```
anew, line 193: in other cases, the identifier is an MD5 of the scope and the table name. Here just the name table id itself?

wrangler-service/src/main/java/co/cask/wrangler/service/directive/DirectivesService.java
```diff
@@ -269,9 +262,9 @@ public void delete(HttpServiceRequest request, HttpServiceResponder responder,
                      @PathParam("id") String id) {
     try {
       table.deleteWorkspace(id);
-      success(responder, String.format("Successfully deleted workspace '%s'", id));
-    } catch (WorkspaceException e) {
-      error(responder, e.getMessage());
+      ServiceUtils.success(responder, String.format("Successfully deleted workspace '%s'", id));
+    } catch (RuntimeException e) {
```
anew: maybe better catch any throwable?

wrangler-service/src/main/java/co/cask/wrangler/service/directive/DirectivesService.java
```diff
@@ -504,43 +498,39 @@ public void upload(HttpServiceRequest request, HttpServiceResponder responder,
       // Extract content.
       byte[] content = handler.getContent();
       if (content == null) {
-        error(responder, "Body not present, please post the file containing the records to be wrangle.");
+        ServiceUtils.error(responder, "Body not present, please post the file containing the records to be wrangle.");
```
anew: to be wrangled.

wrangler-proto/src/main/java/co/cask/wrangler/proto/RequestV1.java
albertshau: I'm not sure. I haven't yet spent the time to figure out what is inside that blob, though I'm hoping it's actually just a string to string map.

wrangler-service/src/main/java/co/cask/wrangler/service/directive/DirectivesService.java
```diff
@@ -582,16 +574,19 @@ public void execute(HttpServiceRequest request, HttpServiceResponder responder,
     try {
       RequestExtractor handler = new RequestExtractor(request);
       Request user = handler.getContent("UTF-8", Request.class);
```
anew: name this request instead of user?

wrangler-service/src/main/java/co/cask/wrangler/service/directive/DirectivesService.java
```diff
@@ -628,20 +623,20 @@ public void execute(HttpServiceRequest request, HttpServiceResponder responder,
...
 
       DirectiveExecutionResponse response = new DirectiveExecutionResponse(values, headers, types,
                                                                            user.getRecipe().getDirectives());
       responder.sendJson(response);
     } catch (JsonParseException e) {
       LOG.warn(e.getMessage(), e);
-      error(responder, "Issue parsing request. " + e.getMessage());
+      ServiceUtils.error(responder, "Issue parsing request. " + e.getMessage());
     } catch (DirectiveParseException e) {
       LOG.warn(e.getMessage(), e);
```
anew: why log in this case? This is simply a bad request, right? And then the response should be a 400

wrangler-storage/src/main/java/co/cask/wrangler/dataset/workspace/RequestDeserializer.java
albertshau: I don't think we should have that deserializer at all or the Request interface. I think it should just be the concrete class, but wanted to limit the changes.

wrangler-service/src/main/java/co/cask/wrangler/service/directive/DirectivesService.java
```diff
@@ -706,14 +701,17 @@ public void summary(HttpServiceRequest request, HttpServiceResponder responder,
     try {
       RequestExtractor handler = new RequestExtractor(request);
       Request user = handler.getContent("UTF-8", Request.class);
```
anew: same here - call it request?

wrangler-service/src/main/java/co/cask/wrangler/service/directive/DirectivesService.java
albertshau: so it's clear there's nothing in the try block that throws checked exceptions and to preserve the existing behavior.

wrangler-storage/src/main/java/co/cask/wrangler/dataset/workspace/WorkspaceDataset.java
albertshau: removed

wrangler-service/src/main/java/co/cask/wrangler/service/directive/DirectivesService.java
```diff
@@ -773,7 +771,7 @@ public void summary(HttpServiceRequest request, HttpServiceResponder responder,
       WorkspaceSummaryResponse response = new WorkspaceSummaryResponse(validationResult);
       responder.sendJson(response);
     } catch (Exception e) {
```
anew: Can this not cause a JsonParseException or DirectiveParseException ?

wrangler-proto/src/main/java/co/cask/wrangler/proto/Workspace.java
albertshau: done

wrangler-storage/src/main/java/co/cask/wrangler/dataset/workspace/WorkspaceMeta.java
albertshau: done

wrangler-service/src/main/java/co/cask/wrangler/service/database/DatabaseService.java
albertshau: done

wrangler-service/src/main/java/co/cask/wrangler/service/directive/DirectivesService.java
albertshau: done

wrangler-service/src/main/java/co/cask/wrangler/service/directive/DirectivesService.java
albertshau: yep, removed.

wrangler-storage/src/main/java/co/cask/wrangler/dataset/workspace/WorkspaceDataset.java
albertshau: done

wrangler-storage/src/main/java/co/cask/wrangler/dataset/workspace/Workspace.java
albertshau: In the future I want these to be proto objects so it has to be inheritance for the fields to be correct. This should be what is returned by the GET /workspaces/id endpoint, but I need to figure out what is going on with that JsonObject blob first.

WorkspaceMeta is what *should* be contained in the workspace creation request body, but the current API has the fields separate and specified through query params.

wrangler-service/src/main/java/co/cask/wrangler/service/explorer/FilesystemExplorer.java
```diff
@@ -258,7 +257,18 @@ private void loadSamplableFile(HttpServiceResponder responder,
       String name = location.getName();
       String id = String.format("%s:%s:%d", location.getName(), location.toURI().getPath(), System.nanoTime());
```
anew: in line 193, the id includes the scope. Why not here?

wrangler-service/src/main/java/co/cask/wrangler/service/kafka/KafkaService.java
```diff
@@ -49,21 +50,23 @@
...
 
 import static co.cask.wrangler.ServiceUtils.error;
-import static co.cask.wrangler.service.directive.DirectivesService.WORKSPACE_DATASET;
 
 /**
  * Service for handling Kafka connections.
  */
 public final class KafkaService extends AbstractHttpServiceHandler {
 
-  @UseDataSet(WORKSPACE_DATASET)
```
anew, line 66: Again I am not sure why this does not extend AbstractWranglerService?

wrangler-service/src/main/java/co/cask/wrangler/service/s3/S3Service.java
```diff
@@ -387,13 +384,11 @@ private void loadFile(String connectionId, HttpServiceResponder responder, Input
       String file = String.format("%s:%s", s3Object.getBucketName(), s3Object.getKey());
```
anew, line 384: scope as part of id?

wrangler-storage/src/main/java/co/cask/wrangler/dataset/workspace/WorkspaceMeta.java
albertshau: done

wrangler-service/src/main/java/co/cask/wrangler/service/spanner/SpannerService.java
```diff
@@ -209,15 +204,22 @@ public void readTable(HttpServiceRequest request, HttpServiceResponder responder
...
         new SpannerSpecification(externalDsName, path, projectId, instanceId, databaseId, tableId, schema);
 
-      // initialize and store workspace properties
       Map<String, String> workspaceProperties = new HashMap<>();
       workspaceProperties.put(PropertyIds.ID, identifier);
       workspaceProperties.put(PropertyIds.NAME, tableId);
       workspaceProperties.put(PropertyIds.CONNECTION_TYPE, ConnectionType.SPANNER.getType());
       workspaceProperties.put(PropertyIds.CONNECTION_ID, connectionId);
       workspaceProperties.put(PropertyIds.PLUGIN_SPECIFICATION, GSON.toJson(specification));
+      WorkspaceMeta workspaceMeta = WorkspaceMeta.builder(identifier, tableId)
```
anew, line 213: this code seems to be duplicated with slight variations across all the handlers.

wrangler-storage/src/main/java/co/cask/wrangler/dataset/workspace/Workspace.java
```diff
@@ -0,0 +1,139 @@
...
+    return created == workspace.created &&
+      updated == workspace.updated &&
+      Arrays.equals(data, workspace.data) &&
+      Objects.equals(request, workspace.request);
+  }
+
+  @Override
+  public int hashCode() {
+    int result = Objects.hash(super.hashCode(), created, updated, request);
+    result = 31 * result + Arrays.hashCode(data);
```
anew, line 83: can this be added as an extra parameter to Objects.hash()?

wrangler-storage/src/main/java/co/cask/wrangler/dataset/workspace/WorkspaceDataset.java
```diff
@@ -186,50 +143,83 @@ public boolean hasWorkspace(String id) throws WorkspaceException {
...
    *
+   * TODO: (CDAP-14692) make sure scope is indexed so this doesn't require a full table scan
+   *
    * @param group to be deleted.
    * @return number of workspaces deleted within a group.
-   * @throws WorkspaceException thrown if there are issues with deleting workspace within group.
    */
-  @WriteOnly
-  public int deleteGroup(String group) throws WorkspaceException {
+  public int deleteGroup(String group) {
```
anew: is a group the same as a scope? If so, maybe name the method and parameters that way

wrangler-service/src/main/java/co/cask/wrangler/service/database/DatabaseService.java
albertshau: Most of these ids are just plain wrong and I have opened a separate blocker jira (CDAP-14691) to fix them.

wrangler-storage/src/main/java/co/cask/wrangler/dataset/workspace/Workspace.java
albertshau: no because the array instance hashcode will be used instead array contents.

This is the IntelliJ auto-generated method.

wrangler-service/src/main/java/co/cask/wrangler/service/database/DatabaseService.java
albertshau: I can rename them to Handler and make them all extend AbstractWranglerService

wrangler-service/src/main/java/co/cask/wrangler/service/database/DatabaseService.java
albertshau: There is a lot of duplication across all the type specific handlers. I have generally avoided dealing with the duplication because when we eventually re-design connections, all of these individual handlers should go away.

wrangler-service/src/main/java/co/cask/wrangler/service/directive/DirectivesService.java
albertshau: fixed

wrangler-service/src/main/java/co/cask/wrangler/service/directive/DirectivesService.java
albertshau: renamed to directiveRequest since 'request' is the http request.

wrangler-service/src/main/java/co/cask/wrangler/service/directive/DirectivesService.java
albertshau: removed the warning logs

wrangler-service/src/main/java/co/cask/wrangler/service/directive/DirectivesService.java
albertshau: added catch clauses

wrangler-service/src/main/java/co/cask/wrangler/service/directive/DirectivesService.java
albertshau: done

wrangler-service/src/main/java/co/cask/wrangler/service/explorer/FilesystemExplorer.java
albertshau: seems like a bug, adding it

wrangler-storage/src/main/java/co/cask/wrangler/dataset/workspace/WorkspaceDataset.java
albertshau: done

wrangler-service/src/main/java/co/cask/wrangler/service/s3/S3Service.java
albertshau: I'm going to do the IDs in a separate pass

wrangler-proto/src/main/java/co/cask/wrangler/proto/Recipe.java
albertshau: done

wrangler-service/src/main/java/co/cask/wrangler/service/directive/DirectivesService.java
albertshau: in this case it's provided by the user. I'm not sure when/if this endpoint called.

wrangler-service/src/main/java/co/cask/wrangler/service/database/DatabaseService.java
anew: makes sense

wrangler-service/src/main/java/co/cask/wrangler/service/directive/DirectivesService.java
anew: got it

wrangler-service/src/main/java/co/cask/wrangler/service/directive/DirectivesService.java
anew: oh right

wrangler-storage/src/main/java/co/cask/wrangler/dataset/workspace/Workspace.java
anew: ok cool.

wrangler-service/src/main/java/co/cask/wrangler/service/explorer/FilesystemExplorer.java
```diff
@@ -210,17 +201,25 @@ private void loadFile(HttpServiceResponder responder, String scope, String path,
...
       Format format = type == DataType.BINARY ? Format.BLOB : Format.TEXT;
       properties.put(PropertyIds.FORMAT, format.name());
-      ws.writeProperties(id, properties);
+      WorkspaceMeta workspaceMeta = WorkspaceMeta.builder(id, name)
+        .setScope(scope)
+        .setProperties(properties)
+        .build();
+      ws.writeWorkspaceMeta(workspaceMeta);
+
+      stream = new BufferedInputStream(location.getInputStream());
```
CuriousVini: should we use try with resources instead? Also should we move variable declaration closer to its usage?

wrangler-service/src/main/java/co/cask/wrangler/service/explorer/FilesystemExplorer.java
```diff
@@ -210,17 +201,25 @@ private void loadFile(HttpServiceResponder responder, String scope, String path,
...
+      byte[] bytes = new byte[(int) location.length() + 1];
+      stream.read(bytes);
```
CuriousVini: do we always update workspace or check if data was read into buffer?

wrangler-storage/src/main/java/co/cask/wrangler/dataset/workspace/WorkspaceDataset.java
CuriousVini: missed?

wrangler-storage/src/main/java/co/cask/wrangler/dataset/workspace/Workspace.java
```diff
@@ -0,0 +1,139 @@
...
+      return this;
+    }
+
+    public Builder setUpdated(long updated) {
+      this.updated = updated;
+      return this;
+    }
+
+    public Builder setData(byte[] data) {
+      this.data = data;
```
CuriousVini, line 126: should we make copy of byte array?

CuriousVini (commented):
couple of minor comments. No additional comments from me.

wrangler-storage/src/main/java/co/cask/wrangler/dataset/workspace/Workspace.java
albertshau: fixed (in constructor)

wrangler-service/src/main/java/co/cask/wrangler/service/explorer/FilesystemExplorer.java
albertshau: I haven't changed the logic here, it never checked.

wrangler-service/src/main/java/co/cask/wrangler/service/explorer/FilesystemExplorer.java
albertshau: fixed

CuriousVini (approved):
lgtm 👍

PR #295, opened 2019-01-10
CDAP-14619 cleanup for connections

albertshau:
Refactored connections to improve error handling and encapsulation.

Consolidated the ConnectionStore and AbstractTableStore into a
single class since there was no other implementation of the
AbstractTableStore. Also changed the way connections are stored
to store each individual connection field in its own column.
This is so that all the stores follow the same paradigm, which
will also be required once they are moved to the new storage SPI.

Introduced proper error handling for connections such that
the endpoints return 404 when a connection is not found instead
of a 500, and return a 409 when a connection already exists instead
of a 500.

Removed the useless 'undefined' ConnectionType that was never
possible to be a value unless the user actually passed in
'undefined' as type, in which case the request would be rejected.

Broke up the Connection class into a Connection class and a
ConnectionMeta class. The Meta class contains information that
the user provides, but not the system added information like
id, created, and updated times. Both of these classes have
final fields instead of setters so that correctness can be
more easily enforced.

Also fixed the Spanner service to use the same error responses
as the other service endpoints.


wrangler-storage/src/main/java/co/cask/wrangler/dataset/connections/ConnectionStore.java
```diff
@@ -35,96 +53,162 @@
...
-    String mangled = getConnectionId(connectionName);
-    return hasKey(mangled);
+    return read(getConnectionId(connectionName)) != null;
+  }
+
+  /**
+   * Scans the namespace to list all the keys applying the filter.
+   *
+   * @param filter to be applied on the data being returned.
+   * @return List of objects of type T.
```
CuriousVini: nit: Connection

wrangler-service/src/main/java/co/cask/wrangler/service/s3/S3Handler.java
```diff
@@ -98,21 +100,12 @@ public void testS3Connection(HttpServiceRequest request, HttpServiceResponder re
     try {
       // Extract the body of the request and transform it to the Connection object.
       RequestExtractor extractor = new RequestExtractor(request);
-      Connection connection = extractor.getContent(Charsets.UTF_8.name(), Connection.class);
-      if (connection == null) {
-        responder.sendError(400, "Connection information is missing from the request body.");
-        return;
-      }
-      ConnectionType connectionType = ConnectionType.fromString(connection.getType().getType());
-      if (connectionType == ConnectionType.UNDEFINED || connectionType != ConnectionType.S3) {
```
CuriousVini, line 107: does ui uses `ConnectionType.UNDEFINED` for anything?

wrangler-service/src/main/java/co/cask/wrangler/service/connections/ConnectionHandler.java
```diff
@@ -366,16 +358,15 @@ public void updateProp(HttpServiceRequest request, HttpServiceResponder responde
...
-      if (connection == null) {
-        notFound(responder, String.format(
-          "Connection with id '%s' not found in the store", id
-        ));
-        return;
-      }
-      responder.sendJson(new ServiceResponse<>(ImmutableList.of(connection)));
+      Connection connection = store.get(id);
+      ConnectionMeta clone = ConnectionMeta.builder(connection)
+        .setName(connection.getName() + "_Clone")
```
CuriousVini, line 363: nit: static final?

CuriousVini (approved):
Few minor comments. Lgtm 👍

wrangler-service/src/main/java/co/cask/wrangler/service/connections/ConnectionHandler.java
albertshau: this is only used once, so I think it's more readable to just have it hardcoded here versus one more layer of indirection.

wrangler-service/src/main/java/co/cask/wrangler/service/s3/S3Handler.java
albertshau: no, because it would just hit this clause and error out. It seems like this was intended to check for the case where the user provides an invalid type like "S4", but the logic was implemented incorrectly. I have another PR that consolidates error handling and will make sure this scenario is properly handled.

wrangler-storage/src/main/java/co/cask/wrangler/dataset/connections/ConnectionStore.java
albertshau: done

PR #299, opened 2019-01-23
CDAP-14794 CDAP-14795 fix loading of UDDs

albertshau:
Fixed a bug where loading UDDs would generate an artifact not
found error because the namespace was not being passed to the
createClassLoader call.

Also changed the usage endpoint to always load UDDs in that
namespace to mirror the behavior that was there before the
namespace refactoring.


albertshau:
cannot be merged until https://github.com/cdapio/cdap/pull/10966 is available.

CuriousVini (approved):
lgtm

PR #303, opened 2019-02-05
[CDAP-14698] Changes 'Directives' section to be named 'Transformation steps'

ajainarayanan:
JIRA: https://issues.cask.co/browse/CDAP-14698

### DIFF
--- a/wrangler-transform/widgets/Wrangler-transform.json
+++ b/wrangler-transform/widgets/Wrangler-transform.json
@@ -29,7 +29,7 @@
       ]
     },
     {
-      "label": "Directives",
+      "label": "Transformation steps",
       "properties": [
         {
           "widget-type": "wrangler-directives",


wrangler-transform/widgets/Wrangler-transform.json
```diff
@@ -29,7 +29,7 @@
       ]
     },
     {
-      "label": "Directives",
+      "label": "Transformation steps",
```
CuriousVini, line 32: Label for UDD is still User defined Directives? Or that should also be changed?

CuriousVini: Also should there be corresponding change in docs as well? https://github.com/data-integrations/wrangler/blob/develop/wrangler-transform/docs/Wrangler-transform.md

ajainarayanan: @CuriousVini Good point. This change is only on the UI. I don't know changing UDD falls under the scope. @bdmogal Any suggestions? 

Will fix the doc.

CuriousVini: yea I was referring to UI label UDD https://github.com/data-integrations/wrangler/blob/develop/wrangler-transform/widgets/Wrangler-transform.json#L44
image

ajainarayanan:
Right now on-hold as @bdmogal still deciding on scope of changes we need to do

bdmogal:
Closing this PR, since we don't want to make this change.

PR #304, opened 2019-02-06
CDAP-14879 namespace test endpoints and namespace jdbc drivers

albertshau:
Fixed a bug where the namespace was not properly being passed
to artifact calls when dealing with jdbc drivers. This would cause
the service to look for system jdbc drivers instead of looking in
the relevant namespace.


elfenheart (approved):
👍 LGTM

PR #307, opened 2019-02-08
CDAP-14692 switch stores to use StructuredTable

albertshau:
Switch to use system tables instead of datasets.


wrangler-service/src/main/java/co/cask/wrangler/service/common/AbstractWranglerHandler.java
```diff
@@ -96,28 +84,37 @@ protected boolean shouldCopyHeader(@Nullable NamespacedId workspaceId) {
...
+      ConnectionStore store = ConnectionStore.get(context);
+      return store.get(connectionId);
+    }, ConnectionNotFoundException.class);
+  }
+
+  protected Connection getValidatedConnection(NamespacedId connectionId, ConnectionType expectedType) {
+    return TransactionRunners.run(getContext(), context -> {
+      ConnectionStore store = ConnectionStore.get(context);
+      return getValidatedConnection(store, connectionId, expectedType);
+    }, ConnectionNotFoundException.class);
```
CuriousVini: should this also include `IOException` ?

CuriousVini: Also what would happen if run method throws a RuntimeException `BadRequestException`?

wrangler-service/src/main/java/co/cask/wrangler/service/gcs/GCSHandler.java
```diff
@@ -325,9 +319,14 @@ public void loadObject(HttpServiceRequest request, HttpServiceResponder responde
         .setScope(scope)
         .setProperties(properties)
         .build();
-      ws.writeWorkspaceMeta(workspaceMeta);
+      if (blob.isDirectory()) {
+        throw new BadRequestException("Path specified is not a file.");
```
CuriousVini: should we add blobpath to error message? Having contextual info in error message would be better?

wrangler-service/src/main/java/co/cask/wrangler/service/schema/SchemaRegistryHandler.java
```diff
@@ -148,19 +132,24 @@ public void upload(HttpServiceRequest request, HttpServiceResponder responder,
   public void upload(HttpServiceRequest request, HttpServiceResponder responder, @PathParam("context") String namespace,
                      @PathParam("id") String id) {
     respond(request, responder, namespace, () -> {
-      byte[] bytes = null;
+      byte[] bytes;
       ByteBuffer content = request.getContent();
       if (content != null && content.hasRemaining()) {
         bytes = new byte[content.remaining()];
         content.get(bytes);
+      } else {
```
CuriousVini, line 139: why this change is needed?

wrangler-storage/src/main/java/co/cask/wrangler/dataset/schema/SchemaRegistry.java
```diff
@@ -41,30 +49,98 @@
  * A schema entry contains the actual schema bytes and is uniquely identified by the schema it belongs to and a
  * version number. A schema entry is added to a schema and can also be removed. In addition, each schema keeps track
  * of the latest schema entry.
+ *
+ * Schema information is stored in two tables, one for the schema metadata and one for the schema entries.
```
CuriousVini, line 53: Nice!  +1 on javadoc

CuriousVini (commented):
Should we add IOException as expected Exception type?

wrangler-storage/src/main/java/co/cask/wrangler/dataset/connections/ConnectionStore.java
```diff
@@ -61,18 +70,44 @@
...
+  private static final String NAMESPACE_COL = "namespace";
+  private static final String ID_COL = "id";
   private static final String TYPE_COL = "type";
   private static final String NAME_COL = "name";
   private static final String DESC_COL = "description";
   private static final String PROPERTIES_COL = "properties";
   private static final String CREATED_COL = "created";
   private static final String UPDATED_COL = "updated";
-  private final Table table;
+  private static final StructuredTableId TABLE_ID = new StructuredTableId("connections");
```
CuriousVini, line 70: should we also change javadoc for this store?

wrangler-storage/src/main/java/co/cask/wrangler/dataset/schema/SchemaRegistry.java
```diff
@@ -41,30 +49,98 @@
...
+ * schema entry version, and the auto column is used for generating version numbers for any new schema entries.
+ *
+ * The schema entry table contains four columns:
+ *
+ * namespace, id, version, schema
+ *
+ * The namespace, id, and version columns form the primary key, while the schema column contains the actual schema.
  */
 public final class SchemaRegistry  {
   public static final String DATASET_NAME = "schemaRegistry";
```
CuriousVini: unused?

CuriousVini (commented):
code changes lgtm 👍  just a few minor comments

wrangler-service/src/main/java/co/cask/wrangler/service/gcs/GCSHandler.java
albertshau: fixed

wrangler-service/src/main/java/co/cask/wrangler/service/schema/SchemaRegistryHandler.java
albertshau: since bytes is now used in a closure, it needs to be effectively final

wrangler-storage/src/main/java/co/cask/wrangler/dataset/schema/SchemaRegistry.java
albertshau: fixed

wrangler-storage/src/main/java/co/cask/wrangler/dataset/connections/ConnectionStore.java
albertshau: added a couple lines about the schema and primary key.

wrangler-service/src/main/java/co/cask/wrangler/service/common/AbstractWranglerHandler.java
albertshau: hm, actually I don't need to explicitly mention any of these exceptions since they're all RuntimeException, which gets propagated by TransactionsRunners. Will change these.

albertshau:
In terms of adding IOException, I don't think it will make a difference, as everything will get caught by the respond() method and return a 500.

CuriousVini (approved):
lgtm 👍

PR #308, opened 2019-02-12
CDAP-14827 add namespace generation to entity keys

albertshau:
Added namespace generation to entity keys to make it behave like
all workspaces and connections are deleted when a CDAP namespace
is deleted then re-created.

### DIFF
--- a/wrangler-proto/src/main/java/co/cask/wrangler/proto/connection/Connection.java
+++ b/wrangler-proto/src/main/java/co/cask/wrangler/proto/connection/Connection.java
@@ -41,7 +41,7 @@ public final class Connection extends ConnectionMeta {
                     Map<String, String> properties) {
     super(type, name, description, properties);
     this.namespacedId = id;
-    this.context = id.getNamespace();
+    this.context = id.getNamespace().getName();
     this.id = id.getId();
     this.created = created;
     this.updated = updated;
@@ -54,9 +54,8 @@ public final class Connection extends ConnectionMeta {
   /**
    * @return id of the connection.
    */
-  public NamespacedId getId() {
-    // only null if this object was created through deserialization
-    return namespacedId == null ? new NamespacedId(context, id) : namespacedId;
+  public String getId() {
+    return id;
   }
 
   /**
--- a/wrangler-service/src/main/java/co/cask/wrangler/service/common/AbstractWranglerHandler.java
+++ b/wrangler-service/src/main/java/co/cask/wrangler/service/common/AbstractWranglerHandler.java
@@ -16,6 +16,7 @@
 
 package co.cask.wrangler.service.common;
 
+import co.cask.cdap.api.NamespaceSummary;
 import co.cask.cdap.api.service.http.AbstractSystemHttpServiceHandler;
 import co.cask.cdap.api.service.http.HttpServiceRequest;
 import co.cask.cdap.api.service.http.HttpServiceResponder;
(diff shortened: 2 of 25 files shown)


wrangler-proto/src/main/java/co/cask/wrangler/proto/connection/Connection.java
```diff
@@ -54,9 +54,8 @@ public String getNamespace() {
   /**
    * @return id of the connection.
    */
-  public NamespacedId getId() {
```
CuriousVini, line 57: why this change is needed?

wrangler-service/src/main/java/co/cask/wrangler/service/common/AbstractWranglerHandler.java
```diff
@@ -123,6 +125,36 @@ protected Connection getValidatedConnection(ConnectionStore store, NamespacedId
...
+   * to create the response.
+   * If a {@link JsonSyntaxException} is thrown, a 400 response will be sent.
+   * If anything else if thrown, a 500 response will be sent.
+   * If nothing is thrown, the result of the callable will be sent as json.
+   *
+   * @param request the http request
+   * @param responder the http responder
+   * @param callable the endpoint logic to run
+   */
+  protected <T> void respond(HttpServiceRequest request, HttpServiceResponder responder, Callable<T> callable) {
```
CuriousVini, line 143: why do we need 2 methods one with namespace and one without?

wrangler-service/src/main/java/co/cask/wrangler/service/common/AbstractWranglerHandler.java
```diff
@@ -136,15 +168,19 @@ protected Connection getValidatedConnection(ConnectionStore store, NamespacedId
...
+                             NamespacedResponder<T> callable) {
     // system namespace does not officially exist, so don't check existence for system namespace.
-    if (namespace != null && !Contexts.SYSTEM.equals(namespace)) {
+    NamespaceSummary namespaceSummary;
+    if (Contexts.SYSTEM.equals(namespace)) {
+      namespaceSummary = new NamespaceSummary(Contexts.SYSTEM, "", 0L);
+    } else {
       try {
-        if (!getContext().getAdmin().namespaceExists(namespace)) {
+        namespaceSummary = getContext().getAdmin().getNamespaceSummary(namespace);
```
CuriousVini, line 182: is this call guaranteed to return latest generation of a namespace? Or only one generation is stored for a namespace at a given time?

CuriousVini (approved):
lgtm 👍  for code changes, just a few questions.

wrangler-service/src/main/java/co/cask/wrangler/service/common/AbstractWranglerHandler.java
albertshau: almost all endpoints are namespaced, but there are a couple that are not. They ones that are not use the method without a namespace, which avoids the unneeded lookup.

wrangler-service/src/main/java/co/cask/wrangler/service/common/AbstractWranglerHandler.java
albertshau: yep, there is only one at a given time.

wrangler-proto/src/main/java/co/cask/wrangler/proto/connection/Connection.java
albertshau: This was only used for the "admin controlled" connections, which are specified in the app config (ConnectionTypeConfig). Those are connections that should be available in every namespace, so they only have an id and not a namespace.

PR #309, opened 2019-02-12
CDAP-14604 remove non-namespaced endpoints

albertshau:
Removed the non namespaced endpoints now that the UI has migrated
over to the namespaced endpoints.

Also changed endpoints to use explicit transaction control to avoid
unneeded extra transactions now that the services have been migrated
to the storage SPI.

Also fixed exception handling in the directives execution endpoint
to make sure the directive error message is displayed to the user
and not some generic transaction exception message.

### DIFF
--- a/wrangler-service/src/main/java/co/cask/wrangler/service/common/AbstractWranglerHandler.java
+++ b/wrangler-service/src/main/java/co/cask/wrangler/service/common/AbstractWranglerHandler.java
@@ -20,6 +20,7 @@ import co.cask.cdap.api.NamespaceSummary;
 import co.cask.cdap.api.service.http.AbstractSystemHttpServiceHandler;
 import co.cask.cdap.api.service.http.HttpServiceRequest;
 import co.cask.cdap.api.service.http.HttpServiceResponder;
+import co.cask.cdap.spi.data.transaction.TransactionException;
 import co.cask.cdap.spi.data.transaction.TransactionRunners;
 import co.cask.wrangler.dataset.connections.ConnectionNotFoundException;
 import co.cask.wrangler.dataset.connections.ConnectionStore;
--- a/wrangler-service/src/main/java/co/cask/wrangler/service/directive/DirectivesHandler.java
+++ b/wrangler-service/src/main/java/co/cask/wrangler/service/directive/DirectivesHandler.java
@@ -16,6 +16,8 @@
 
 package co.cask.wrangler.service.directive;
 
+import co.cask.cdap.api.annotation.TransactionControl;
+import co.cask.cdap.api.annotation.TransactionPolicy;
 import co.cask.cdap.api.artifact.ArtifactInfo;
 import co.cask.cdap.api.common.Bytes;
 import co.cask.cdap.api.data.schema.Schema;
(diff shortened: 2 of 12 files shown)


wrangler-service/src/main/java/co/cask/wrangler/service/common/AbstractWranglerHandler.java
```diff
@@ -198,6 +199,13 @@ protected Connection getValidatedConnection(ConnectionStore store, NamespacedId
       responder.sendJson(e.getCode(), new ServiceResponse<>(e.getMessage()));
     } catch (JsonSyntaxException e) {
       responder.sendJson(HttpURLConnection.HTTP_BAD_REQUEST, new ServiceResponse<Void>(e.getMessage()));
+    } catch (TransactionException e) {
+      // in case one of the handlers did not propagate exceptions within a transaction correctly.
+      if (e.getCause() != null) {
```
CuriousVini: When would cause be null?

wrangler-service/src/main/java/co/cask/wrangler/service/directive/DirectivesHandler.java
```diff
@@ -1166,7 +1092,7 @@ private static Row createUberRecord(List<Row> rows) {
    * @return records generated from the directives.
    */
   private List<Row> executeDirectives(NamespacedId id, @Nullable Request user,
-                                      Function<List<Row>, List<Row>> sample) {
+                                      Function<List<Row>, List<Row>> sample) throws RecipeException {
```
CuriousVini: Isn't `RecipeException` handled in line 1120?

CuriousVini (commented):
just one minor comment and a question. Rest of the changes lgtm 👍

wrangler-service/src/main/java/co/cask/wrangler/service/directive/DirectivesHandler.java
albertshau: whoops, good catch.

wrangler-service/src/main/java/co/cask/wrangler/service/common/AbstractWranglerHandler.java
albertshau: not sure, maybe if there is an actual transaction problem when talking to the datastore. Haven't actually seen a case myself

albertshau:
@CuriousVini please take another look when you get a chance. I force pushed since it was just removing a throws declaration.

wrangler-service/src/main/java/co/cask/wrangler/service/common/AbstractWranglerHandler.java
CuriousVini: Should not add this check then? I think if this happens, we should fix the handler/whoever is throwing the exception to propagate correct exception rather than handling it here. What do you think?

CuriousVini (approved):
Just one minor question. Code changes lgtm 👍

wrangler-service/src/main/java/co/cask/wrangler/service/common/AbstractWranglerHandler.java
albertshau: I'll remove this catch clause

PR #311, opened 2019-03-11
CDAP-14573 Change to use io.cdap.netty

anwar6953:
Similar to https://github.com/cdapio/cdap/pull/11180, but for other projects.


chtyim (approved):
LGTM

PR #314, opened 2019-03-22
Add CloudSQL and SAP HANA to database options

elfenheart:


wrangler-proto/src/main/java/co/cask/wrangler/proto/db/AllowedDriverInfo.java
```diff
@@ -29,12 +29,14 @@
   private final String name;
   @SerializedName("default.port")
   private final String port;
+  private final boolean basicAllowed;
 
-  public AllowedDriverInfo(String clazz, String label, String tag, String name, String port) {
+  public AllowedDriverInfo(String clazz, String label, String tag, String name, String port, boolean basicAllowed) {
```
bdmogal, line 35: so we know in future, can you add a comment as to why we maintain two classes for what seem like similar things?

wrangler-service/src/main/java/co/cask/wrangler/service/database/DatabaseHandler.java
```diff
@@ -166,6 +168,11 @@ public String getTag() {
     public String getPort() {
       return port;
     }
+
+    public boolean getBasicAllowed() {
```
bdmogal: should be ``isBasicAllowed`` per convention

bdmogal (commented):
couple comments ...

wrangler-proto/src/main/java/co/cask/wrangler/proto/db/AllowedDriverInfo.java
elfenheart: I'm quite unsure why there are two classes for this

bdmogal (approved):

PR #315, opened 2019-03-23
Adding error handling option for records

nitinmotgi:
Support three options

- Skip record on error
- Send to error
- Stop on error


pom.xml
```diff
@@ -19,7 +19,7 @@
 
   <groupId>co.cask.wrangler</groupId>
   <artifactId>wrangler</artifactId>
-  <version>4.0.0-SNAPSHOT</version>
```
sreevatsanraman: We don't need to bump the version yet. since we have not released 4.0.0 version.

sreevatsanraman:
Fix the single comment and merge

pom.xml
nitinmotgi: fixed!

sreevatsanraman (approved):

wrangler-transform/src/main/java/co/cask/wrangler/Wrangler.java
```diff
@@ -431,33 +433,31 @@ public void transform(StructuredRecord input, Emitter<StructuredRecord> emitter)
...
-      // If error threshold is reached, then terminate processing
-      // If threshold is set to -1, it tolerant unlimited errors
-      if (config.threshold != -1 && errorCounter > config.threshold) {
+      getContext().getMetrics().count("failure", 1);
+      if (config.onError.equalsIgnoreCase("send-to-error-port")) {
+        // Emit error record, if the Error flattener or error handlers are not connected, then
+        // the record is automatically omitted.
+        emitter.emitError(new InvalidEntry<>(0, e.getMessage(), input));
+        return;
+      } else if (config.onError.equalsIgnoreCase("fail-pipeline")) {
```
chtyim: No need to have `else` after a return block. This is easier to read:

```java
if (...) {
  ...
  return;
}
if (...) {
  ...
}

File: wrangler-transform/src/main/java/co/cask/wrangler/Wrangler.java [hunk 3]
```diff
@@ -542,35 +542,33 @@ private URL getDPServiceURL(String method) throws URISyntaxException, MalformedU
...
     @Name("schema")
-    @Description("Specifies the schema that has to be output.")
+    @Description("Specifies the schema that has to be output")
     @Macro
     private final String schema;
 
+    @Name("on-error")
+    @Description("How to handle error in record processing")
+    @Macro
+    private final String onError;
```
elfenheart: Can we make this property nullable? It will break UI for a lot of the default configurations

nitinmotgi: Not sure why it would break ? Could you elaborate.

nitinmotgi:
@sreevatsanraman @chtyim addressed PR comments.

wrangler-transform/src/main/java/co/cask/wrangler/Wrangler.java
nitinmotgi: Also, we don't want it to nullable. We have default, so not sure were it would cause problem and why.

albertshau:
This change breaks backwards compatibility by adding a new property that is not nullable.

PR #316, opened 2019-03-25
Added a way to determine if a connection is a built-in connection or …

elfenheart:
…is manually added

JIRA: https://issues.cask.co/browse/CDAP-14385

UI needs to disable the popover (edit, duplicate, delete) for default connections that are preconfigured by the system. This PR is to add a boolean indicator on the connections response to indicate whether the connection is a default system connection or not.

### DIFF
--- a/wrangler-proto/src/main/java/co/cask/wrangler/proto/connection/Connection.java
+++ b/wrangler-proto/src/main/java/co/cask/wrangler/proto/connection/Connection.java
@@ -37,14 +37,23 @@ public final class Connection extends ConnectionMeta {
   // Time in second - when it was last updated.
   private final long updated;
 
+  // Determines if the connection is configured automatically
+  private final boolean preconfigured;
+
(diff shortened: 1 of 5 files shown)


wrangler-proto/src/main/java/co/cask/wrangler/proto/connection/Connection.java
```diff
@@ -37,14 +37,23 @@
   // Time in second - when it was last updated.
   private final long updated;
 
+  // Determines if the connection is configured automatically
+  private final boolean builtIn;
```
sreevatsanraman: How about preconfiguredConnection as the parameter name?

bdmogal: I like it. perhaps just ``preconfigured`` though, since this is the Connection class already?

elfenheart: addressed

CuriousVini (approved):
lgtm 👍

PR #324, opened 2019-05-02
CDAP-14691 fix conflicts in workspace id

albertshau:
Changed workspace creation such that each new workspace is given
an actual unique identifier instead of hashing some non-unique
attribute of the connection.

This fixes bugs where a workspace for a table in one database
could clobber the workspace for a table from a different database
that has the same name, and similar conflict problems with all
of the other connection types.

A side effect of this is that there can now be more than one
workspace for the same data object being wrangled. For example,
a database table can now have more than one namespace. This was
already the case for files, but not for other connection types.


CuriousVini (approved):
lgtm 👍

PR #325, opened 2019-05-06
[CDAP-15338] Fixed TextList extraction and also added a test case

nitinmotgi:


wrangler-core/src/test/java/io/cdap/wrangler/parser/MapArgumentsTest.java
```diff
@@ -110,4 +112,24 @@ public void testMultipleArgumentsOptional() throws Exception {
...
+    CompileStatus status = compiler.compile("remove-sensitive-data :column \"ALL_BASIC\", \"AGE\";");
+
+    UsageDefinition.Builder builder = UsageDefinition.builder("rename");
+    builder.define("column", TokenType.COLUMN_NAME);
+    builder.define("infoTypes", TokenType.TEXT_LIST);
+
+    Iterator<TokenGroup> iterator = status.getSymbols().iterator();
+    Arguments arguments = new MapArguments(builder.build(), iterator.next());
+    Assert.assertEquals(2, arguments.size());
+    Assert.assertEquals(true, arguments.contains("column"));
```
CuriousVini: this can be further simplified. 
```
 Assert.assertTrue(arguments.contains("column"));
 Assert.assertTrue(arguments.contains("infoTypes"));

File: wrangler-core/src/main/java/io/cdap/wrangler/parser/RecipeVisitor.java [hunk 2]
```diff
@@ -310,7 +310,8 @@ public RecipeSymbol getCompiledUnit() {
     List<TerminalNode> strings = ctx.String();
     List<String> strs = new ArrayList<>();
     for (TerminalNode string : strings) {
-      strs.add(string.getText());
+      String text = string.getText();
+      strs.add(text.substring(1, text.length() - 1));
```
CuriousVini, line 314: do we always expect the text to be in the format of `"text"`?

chtyim: What’s the contract of `string.getText()`? Will it ever return an empty string?

wrangler-core/src/test/java/io/cdap/wrangler/parser/MapArgumentsTest.java
nitinmotgi: fixed

wrangler-core/src/main/java/io/cdap/wrangler/parser/RecipeVisitor.java
nitinmotgi: Yes, the contract is defined by the grammar in Directive.g4 (https://github.com/data-integrations/wrangler/blob/develop/wrangler-core/src/main/antlr4/io/cdap/wrangler/parser/Directives.g4#L272)

nitinmotgi:
@CuriousVini & @chtyim could you please take a look at it again.

wrangler-core/src/main/java/io/cdap/wrangler/parser/RecipeVisitor.java
albertshau: if it can return an empty string, it needs to be checked before the substring call

wrangler-core/src/test/java/io/cdap/wrangler/parser/MapArgumentsTest.java
```diff
@@ -110,4 +112,24 @@ public void testMultipleArgumentsOptional() throws Exception {
...
+    UsageDefinition.Builder builder = UsageDefinition.builder("rename");
+    builder.define("column", TokenType.COLUMN_NAME);
+    builder.define("infoTypes", TokenType.TEXT_LIST);
+
+    Iterator<TokenGroup> iterator = status.getSymbols().iterator();
+    Arguments arguments = new MapArguments(builder.build(), iterator.next());
+    Assert.assertEquals(2, arguments.size());
+    Assert.assertTrue(arguments.contains("column"));
+    Assert.assertTrue(arguments.contains("infoTypes"));
+    List<String> infoTypes = ((TextList) arguments.value("infoTypes")).value();
```
albertshau, line 132: seems like the value for "column" should also be checked

wrangler-core/src/main/java/io/cdap/wrangler/parser/RecipeVisitor.java
nitinmotgi: a text without '' or "" will never trigger this. So atleast there will be two characters present. That is defined by the grammar. The state machine wouldn't get you here.

wrangler-core/src/test/java/io/cdap/wrangler/parser/MapArgumentsTest.java
nitinmotgi: done!

albertshau (approved):
lgtm

PR #327, opened 2019-05-08
[CDAP-15256] support decimal type in big query connection

CuriousVini:


wrangler-core/src/main/java/io/cdap/wrangler/utils/Json2Schema.java
```diff
@@ -87,7 +87,8 @@ public Schema toSchema(String id, Row row) throws RecordConvertorException {
       }
 
       if (value instanceof BigDecimal) {
-        Schema schema = Schema.nullableOf(Schema.of(Schema.Type.DOUBLE));
+        BigDecimal decimal = (BigDecimal) value;
+        Schema schema = Schema.nullableOf(Schema.decimalOf(decimal.precision(), decimal.scale()));
```
albertshau: Decimals are one of the reasons determining schema from inspecting values is bad... along with unions. It can very easily be the case that the BigDecimal here happens to have a lower precision than some of the other values. 

I'm not sure there's a good way to handle this without re-designing wrangler... we might have to just always use a big precision number? Please open a jira to handle this better.

pom.xml
```diff
@@ -93,11 +93,11 @@
     <commons-lang3.version>3.5</commons-lang3.version>
     <commons-math3.version>3.6.1</commons-math3.version>
     <geogson.version>1.1.97</geogson.version>
-    <google.cloud.bigquery.version>0.32.0-beta</google.cloud.bigquery.version>
+    <google.cloud.bigquery.version>1.36.0</google.cloud.bigquery.version>
     <google.cloud.core.version>1.15.0</google.cloud.core.version>
     <google.findbugs.version>2.0.1</google.findbugs.version>
     <google.http.client.version>1.22.0</google.http.client.version>
-    <google.storage.version>1.7.0</google.storage.version>
+    <google.storage.version>1.23.0</google.storage.version>
```
albertshau, line 100: did you test the spanner connection? I'm surprised we don't have to update the spanner library version.

albertshau (commented):
Do you know what happens when other directives try to use numeric fields? For example, what happens if it is multiplied in a jexl expression? I imagine it breaks.

Please open a jira to support numeric types in directives.

CuriousVini:
Yes, this PR only focuses on bq connection support for numeric types. Dataprep directives should support this type. Also we should add more directives for conversion. Filed: https://issues.cask.co/browse/CDAP-15358

pom.xml
CuriousVini: yes I did test this for spanner, gcs and bq. There was no runtime exception
Spanner
image

GCS
image

wrangler-core/src/main/java/io/cdap/wrangler/utils/Json2Schema.java
albertshau: Do you know what the maximum precision supported by common data storage system is? For example, BigQuery numeric types is 38. In the short term, I think we should just use the maximum that we know of.

CuriousVini: For dataprep filed: https://issues.cask.co/browse/CDAP-15361

CuriousVini: Since we are only supporting this type in bq connection, I have kept precision as 38.

albertshau (approved):
lgtm

PR #328, opened 2019-05-10
[CDAP-15256] support arrays in bq connection

CuriousVini:
image

image

image

### DIFF
--- a/wrangler-core/src/main/java/io/cdap/wrangler/utils/Json2Schema.java
+++ b/wrangler-core/src/main/java/io/cdap/wrangler/utils/Json2Schema.java
@@ -31,12 +31,14 @@ import io.cdap.wrangler.api.Row;
 import org.json.JSONException;
 
 import java.math.BigDecimal;
+import java.nio.ByteBuffer;
 import java.sql.Time;
 import java.sql.Timestamp;
 import java.time.LocalDate;
 import java.time.LocalTime;
 import java.time.ZonedDateTime;
 import java.util.ArrayList;
+import java.util.Collection;
 import java.util.Date;
 import java.util.List;
 import java.util.Map;
@@ -89,14 +91,13 @@ public final class Json2Schema {
   private Schema getSchema(Object value, String name) throws RecordConvertorException {
     // First, we check if object is of simple type.
     if (value instanceof String || value instanceof Integer || value instanceof Long || value instanceof Short ||
-      value instanceof Double || value instanceof Float || value instanceof Boolean || value instanceof byte[]) {
+      value instanceof Double || value instanceof Float || value instanceof Boolean || value instanceof byte[] ||
+      value instanceof ByteBuffer) {
       try {
         return Schema.nullableOf(new SimpleSchemaGenerator().generate(value.getClass()));
-
       } catch (UnsupportedTypeException e) {
         throw new RecordConvertorException(
-          String.format("Unable to convert field '%s' to basic type.", name)
-        );
+          String.format("Unable to convert field '%s' to basic type.", name), e);
       }
     }
 
@@ -114,7 +115,7 @@ public final class Json2Schema {
     }
 
     if (value instanceof ZonedDateTime) {
-      return  Schema.nullableOf(Schema.of(Schema.LogicalType.TIMESTAMP_MICROS));
+      return Schema.nullableOf(Schema.of(Schema.LogicalType.TIMESTAMP_MICROS));
     }
 
     // TODO - remove all the instaces of java.util.Date once all the directives support LogicalType.
@@ -124,16 +125,16 @@ public final class Json2Schema {
     }
 
     if (value instanceof Map) {
-      return Schema.nullableOf (Schema.mapOf(Schema.of(Schema.Type.STRING), Schema.of(Schema.Type.STRING)));
+      return Schema.nullableOf(Schema.mapOf(Schema.of(Schema.Type.STRING), Schema.of(Schema.Type.STRING)));
     }
 
     if (value instanceof JsonElement) {
       return toSchema(name, (JsonElement) value);
     }
 
-    if (value instanceof List) {
-      List<?> list = (List) value;
-      for (Object listObject : list) {
+    if (value instanceof Collection) {
+      Collection<?> collection = (Collection) value;
+      for (Object listObject : collection) {
         Schema schema = getSchema(listObject, name);
         if (schema != null) {
           return Schema.nullableOf(Schema.arrayOf(schema));
@@ -192,8 +193,7 @@ public final class Json2Schema {
       return Schema.nullableOf(new SimpleSchemaGenerator().generate(value.getClass()));
     } catch (UnsupportedTypeException e) {
       throw new RecordConvertorException(
-        String.format("Unable to convert field '%s' to basic type.", name)
-      );
+        String.format("Unable to convert field '%s' to basic type.", name), e);
     }
   }
 
--- a/wrangler-core/src/main/java/io/cdap/wrangler/utils/Json2Schema.java
+++ b/wrangler-core/src/main/java/io/cdap/wrangler/utils/Json2Schema.java
@@ -135,10 +135,16 @@ public final class Json2Schema {
     if (value instanceof Collection) {
       Collection<?> collection = (Collection) value;
       for (Object listObject : collection) {
-        Schema schema = getSchema(listObject, name);
-        if (schema != null) {
-          return Schema.nullableOf(Schema.arrayOf(schema));
+        if (listObject == null) {
+          continue;
         }
+
+        Schema schema = getSchema(listObject, name);
+        // this means schema is unknown and is not supported.
+        if (schema == null) {
+          return null;
+        }
+        return Schema.nullableOf(Schema.arrayOf(schema));
       }
     }
 
--- a/wrangler-core/src/main/java/io/cdap/wrangler/utils/Json2Schema.java
+++ b/wrangler-core/src/main/java/io/cdap/wrangler/utils/Json2Schema.java
@@ -73,62 +73,75 @@ public final class Json2Schema {
       String name = column.getFirst();
       Object value = column.getSecond();
 
-      // First, we check if object is of simple type.
-      if (value instanceof String || value instanceof Integer || value instanceof Long || value instanceof Short ||
-          value instanceof Double || value instanceof Float || value instanceof Boolean || value instanceof byte[]) {
-        try {
-          Schema schema = Schema.nullableOf(new SimpleSchemaGenerator().generate(value.getClass()));
-          fields.add(Schema.Field.of(name, schema));
-        } catch (UnsupportedTypeException e) {
-          throw new RecordConvertorException(
-            String.format("Unable to convert field '%s' to basic type.", name)
-          );
-        }
-      }
-
-      if (value instanceof BigDecimal) {
-        // TODO CDAP-15361 precision should be derived from the schema of the row.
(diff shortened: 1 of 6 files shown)


wrangler-core/src/main/java/io/cdap/wrangler/utils/Json2Schema.java
```diff
@@ -73,62 +73,75 @@ public Schema toSchema(String id, Row row) throws RecordConvertorException {
       String name = column.getFirst();
       Object value = column.getSecond();
 
-      // First, we check if object is of simple type.
-      if (value instanceof String || value instanceof Integer || value instanceof Long || value instanceof Short ||
-          value instanceof Double || value instanceof Float || value instanceof Boolean || value instanceof byte[]) {
-        try {
-          Schema schema = Schema.nullableOf(new SimpleSchemaGenerator().generate(value.getClass()));
+      if (value instanceof Schema) {
```
albertshau, line 78: when would the value be a Schema?

wrangler-core/src/main/java/io/cdap/wrangler/utils/RecordConvertor.java
```diff
@@ -401,7 +401,7 @@ private Object decodeUnion(String name, Object object, List<Schema> schemas) thr
 
   private List<Object> decodeArray(String name, List list, Schema schema) throws RecordConvertorException {
     List<Object> array = Lists.newArrayListWithCapacity(list.size());
-    for (Object object : array) {
+    for (Object object : list) {
```
albertshau, line 398: :)

wrangler-core/src/main/java/io/cdap/wrangler/utils/Json2Schema.java
```diff
@@ -73,62 +73,75 @@ public Schema toSchema(String id, Row row) throws RecordConvertorException {
...
+  }
 
-      if (value instanceof BigDecimal) {
-        // TODO CDAP-15361 precision should be derived from the schema of the row.
-        Schema schema = Schema.nullableOf(Schema.decimalOf(38, ((BigDecimal) value).scale()));
-        fields.add(Schema.Field.of(name, schema));
+  @Nullable
+  private Schema getSchema(Object value, String name) throws RecordConvertorException {
+    // First, we check if object is of simple type.
+    if (value instanceof String || value instanceof Integer || value instanceof Long || value instanceof Short ||
```
albertshau, line 93: not from this PR, but this is missing ByteBuffer

wrangler-core/src/main/java/io/cdap/wrangler/utils/Json2Schema.java
```diff
@@ -73,62 +73,75 @@ public Schema toSchema(String id, Row row) throws RecordConvertorException {
...
+      value instanceof Double || value instanceof Float || value instanceof Boolean || value instanceof byte[]) {
+      try {
+        return Schema.nullableOf(new SimpleSchemaGenerator().generate(value.getClass()));
+
+      } catch (UnsupportedTypeException e) {
+        throw new RecordConvertorException(
+          String.format("Unable to convert field '%s' to basic type.", name)
+        );
```
albertshau: not from this PR, but the exception should have 'e' as the cause.

wrangler-core/src/main/java/io/cdap/wrangler/utils/Json2Schema.java
```diff
@@ -73,62 +73,75 @@ public Schema toSchema(String id, Row row) throws RecordConvertorException {
...
+    }
 
-      // TODO - remove all the instaces of java.util.Date once all the directives support LogicalType.
-      if (value instanceof Date || value instanceof java.sql.Date || value instanceof Time
-        || value instanceof Timestamp) {
-        Schema schema = Schema.nullableOf(Schema.of(Schema.Type.LONG));
-        fields.add(Schema.Field.of(name, schema));
-      }
+    if (value instanceof ZonedDateTime) {
+      return  Schema.nullableOf(Schema.of(Schema.LogicalType.TIMESTAMP_MICROS));
```
albertshau: extra space after 'return'

wrangler-core/src/main/java/io/cdap/wrangler/utils/Json2Schema.java
```diff
@@ -73,62 +73,75 @@ public Schema toSchema(String id, Row row) throws RecordConvertorException {
...
+      || value instanceof Timestamp) {
+      return Schema.nullableOf(Schema.of(Schema.Type.LONG));
+    }
 
-      if (value instanceof JsonElement) {
-        Schema schema = toSchema(name, (JsonElement) value);
-        fields.add(Schema.Field.of(name, schema));
-      }
+    if (value instanceof Map) {
+      return Schema.nullableOf (Schema.mapOf(Schema.of(Schema.Type.STRING), Schema.of(Schema.Type.STRING)));
```
albertshau: extra space after nullableOf

albertshau: Yet another reason wrangler shouldn't be examining values to determine schema. This is incorrect for maps that aren't string -> string. And there is simply no way to determine the schema if the map is empty.

You don't need to fix in this PR.

wrangler-core/src/main/java/io/cdap/wrangler/utils/Json2Schema.java
```diff
@@ -73,62 +73,75 @@ public Schema toSchema(String id, Row row) throws RecordConvertorException {
...
+    }
 
-      if (value instanceof Schema) {
-        fields.addAll(((Schema) value).getFields());
+    if (value instanceof JsonElement) {
+      return toSchema(name, (JsonElement) value);
+    }
+
+    if (value instanceof List) {
```
albertshau: I think it's better to check against Collection instead of just List.

wrangler-core/src/main/java/io/cdap/wrangler/utils/Json2Schema.java
```diff
@@ -73,62 +73,75 @@ public Schema toSchema(String id, Row row) throws RecordConvertorException {
...
+      List<?> list = (List) value;
+      for (Object listObject : list) {
```
albertshau: if the list is empty, a null schema is returned. There's not really anything you can do about this, but do you know what that looks like in the UI?

wrangler-core/src/main/java/io/cdap/wrangler/utils/Json2Schema.java
```diff
@@ -73,62 +73,75 @@ public Schema toSchema(String id, Row row) throws RecordConvertorException {
...
+        Schema schema = getSchema(listObject, name);
+        if (schema != null) {
```
albertshau: if null, that means the object is something we can't express with the CDAP schema, which means it is at least a union of this unknown type and some other type. In this situation, I think it's better to just return a null schema rather than getting the type of the next value.

wrangler-core/src/main/java/io/cdap/wrangler/utils/Json2Schema.java
```diff
@@ -73,62 +73,75 @@ public Schema toSchema(String id, Row row) throws RecordConvertorException {
...
+          return Schema.nullableOf(Schema.arrayOf(schema));
```
albertshau: Doesn't the BQ sink fail if the array is a nullable array? 

Generally it's best if collections themselves are not nullable, with empty collections used instead of null, but I don't know if the directives have been following that.

wrangler-service/src/main/java/io/cdap/wrangler/service/bigquery/BigQueryHandler.java
```diff
@@ -429,6 +397,41 @@ public void specification(HttpServiceRequest request, HttpServiceResponder respo
...
+        BigDecimal decimal = fieldValue.getNumericValue();
+        if (decimal.scale() < 9) {
+          // scale up the big decimal. this is because structured record expects scale to be exactly same as schema
+          // Big Query supports maximum unscaled value up to 38 digits. so scaling up should still be <= max
+          // precision
+          decimal = decimal.setScale(9);
+        }
+        return decimal;
+
+      case DATETIME:
```
albertshau, line 419: unrelated note: it seems like we need to support a datetime type in CDAP, otherwise there is no way the BQ sink can create a table with datetime column type. For example, if somebody wants to copy the contents of one BQ table to another BQ table. Can you open a jira for this?

wrangler-service/src/main/java/io/cdap/wrangler/service/bigquery/BigQueryHandler.java
```diff
@@ -429,6 +397,41 @@ public void specification(HttpServiceRequest request, HttpServiceResponder respo
...
+      case STRING:
+        return  fieldValue.getStringValue();
```
albertshau: extra space after return

wrangler-core/src/main/java/io/cdap/wrangler/utils/Json2Schema.java
CuriousVini: yea even I am not sure. But that is part of restructuring of existing code.
https://github.com/data-integrations/wrangler/blob/release/4.0/wrangler-core/src/main/java/io/cdap/wrangler/utils/Json2Schema.java#L127

wrangler-core/src/main/java/io/cdap/wrangler/utils/Json2Schema.java
CuriousVini: yes bq sink would fail if schema for array or array component is null. However, we can modify the schema from pipeline UI to make array and its component to be non-nullable. 
Array is kept as nullable for other cases.

wrangler-service/src/main/java/io/cdap/wrangler/service/bigquery/BigQueryHandler.java
CuriousVini: Issue: https://issues.cask.co/browse/CDAP-15370

wrangler-core/src/main/java/io/cdap/wrangler/utils/Json2Schema.java
CuriousVini: This is for the case when first few values of the list are null but later actual values are present

wrangler-core/src/main/java/io/cdap/wrangler/utils/Json2Schema.java
CuriousVini: If schema is null, we do not add that field schema into schema in line 83. Meaning, UI will show type as unknown. Which is same as current behavior. In that case UI shows type as `unknown`.

image

image

wrangler-core/src/main/java/io/cdap/wrangler/utils/Json2Schema.java
albertshau: then it should be a check on listObject being null, not the schema of listObject.

```
if (listObject == null) {
  continue;
}
Schema schema = getSchema(listObject, name);
if (schema == null) {
  return null;
}
...
```

CuriousVini: Current implementation relies on the fact that if value object does not match with any of the supported types, schema will be null for item and move to next (similar to what we had before). It does not check null values. Modified it now.

albertshau (approved):
lgtm

CuriousVini:
squashed commits.

PR #335, opened 2019-05-23
[CDAP-15429] Updated description of encoding type of ParseAvro directive

miguelvelezmj25:
https://issues.cask.co/browse/CDAP-15429

### DIFF
--- a/wrangler-core/src/main/java/io/cdap/directives/parser/ParseAvro.java
+++ b/wrangler-core/src/main/java/io/cdap/directives/parser/ParseAvro.java
@@ -56,7 +56,7 @@ import java.util.concurrent.ExecutionException;
 import java.util.concurrent.TimeUnit;
 
 /**
- * A step to parse AVRO json or binary format.
+ * A step to parse json or binary encoded AVRO.
  */
 @Plugin(type = Directive.TYPE)
 @Name("parse-as-avro")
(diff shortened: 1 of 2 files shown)


miguelvelezmj25:
Build at https://builds.cask.co/browse/HYP-WT316

wrangler-core/src/main/java/io/cdap/directives/parser/ParseAvro.java
```diff
@@ -77,7 +77,7 @@ public UsageDefinition define() {
     UsageDefinition.Builder builder = UsageDefinition.builder(NAME);
     builder.define("column", TokenType.COLUMN_NAME);
     builder.define("schema-id", TokenType.IDENTIFIER);
-    builder.define("encode-type", TokenType.IDENTIFIER);
+    builder.define("encoding-type", TokenType.IDENTIFIER);
```
albertshau, line 80: this is a backwards incompatible change and cannot be done. If somebody upgrades a pipeline that is using 'encode-type' it will now break.

miguelvelezmj25: Is there a compatible way to make sure that there is no confusion regarding the encoding type?

albertshau: we can update documentation but we cannot make backwards incompatible API changes

miguelvelezmj25: So I can change the javadoc and the README.md file. I think those are the only places where this info is documented.

nitinmotgi: These are used as usage definition and are really not stored anywhere other than in Directive. So the change is fully compatible.

PR #338, opened 2019-05-25
[CDAP-15437] Fixed User Defined Directive (UDD) class isolation

nitinmotgi:
Fixed User Defined Directive (UDD) class isolation problem in DirectiveServiceHandler, upgraded GCP service versions and improved documentation around the classes that were touched.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -83,7 +83,7 @@
     <aws.sdk.version>1.11.133</aws.sdk.version>
     <bigquery.connector.hadoop2.version>0.10.2-hadoop2</bigquery.connector.hadoop2.version>
     <bouncycastle.version>1.56</bouncycastle.version>
-    <cdap.version>6.1.0-SNAPSHOT</cdap.version>
+    <cdap.version>6.0.1-SNAPSHOT</cdap.version>
     <chlorine.version>1.1.5</chlorine.version>
     <commons.validator.version>1.6</commons.validator.version>
     <commons-io.version>2.5</commons-io.version>
@@ -93,11 +93,11 @@
     <commons-lang3.version>3.5</commons-lang3.version>
     <commons-math3.version>3.6.1</commons-math3.version>
     <geogson.version>1.1.97</geogson.version>
-    <google.cloud.bigquery.version>1.36.0</google.cloud.bigquery.version>
-    <google.cloud.core.version>1.15.0</google.cloud.core.version>
+    <google.cloud.bigquery.version>1.75.0</google.cloud.bigquery.version>
+    <google.cloud.core.version>1.75.0</google.cloud.core.version>
     <google.findbugs.version>2.0.1</google.findbugs.version>
     <google.http.client.version>1.22.0</google.http.client.version>
-    <google.storage.version>1.23.0</google.storage.version>
+    <google.storage.version>1.75.0</google.storage.version>
     <gson.version>2.6.2</gson.version>
     <guava.retrying.version>2.0.0</guava.retrying.version>
     <guava.version>20.0</guava.version>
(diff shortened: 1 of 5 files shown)


pom.xml
```diff
@@ -93,11 +93,11 @@
     <commons-lang3.version>3.5</commons-lang3.version>
     <commons-math3.version>3.6.1</commons-math3.version>
     <geogson.version>1.1.97</geogson.version>
-    <google.cloud.bigquery.version>1.36.0</google.cloud.bigquery.version>
-    <google.cloud.core.version>1.15.0</google.cloud.core.version>
+    <google.cloud.bigquery.version>1.75.0</google.cloud.bigquery.version>
```
albertshau, line 96: have we verified that this doesn't break anything?

wrangler-core/src/main/java/io/cdap/wrangler/registry/UserDirectiveRegistry.java
```diff
@@ -112,28 +116,25 @@ public UserDirectiveRegistry(StageContext context) {
...
-          DirectiveInfo classz = new DirectiveInfo(DirectiveInfo.Scope.USER, directive);
-          namespaceRegistry.put(classz.name(), classz);
-          return classz;
-        }
+      if (context != null) {
+        directive = context.loadPluginClass(name);
       } else {
-        return namespaceRegistry.get(name);
+        if (manager != null) {
+          PluginConfigurer configurer = manager.createPluginConfigurer(namespace);
```
albertshau, line 125: I believe this is a fairly expensive operation. Do you notice slowness when many directives are used? Though I guess it's probably similar to the previous implementation that loads all plugin artifacts.

pom.xml
```diff
@@ -83,7 +83,7 @@
     <aws.sdk.version>1.11.133</aws.sdk.version>
     <bigquery.connector.hadoop2.version>0.10.2-hadoop2</bigquery.connector.hadoop2.version>
     <bouncycastle.version>1.56</bouncycastle.version>
-    <cdap.version>6.1.0-SNAPSHOT</cdap.version>
+    <cdap.version>6.0.1-SNAPSHOT</cdap.version>
```
albertshau: how come this is needed?

pom.xml
nitinmotgi: Yes Albert, I ran my full 'root' tests with this change.

wrangler-core/src/main/java/io/cdap/wrangler/registry/UserDirectiveRegistry.java
nitinmotgi: I did see a bit of slowness as we don't cache it anymore. I was thinking of caching it at this point to make sure it's faster in subsequent operations, but it happens only once so I decided not to do it.

pom.xml
nitinmotgi: Ah! will change this. When I was going the change on Saturday, the issue was 'develop' builds hadn't passed yet, so I used the release that was available to make progress. I will change this to 6.1.0-SNAPSHOT. Thanks for catching it.

nitinmotgi:
@albertshau comments and changes in. Could you please review when u get a chance.

wrangler-core/src/main/java/io/cdap/wrangler/registry/UserDirectiveRegistry.java
albertshau: The correct thing would be for the service endpoint to create one PluginConfigurer, and for that same instance to be used throughout the lifetime of the call. This is a bigger refactor so I've opened https://issues.cask.co/browse/CDAP-15454 to revisit it later.

pom.xml
albertshau: OK, though in the future it's better if this were done in a separate PR, as it is not related to classloading isolation.

albertshau (approved):
lgtm, please squash before merging

nitinmotgi:
Squashed and merged.

PR #345, opened 2019-08-13
[CDAP-15499] Fix incomplete error message for driver class

CuriousVini:
Screen Shot 2019-08-12 at 8 52 26 PM

Screen Shot 2019-08-12 at 8 56 11 PM

### DIFF
--- a/wrangler-service/src/main/java/io/cdap/wrangler/service/database/DatabaseHandler.java
+++ b/wrangler-service/src/main/java/io/cdap/wrangler/service/database/DatabaseHandler.java
@@ -609,6 +609,7 @@ public class DatabaseHandler extends AbstractWranglerHandler {
    * @param connection the connection to be connected to.
    * @return pair of connection and the driver cleanup.
    */
+  @SuppressWarnings("unchecked")
   private DriverCleanup loadAndExecute(Namespace namespace, ConnectionMeta connection,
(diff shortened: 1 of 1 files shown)


wrangler-service/src/main/java/io/cdap/wrangler/service/database/DatabaseHandler.java
```diff
@@ -619,8 +620,13 @@ private DriverCleanup loadAndExecute(Namespace namespace, ConnectionMeta connect
     NamespacedId key = new NamespacedId(namespace, name);
     ClassLoader classLoader = cache.get(key);
     try {
-      @SuppressWarnings("unchecked")
-      Class<? extends Driver> driverClass = (Class<? extends Driver>) classLoader.loadClass(classz);
+      Class<? extends Driver> driverClass;
+      try {
+        driverClass = (Class<? extends Driver>) classLoader.loadClass(classz);
+      } catch (ClassNotFoundException e) {
+        throw new RuntimeException(String.format("Driver class '%s' is not available in the uploaded jar. " +
```
albertshau, line 627: this talks about a jar, but the user hasn't done anything with a jar when they hit this problem. This should be some message about the jdbc driver not being installed correctly.

This also isn't the best place to fix the problem, since the problem was during the jdbc driver upload and not here. We should error out when somebody loads an artifact and specifies a plugin that uses a class that doesn't exist in the artifact.

PR #351, opened 2019-09-03
Improve error handling in wrangler

CuriousVini:
**Summary**
* Removed usage of toString() from error messages
* Made error messages consistent for Directive's `initialize()` and `execute()` methods
* Using `JexlException#getInfo()` method to construct error message. This is to avoid having classname that threw exception in the error message, i.e. `io.cdap.wrangler.expression.EL`
* Also did some code cleanup and error message cleanup in directives

image

image

Custom Transforms:
image

### DIFF
--- a/wrangler-core/src/main/java/io/cdap/directives/aggregates/IncrementTransientVariable.java
+++ b/wrangler-core/src/main/java/io/cdap/directives/aggregates/IncrementTransientVariable.java
@@ -71,7 +71,7 @@ public class IncrementTransientVariable implements Directive {
     try {
       el.compile(expression);
     } catch (ELException e) {
-      throw new DirectiveParseException(e.getMessage());
+      throw new DirectiveParseException(NAME, e.getMessage(), e);
     }
   }
 
@@ -100,7 +100,7 @@ public class IncrementTransientVariable implements Directive {
           context.getTransientStore().increment(TransientVariableScope.GLOBAL, variable, incrementBy);
         }
       } catch (ELException e) {
-        throw new DirectiveExecutionException(e.getMessage());
+        throw new DirectiveExecutionException(NAME, e.getMessage(), e);
       }
     }
     return rows;
--- a/wrangler-core/src/main/java/io/cdap/directives/column/Copy.java
+++ b/wrangler-core/src/main/java/io/cdap/directives/column/Copy.java
@@ -30,8 +30,6 @@ import io.cdap.wrangler.api.annotations.Categories;
 import io.cdap.wrangler.api.parser.ColumnName;
 import io.cdap.wrangler.api.parser.TokenType;
 import io.cdap.wrangler.api.parser.UsageDefinition;
-import io.cdap.wrangler.i18n.Messages;
-import io.cdap.wrangler.i18n.MessagesFactory;
 
 import java.util.List;
 
@@ -43,7 +41,6 @@ import java.util.List;
 @Categories(categories = { "column"})
 @Description("Copies values from a source column into a destination column.")
 public class Copy implements Directive {
-  private static final Messages MSG = MessagesFactory.getMessages();
   public static final String NAME = "copy";
   private ColumnName source;
   private ColumnName destination;
@@ -77,7 +74,7 @@ public class Copy implements Directive {
     for (Row row : rows) {
       int sidx = row.find(source.value());
       if (sidx == -1) {
-        throw new DirectiveExecutionException(MSG.get("column.not.found", toString(), source.value()));
+        throw new DirectiveExecutionException(NAME, String.format("Column '%s' does not exist.", source.value()));
       }
 
       int didx = row.find(destination.value());
@@ -93,8 +90,9 @@ public class Copy implements Directive {
         // if destination column exists, and force is set to false, then throw exception, else
         // overwrite it.
         if (!force) {
-          throw new DirectiveExecutionException(toString() + " : Destination column '" + destination.value()
-                                    + "' does not exist in the row. Use 'force' option to add new column.");
+          throw new DirectiveExecutionException(
+            NAME, String.format("Destination column '%s' already exists in the row. Use 'force' " +
+                                  "option to overwrite the column.", destination.value()));
         }
         row.setValue(didx, row.getValue(sidx));
       }
--- a/wrangler-core/src/main/java/io/cdap/directives/column/Rename.java
+++ b/wrangler-core/src/main/java/io/cdap/directives/column/Rename.java
@@ -74,11 +74,9 @@ public final class Rename implements Directive {
           row.setColumn(idx, target.value());
         } else {
           throw new DirectiveExecutionException(
-            String.format(
-              "%s : %s column already exists. Apply the directive 'drop %s' before renaming %s to %s.", toString(),
-              target.value(), target.value(), source.value(), source.value()
-            )
-          );
+            NAME, String.format("Column '%s' already exists. Apply the 'drop %s' directive before " +
+                                  "renaming '%s' to '%s'.",
+                                target.value(), target.value(), source.value(), target.value()));
         }
       }
     }
--- a/wrangler-core/src/main/java/io/cdap/directives/column/SetType.java
+++ b/wrangler-core/src/main/java/io/cdap/directives/column/SetType.java
@@ -79,8 +79,8 @@ public final class SetType implements Directive {
         } catch (DirectiveExecutionException e) {
           throw e;
         } catch (Exception e) {
-          throw new DirectiveExecutionException(String.format("Invalid data: Column '%s' can not be converted to '%s'",
-                                                              col, type));
+          throw new DirectiveExecutionException(
+            NAME, String.format("Column '%s' cannot be converted to a '%s'.", col, type), e);
         }
       }
     }
@@ -162,11 +162,11 @@ public final class SetType implements Directive {
         } else if (object instanceof Float) {
           return ((Float) object) > 0 ? true : false;
         } else if (object instanceof Double) {
-          return ((Double) object)  > 0 ? true : false;
+          return ((Double) object) > 0 ? true : false;
         } else if (object instanceof Integer) {
-          return ((Integer) object)  > 0 ? true : false;
+          return ((Integer) object) > 0 ? true : false;
         } else if (object instanceof Long) {
-          return ((Long) object)  > 0 ? true : false;
+          return ((Long) object) > 0 ? true : false;
         } else if (object instanceof byte[]) {
           return Bytes.toBoolean((byte[]) object);
         } else {
@@ -258,9 +258,9 @@ public final class SetType implements Directive {
 
       default:
         throw new DirectiveExecutionException(
-          String.format("Unknown data type '%s' found in the directive. " +
-                  "Accepted types are: int, short, long, double, boolean, string, bytes", toType)
-        );
+          NAME, String.format(
+            "Column '%s' is of unsupported type '%s'. Supported types are: " +
+              "int, short, long, double, boolean, string, bytes", col, toType));
     }
   }
 }
--- a/wrangler-core/src/main/java/io/cdap/directives/column/SplitToColumns.java
+++ b/wrangler-core/src/main/java/io/cdap/directives/column/SplitToColumns.java
@@ -76,20 +76,25 @@ public class SplitToColumns implements Directive {
       int idx = row.find(column);
       if (idx != -1) {
         Object object = row.getValue(idx);
-        if (object instanceof String) {
-          String[] lines = ((String) object).split(regex);
-          int i = 1;
-          for (String line : lines) {
-            row.add(String.format("%s_%d", column, i), line);
-            ++i;
-          }
-          results.add(row);
-        } else {
+
+        if (object == null) {
           throw new DirectiveExecutionException(
-            String.format("%s : Invalid type '%s' of column '%s'. Should be of type String.", toString(),
-                          object != null ? object.getClass().getName() : "null", column)
-          );
+            NAME, String.format("Column '%s' has null value. It should be a non-null 'String'.", column));
         }
+
+        if (!(object instanceof String)) {
+          throw new DirectiveExecutionException(
+            NAME, String.format("Column '%s' has invalid type '%s'. It should be of type 'String'.",
+                                column, object.getClass().getSimpleName()));
+        }
+
+        String[] lines = ((String) object).split(regex);
+        int i = 1;
+        for (String line : lines) {
+          row.add(String.format("%s_%d", column, i), line);
+          ++i;
+        }
+        results.add(row);
       }
     }
     return results;
--- a/wrangler-core/src/main/java/io/cdap/directives/column/Swap.java
+++ b/wrangler-core/src/main/java/io/cdap/directives/column/Swap.java
@@ -29,8 +29,6 @@ import io.cdap.wrangler.api.annotations.Categories;
 import io.cdap.wrangler.api.parser.ColumnName;
 import io.cdap.wrangler.api.parser.TokenType;
 import io.cdap.wrangler.api.parser.UsageDefinition;
-import io.cdap.wrangler.i18n.Messages;
-import io.cdap.wrangler.i18n.MessagesFactory;
 
 import java.util.List;
 
@@ -43,7 +41,6 @@ import java.util.List;
 @Description("Swaps the column names of two columns.")
 public class Swap implements Directive {
   public static final String NAME = "swap";
-  private static final Messages MSG = MessagesFactory.getMessages();
   private String left;
   private String right;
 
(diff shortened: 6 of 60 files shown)


wrangler-core/src/main/java/io/cdap/directives/aggregates/IncrementTransientVariable.java
```diff
@@ -71,7 +71,7 @@ public void initialize(Arguments args) throws DirectiveParseException {
     try {
       el.compile(expression);
     } catch (ELException e) {
-      throw new DirectiveParseException(e.getMessage());
+      throw new DirectiveParseException("Error encountered while initializing '" + NAME + "' : " + e.getMessage());
```
chtyim: When wrapping exception, always includes the origin exception cause. Otherwise debugging would be extremely difficult.

wrangler-core/src/main/java/io/cdap/directives/column/SplitToColumns.java
```diff
@@ -84,10 +84,15 @@ public void destroy() {
             ++i;
           }
           results.add(row);
+        } else if (object == null) {
+          throw new DirectiveExecutionException(
+            String.format("Error encountered while executing '%s' : Column '%s' has null value. " +
+                            "It should be of type 'String'.", NAME, column)
+          );
         } else {
```
chtyim: You don't need `else` if the previous block throws.

wrangler-core/src/main/java/io/cdap/directives/lookup/TableLookup.java
```diff
@@ -101,10 +104,15 @@ private void ensureInitialized(ExecutorContext context) throws DirectiveExecutio
         continue;
       }
       Object object = row.getValue(idx);
-      if (!(object instanceof String)) {
+      if (object == null) {
+        throw new DirectiveExecutionException(
+          String.format("Error encountered while executing '%s' : Column '%s' has null value. " +
+                          "It should be of type 'String'.", NAME, column)
+        );
+      } else if (!(object instanceof String)) {
```
chtyim: Same here, no need to have `else`. It's easier to read if the logic is:

```java
if (object == null) {
  throw ...
}
if (!(object instanceof String)) {
  throw ...
}

File: wrangler-core/src/main/java/io/cdap/directives/column/Copy.java [hunk 4]
```diff
@@ -93,8 +92,9 @@ public void destroy() {
         // if destination column exists, and force is set to false, then throw exception, else
         // overwrite it.
         if (!force) {
-          throw new DirectiveExecutionException(toString() + " : Destination column '" + destination.value()
-                                    + "' does not exist in the row. Use 'force' option to add new column.");
+          throw new DirectiveExecutionException(
+            String.format("Error encountered while executing '%s' : Destination column '%s' does not exist " +
```
albertshau: should this be 'already exists' instead of 'does not exist'?

wrangler-core/src/main/java/io/cdap/directives/column/Rename.java
```diff
@@ -75,8 +75,9 @@ public void destroy() {
         } else {
           throw new DirectiveExecutionException(
             String.format(
-              "%s : %s column already exists. Apply the directive 'drop %s' before renaming %s to %s.", toString(),
-              target.value(), target.value(), source.value(), source.value()
+              "Error encountered while executing '%s' : Column '%s' already exists. " +
+                "Apply the directive 'drop %s' before renaming '%s' to '%s'.",
```
albertshau: nit: Apply the 'drop %s' directive before renaming ...

wrangler-core/src/main/java/io/cdap/directives/column/SetType.java
```diff
@@ -79,8 +79,9 @@ public void destroy() {
         } catch (DirectiveExecutionException e) {
           throw e;
         } catch (Exception e) {
-          throw new DirectiveExecutionException(String.format("Invalid data: Column '%s' can not be converted to '%s'",
-                                                              col, type));
+          throw new DirectiveExecutionException(
+            String.format("Error encountered while executing '%s' : Column '%s' can not be converted to '%s'.",
```
albertshau: Column '%s' cannot be converted to a '%s'.

wrangler-core/src/main/java/io/cdap/directives/column/SetType.java
```diff
@@ -79,8 +79,9 @@ public void destroy() {
         } catch (DirectiveExecutionException e) {
           throw e;
         } catch (Exception e) {
-          throw new DirectiveExecutionException(String.format("Invalid data: Column '%s' can not be converted to '%s'",
-                                                              col, type));
+          throw new DirectiveExecutionException(
```
albertshau, line 82: These error messages all seem to follow the same pattern. It may make sense to add another constructor that takes the directive name and a submessage, so that the full message is forced to be the same everywhere.

wrangler-core/src/main/java/io/cdap/directives/column/Swap.java
```diff
@@ -43,7 +43,6 @@
 @Description("Swaps the column names of two columns.")
 public class Swap implements Directive {
   public static final String NAME = "swap";
-  private static final Messages MSG = MessagesFactory.getMessages();
```
albertshau, line 46: are there imports that can be removed now?

wrangler-core/src/main/java/io/cdap/directives/column/Swap.java
```diff
@@ -73,11 +72,15 @@ public void destroy() {
       int didx = row.find(right);
 
       if (sidx == -1) {
-        throw new DirectiveExecutionException(MSG.get("column.not.found", toString(), left));
+        throw new DirectiveExecutionException(
+          String.format("Error encountered while executing '%s' : Column '%s' does not exist in the row.",
```
albertshau: to be more consistent with the other messages, "Column '%s' does not exist".

wrangler-core/src/main/java/io/cdap/directives/lookup/TableLookup.java
```diff
@@ -101,10 +104,15 @@ private void ensureInitialized(ExecutorContext context) throws DirectiveExecutio
         continue;
       }
       Object object = row.getValue(idx);
-      if (!(object instanceof String)) {
+      if (object == null) {
+        throw new DirectiveExecutionException(
+          String.format("Error encountered while executing '%s' : Column '%s' has null value. " +
+                          "It should be of type 'String'.", NAME, column)
```
albertshau: 'It should be a non-null String.'

wrangler-core/src/main/java/io/cdap/directives/parser/FixedLengthParser.java
```diff
@@ -123,10 +123,15 @@ public void destroy() {
             results.add(newRow);
             index = (index + recordLength);
           }
+        } else if (object == null) {
+          throw new DirectiveExecutionException(
+            String.format("Error encountered while executing '%s' : Column '%s' has null value. " +
+                            "It should be of type 'String'.", NAME, col)
```
albertshau: It should be a non-null String

wrangler-core/src/main/java/io/cdap/directives/parser/HL7Parser.java
```diff
@@ -109,15 +109,22 @@ public void destroy() {
             HL7MessageVisitor visitor = new HL7MessageVisitor(row, column + "_hl7", depth);
             MessageVisitors.visit(message,
                                   MessageVisitors.visitPopulatedElements(visitor)).getDelegate();
+          } else if (object == null) {
+            throw new DirectiveExecutionException(
+              String.format("Error encountered while executing '%s' : Column '%s' has null value. " +
+                              "It should be of type 'String'.", NAME, column)
```
albertshau: It should be a non-null String

wrangler-core/src/main/java/io/cdap/directives/parser/ParseAvro.java
```diff
@@ -88,7 +88,9 @@ public void initialize(Arguments args) throws DirectiveParseException {
     this.schemaId = ((Identifier) args.value("schema-id")).value();
     this.type = ((Identifier) args.value("encode-type")).value();
     if (!"json".equalsIgnoreCase(type) && !"binary".equalsIgnoreCase(type)) {
-      throw new DirectiveParseException("Parsing AVRO can be either of type 'json' or 'binary'");
+      throw new DirectiveParseException(
+        String.format("Error encountered while initializing '%s' : " +
+                        "Parsing AVRO can be either of type 'json' or 'binary'.", NAME));
```
albertshau: Invalid encoding type '%s'. The type must be either 'json' or 'binary'.

wrangler-core/src/main/java/io/cdap/directives/parser/ParseAvro.java
```diff
@@ -143,15 +145,14 @@ public void destroy() {
         if (decoder != null) {
           decoderInitialized = true;
         } else {
```
albertshau: nit: this is existing code, but it's more readable to do:

```
if (decoder == null) {
  throw ...
}
decoderInitialized = true;
```

wrangler-core/src/main/java/io/cdap/directives/parser/ParseLog.java
```diff
@@ -89,10 +89,16 @@ public void destroy() {
           log = (String) object;
         } else if (object instanceof byte[]) {
           log = new String((byte[]) object);
+        } else if (object == null) {
+          throw new DirectiveExecutionException(
+            String.format("Error encountered while executing '%s' : Column '%s' has null value. " +
+                            "It should be of type 'String' or 'byte array'.", NAME, column)
```
albertshau: A String can be null, so the message can be a little confusing.

be of type -> be a non-null

wrangler-core/src/main/java/io/cdap/directives/row/RecordRegexFilter.java
```diff
@@ -107,10 +109,16 @@ public void destroy() {
           if (matchPattern((row.getValue(idx)).toString())) {
             continue;
           }
+        } else if (object == null) {
+          throw new DirectiveExecutionException(
+            String.format("Error encountered while executing '%s' : Column '%s' has null value. " +
+                            "It should be of type 'String', 'JSONObject' or 'Number'.", NAME, column)
```
albertshau: similar comment everywhere a null check is performed.

be of type -> be a non-null

wrangler-core/src/main/java/io/cdap/directives/transformation/Decode.java
```diff
@@ -88,8 +88,8 @@ public void initialize(Arguments args) throws DirectiveParseException {
     type = type.toUpperCase();
     if (!type.equals("BASE64") && !type.equals("BASE32") && !type.equals("HEX")) {
       throw new DirectiveParseException(
-        String.format("Type of decoding specified '%s' is not supported. Supports base64, base32 & hex.",
-                      type)
+        String.format("Error encountered while initializing '%s' : Type of encoding specified '%s' is not supported. " +
```
albertshau: Encoding type '%s' is not supported. Supported types are ...

wrangler-core/src/main/java/io/cdap/wrangler/expression/EL.java
```diff
@@ -67,11 +68,13 @@ public void compile(String expression) throws ELException {
...
-        throw new ELException(e.getCause().getMessage());
-      } else {
-        throw new ELException(e.getMessage());
-      }
+      // JexlException.getMessage() uses 'io.cdap.wrangler.expression.EL' class name in the error message.
+      // So instead use info object to get information about error message and create custom error message.
+      JexlInfo info = e.getInfo();
+      throw new ELException(
+        String.format("Error encountered while executing '%s' at line '%d' and column '%d'. " +
+                        "Make sure valid jexl transformation is provided.",
```
albertshau: Make sure a valid ...

wrangler-core/src/main/java/io/cdap/wrangler/expression/EL.java
```diff
@@ -94,14 +97,13 @@ public ELResult execute(ELContext context, boolean nullMissingFields) throws ELE
...
-        throw new ELException(e.getCause().getMessage());
-      } else {
-        throw new ELException(e.getMessage());
-      }
+      // JexlException.getMessage() uses 'io.cdap.wrangler.expression.EL' class name in the error message.
+      // So instead use info object to get information about error message and create custom error message.
+      JexlInfo info = e.getInfo();
+      throw new ELException(
+        String.format("Error encountered while executing '%s', at line '%d' and column '%d'. " +
+                        "Make sure valid jexl transformation is provided.",
```
albertshau: Make sure a valid ...

wrangler-core/src/main/java/io/cdap/directives/aggregates/IncrementTransientVariable.java
CuriousVini: Agreed. Unfortunately, this existing behavior is replicated at multiple places. Fixing it for the directives.

wrangler-core/src/main/java/io/cdap/directives/column/Copy.java
CuriousVini: yea, good catch!

wrangler-core/src/main/java/io/cdap/directives/transformation/Decode.java
CuriousVini: I think it should be `Decoding type...`

wrangler-core/src/main/java/io/cdap/directives/column/Swap.java
CuriousVini: yea for some reason intellij did not remove unused imports on the fly for this directive. Fixed.

wrangler-core/src/main/java/io/cdap/directives/column/SetType.java
```diff
@@ -258,9 +258,8 @@ private Object convertType(String toType, Object object) throws Exception {
 
       default:
         throw new DirectiveExecutionException(
-          String.format("Unknown data type '%s' found in the directive. " +
-                  "Accepted types are: int, short, long, double, boolean, string, bytes", toType)
-        );
+          NAME, String.format("Unknown data type '%s' found in the directive. " +
```
albertshau: Can we enhance this message to include the column name? Something like

Column '%s' is of unsupported type '%s'. Supported types are: ...

wrangler-core/src/main/java/io/cdap/directives/date/FormatDate.java
```diff
@@ -80,9 +80,7 @@ public void destroy() {
       int idx = dt.find(column);
 
       if (idx == -1) {
-        throw new DirectiveExecutionException(toString() + " : '" + column + "' column is not defined in the row. " +
-                                                "Please check the wrangling step."
-        );
+        throw new DirectiveExecutionException(NAME, "Column '" + column + "' does not exist.");
```
albertshau: nit: should use String.format() to be consistent

wrangler-core/src/main/java/io/cdap/directives/external/InvokeHttp.java
```diff
@@ -96,21 +96,18 @@ public void initialize(Arguments args) throws DirectiveParseException {
         String[] components = header.split("=");
         if (components.length != 2) {
           throw new DirectiveParseException (
-            String.format("Incorrect header '%s' specified. " +
-                            "Header should be specified as 'key=value' pairs separated by a comma (,).", header)
-          );
+            NAME, String.format("Incorrect header '%s' specified. Header should be specified as 'key=value' " +
```
albertshau, line 99: Can we enhance the message to also include the column name?

Same with the other messages in this class

wrangler-core/src/main/java/io/cdap/directives/parser/FixedLengthParser.java
```diff
@@ -94,16 +94,20 @@ public void destroy() {
...
           String data = (String) object;
           int length = data.length();
           // If the recordLength length doesn't match the string length.
           if (length < recordLength) {
             throw new ErrorRowException(
-              String.format("Fewer bytes than length of row specified - expected atleast %d bytes, found %s bytes.",
-                            recordLength, length),
-              2
-            );
+              NAME, String.format("Fewer bytes than length of row specified - expected atleast %d bytes, " +
```
albertshau: should include the column name here. Let's also fix this to have a space in between atleast and to accurately talk about number of characters and not bytes:

Column '%s' contains a value with fewer characters than the specified length of row. Expected at least %d characters but found %s characters.

albertshau (commented):
mostly comments about improving the existing error messages

wrangler-core/src/main/java/io/cdap/directives/external/InvokeHttp.java
CuriousVini: Did not add column name because this directive expects list of columns.

albertshau (approved):

PR #363, opened 2019-11-08
Workaround for wrangler

yaojiefeng:
This is a workaround for wrangler FLL. Currently applied to the following directives: Delete a Column, Set a Column, Send to Error, Mask, Filter, Fill-Null-Or-Empty
Change Data Type

### DIFF
--- a/wrangler-api/src/main/java/io/cdap/wrangler/api/Directive.java
+++ b/wrangler-api/src/main/java/io/cdap/wrangler/api/Directive.java
@@ -16,6 +16,7 @@
 
 package io.cdap.wrangler.api;
 
+import io.cdap.cdap.etl.api.lineage.field.FieldTransformOperation;
 import io.cdap.wrangler.api.parser.UsageDefinition;
 
 import java.util.List;
@@ -115,4 +116,13 @@ public interface Directive extends Executor<List<Row>, List<Row>> {
    * @see io.cdap.wrangler.api.parser.TokenType
    */
   UsageDefinition define();
+
+  /**
+   * Get the field operations related to get this directive.
+   *
+   * @return a list of field operations about this directive
+   */
+  default List<FieldTransformOperation> getFieldOperation() {
+    throw new UnsupportedOperationException("This directive does not support field level lineage.");
+  }
 }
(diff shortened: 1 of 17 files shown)


wrangler-transform/src/main/java/io/cdap/wrangler/Wrangler.java
```diff
@@ -511,14 +471,89 @@ private Object getValue(StructuredRecord input, String fieldName) {
...
+    return true;
+  }
+
+  /**
+   * Validates input schema.
+   *
+   * @param inputSchema configured for the plugin
+   * @param collector failure collector
+   */
+  private void validateInputSchema(Schema inputSchema, FailureCollector collector) {
```
chtyim: `InputSchema` should be `@Nullable`

chtyim (approved):
Just one minor comment

wrangler-api/src/main/java/io/cdap/wrangler/api/Directive.java
```diff
@@ -115,4 +116,13 @@
    * @see io.cdap.wrangler.api.parser.TokenType
    */
   UsageDefinition define();
+
```
nitinmotgi, line 119: Should not be added to this interface. Could we add another interface named Mutation that has this method. Only directives that implement that interface will that. This is very clunky.

wrangler-transform/src/main/java/io/cdap/wrangler/Wrangler.java
```diff
@@ -511,14 +471,89 @@ private Object getValue(StructuredRecord input, String fieldName) {
...
    * Retrieves the base url from the context and appends method to value to the final url.
    *
    * @param method to be invoked.
    * @return fully formed url to the method.
    */
-  private URL getDPServiceURL(String method) throws URISyntaxException, MalformedURLException {
-    URL url = getContext().getServiceURL(Contexts.SYSTEM, APPLICATION_NAME, SERVICE_NAME);
+  private URL getDPServiceURL(ServiceDiscoverer discoverer,
+                              String method) throws URISyntaxException, MalformedURLException {
+    URL url = discoverer.getServiceURL(Contexts.SYSTEM, APPLICATION_NAME, SERVICE_NAME);
```
nitinmotgi, line 556: what if discoverer is null ?

yaojiefeng: The discoverer here is the StageContext, it cannot be null

wrangler-api/src/main/java/io/cdap/wrangler/api/Directive.java
yaojiefeng: Addressed in https://github.com/data-integrations/wrangler/pull/365

PR #365, opened 2019-11-08
fix fll for directives that can g…

yaojiefeng:
…enerate field

Addressed comments on https://github.com/data-integrations/wrangler/pull/363#discussion_r343982522

And fix the issue for directives that can generate columns not in the input fields

### DIFF
--- a/wrangler-api/src/main/java/io/cdap/wrangler/api/Directive.java
+++ b/wrangler-api/src/main/java/io/cdap/wrangler/api/Directive.java
@@ -120,6 +120,8 @@ public interface Directive extends Executor<List<Row>, List<Row>> {
 
   /**
    * Get the field operations related to get this directive.
+   * TODO: CDAP-16101 correctly emit the field lineage for wrangler, this method might be deprecated when we have
+   * the correct fix.
    *
    * @return a list of field operations about this directive
    */
--- a/wrangler-api/src/main/java/io/cdap/wrangler/api/Directive.java
+++ b/wrangler-api/src/main/java/io/cdap/wrangler/api/Directive.java
@@ -16,6 +16,7 @@
 
 package io.cdap.wrangler.api;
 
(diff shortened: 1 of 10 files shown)


wrangler-core/src/main/java/io/cdap/directives/transformation/ColumnExpression.java
```diff
@@ -76,11 +79,23 @@ public UsageDefinition define() {
...
   @Override
-  public List<FieldTransformOperation> getFieldOperation() {
-    return Collections.singletonList(
-      new FieldTransformOperation(String.format("Set column %s", column),
-                                  String.format("Set the column %s with expression %s", column, expression),
-                                  Collections.emptyList(), column));
+  public List<FieldTransformOperation> getFieldOperations(StageContext context) {
+    Schema schema = context.getInputSchema();
+    if (schema != null && schema.getFields() != null) {
+      List<String> inputFields = schema.getFields().stream().map(Schema.Field::getName).collect(Collectors.toList());
```
CuriousVini: why do we need list of inputFields? why not rely on schema.getField() method?

wrangler-api/src/main/java/io/cdap/wrangler/api/Directive.java
```diff
@@ -122,7 +123,7 @@
    *
    * @return a list of field operations about this directive
    */
-  default List<FieldTransformOperation> getFieldOperation() {
+  default List<FieldTransformOperation> getFieldOperations(StageContext context) {
```
CuriousVini, line 128: we should add this usage to the javadoc of the Directive interface

yaojiefeng: I added a TODO for this method

CuriousVini (approved):

PR #371, opened 2019-12-14
Lineage version V2 with right APIs

nitinmotgi:


nitinmotgi:
Will squash and merge once approved.

wrangler-api/src/main/java/io/cdap/wrangler/api/lineage/Lineage.java
```diff
@@ -0,0 +1,87 @@
...
+package io.cdap.wrangler.api.lineage;
+
+import io.cdap.cdap.api.annotation.Beta;
+
+/**
+ * This class <code>Lineage</code> defines a mutation defined by the directive that gets capture as lineage.
+ *
+ * <p>Directives have to implement this class to inject their mutations for lineage to be constructed.</p>
+ * <p>
+ *   The method <code>lineage</code> is invoked separately in the <code>preRun</code> phase of the pipeline
```
yaojiefeng: nit: prepareRun

wrangler-api/src/main/java/io/cdap/wrangler/api/lineage/Mutation.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright © 2017-2019 Cask Data, Inc.
+ * Copyright © 2019 Cask Data, Inc.
```
yaojiefeng: This should still be 2017-2019 since the file is not added in this pr

wrangler-api/src/main/java/io/cdap/wrangler/api/lineage/Relation.java
```diff
@@ -0,0 +1,123 @@
...
+package io.cdap.wrangler.api.lineage;
+
+import io.cdap.cdap.api.annotation.Beta;
+
+import java.io.Serializable;
+import java.util.List;
+
+/**
+ * The <code>Relation</code> class represents a relation between a list of sources to a list of targets.
+ * The relation can have 4 different types and can be many-to-many, many-to-one, or one-to-many.
```
yaojiefeng: Missing one-to-one here.

wrangler-core/src/main/java/io/cdap/directives/column/Rename.java
```diff
@@ -94,4 +85,12 @@ public void destroy() {
     }
     return rows;
   }
+
+  @Override
+  public Mutation lineage() {
+    return Mutation.builder()
+      .readable("Renamed column '%s' to '%s'", source.value(), target.value())
+      .relation(source, Many.columns(source, target))
```
yaojiefeng: this should be a one-to-one mapping since there is no source col after the rename

wrangler-core/src/main/java/io/cdap/directives/column/SplitToColumns.java
```diff
@@ -99,5 +102,26 @@ public void destroy() {
...
+  @Override
+  public Mutation lineage() {
+    return Mutation.builder()
+      .readable("Split the column '%s' with regex '%s'", column, regex)
+      .relation(
+        column,
+        Many.columns(
+          column,
+          String.format("%s_%d", column, 1),
+          String.format("%s_%d", column, 2),
```
yaojiefeng, line 115: We are hard coding the columns here, should we use ALL relation in this case?

wrangler-core/src/main/java/io/cdap/wrangler/lineage/LineageOperations.java
```diff
@@ -0,0 +1,155 @@
...
+   * Generates the list of {@link FieldOperation} required for generating lineage.
+   *
+   * @return a {@link List} of {@link FieldOperation}
+   *
+   * @throws Exception thrown if there is any issue generating field operations.
+   */
+  public List<FieldOperation> generate() throws Exception {
+    // After input and output schema are validated, it's time to extract
+    // all the fields from input and output schema.
+    Set<String> input = inputSchema.getFields().stream()
```
yaojiefeng: We can change the constructor to accept Set<String> as the input and output fields instead of Schema, since in `Wrangler` we validate the input and output schema are not null, so it is better to have this logic there. Doing it here will have IDE warning about NPE

yaojiefeng (commented):
Some minor comments

wrangler-api/src/main/java/io/cdap/wrangler/api/lineage/Lineage.java
nitinmotgi: fixed

wrangler-api/src/main/java/io/cdap/wrangler/api/lineage/Mutation.java
nitinmotgi: fixed

wrangler-api/src/main/java/io/cdap/wrangler/api/lineage/Relation.java
nitinmotgi: fixed

wrangler-core/src/main/java/io/cdap/directives/column/Rename.java
nitinmotgi: fixed

wrangler-core/src/main/java/io/cdap/directives/column/SplitToColumns.java
nitinmotgi: No, I would rather use fixed set. If there is nothing beyond this for a column, it would be ignored anyway.

wrangler-core/src/main/java/io/cdap/wrangler/lineage/LineageOperations.java
nitinmotgi: fixed!

nitinmotgi:
@yaojiefeng addressed all your comments.

yaojiefeng (approved):

wrangler-api/src/main/java/io/cdap/wrangler/api/lineage/Many.java
```diff
@@ -0,0 +1,147 @@
...
+ *    .readable("string")
+ *    .relation("source", Many.columns("col3", "col4", "col5").build();
+ * </code>
+ *
+ * @see Lineage
+ * @see Mutation
+ * @see Relation
+ */
+@Beta
+public final class Many implements Serializable {
```
CuriousVini, line 52: should have serialVersionUID

wrangler-api/src/main/java/io/cdap/wrangler/api/lineage/Mutation.java
```diff
@@ -16,60 +16,242 @@
...
+ * @see Many
  */
-public final class Mutation {
-  // Name of the column on which the mutation is applied.
-  private final String column;
-
-  // Specifies the type of the mutation being applied on the column.
-  private final MutationType type;
+@Beta
+public final class Mutation implements Serializable {
```
CuriousVini, line 43: should have serialVersionUID

wrangler-api/src/main/java/io/cdap/wrangler/api/lineage/Relation.java
```diff
@@ -0,0 +1,123 @@
...
+ *     one-to-many.
+ *   </li>
+ * </ul>
+ *
+ * @see Lineage
+ * @see Many
+ * @see Mutation
+ */
+@Beta
+public final class Relation implements Serializable {
```
CuriousVini, line 53: same here

wrangler-transform/src/main/java/io/cdap/wrangler/Wrangler.java
```diff
@@ -517,31 +506,13 @@ private RecipeParser initializeAndGetParser(
...
 
-    // Make a call to retrieve the directive config and create a context.
-    // In this plugin, a call is made to the service at runtime.
-    // NOTE: This has to be moved to pre-start-plugin phase so that the information
-    // it's not heavily parallelized to crash the service.
-    try {
-      URL url = getDPServiceURL(context, CONFIG_METHOD);
-      if (url != null) {
-        ConfigDirectiveContext dContext = new ConfigDirectiveContext(url);
-        recipe.initialize(dContext);
```
CuriousVini, line 531: why is this change needed?

wrangler-api/src/main/java/io/cdap/wrangler/api/lineage/Lineage.java
```diff
@@ -0,0 +1,87 @@
...
+ *   are available to be used within the <code>lineage</code> method.
+ * </p>
+ *
+ * <p>
+ *   {@link Mutation} captures all the changes the directive is going to be applying of the data. It has
+ *   two major methods:
+ *
+ *   <ul>
+ *     <il>
+ *       <code>readable</code> - This method is defined to provide the post readable description of the mutation
```
sreevatsanraman: replace post readable with post transformation

wrangler-api/src/main/java/io/cdap/wrangler/api/lineage/Lineage.java
```diff
@@ -0,0 +1,87 @@
...
+ *     the directive is applying on data. Care should be taken to use the right tense as the lineage would be consumed
+ *     by users after the transformation has been applied on the data. As best practise, it's highly recommended to use
+ *     past-tense for describing the transformations. Additionally, the language of the description is not trying
+ *     to provide complete details and configuration, but actually focusing on operations that directive has
+ *     performed on data.
```
sreevatsanraman: Can you add an example?

wrangler-api/src/main/java/io/cdap/wrangler/api/lineage/Many.java
```diff
@@ -0,0 +1,147 @@
...
+    return new Many(columns);
+  }
+
+  /**
+   * List of {@link ColumnName} specified as source or target.
+   *
+   * @param columns list source or target columns of type {@link List} of {@link ColumnName}.
+   * @return a instance of {@link Many}.
+   */
+  public static Many columns(List<ColumnName> columns) {
```
sreevatsanraman: It is unclear when to use columns(List<ColumnName>) vs columns(List<String>). Can you provide guidance in the javadoc

nitinmotgi: One is List with ColumnName and other with String.

wrangler-transform/src/main/java/io/cdap/wrangler/Wrangler.java
```diff
@@ -517,31 +506,13 @@ private RecipeParser initializeAndGetParser(
...
+      new MigrateToV2(directives).migrate(),
+      registry
+    );
 
-    // Make a call to retrieve the directive config and create a context.
-    // In this plugin, a call is made to the service at runtime.
-    // NOTE: This has to be moved to pre-start-plugin phase so that the information
-    // it's not heavily parallelized to crash the service.
-    try {
-      URL url = getDPServiceURL(context, CONFIG_METHOD);
```
CuriousVini, line 528: also `getDPServiceURL ` is an unused method. It should be removed

wrangler-core/src/main/java/io/cdap/wrangler/lineage/LineageOperations.java
```diff
@@ -0,0 +1,141 @@
...
+            definedSources.addAll(relation.getSources());
+          }
+        );
+      }
+    }
+
+    // We iterate through all the input fields in the schema, check if there is corresponding
+    // field in the output schema. If both exists, then a identity mapping transform is added
+    // to the {@code FieldTransformationOperation} is added.
+    input.removeAll(definedSources);
```
CuriousVini: this will modify the input hashset that was passed to this class. we should make a copy of it in the constructor.

wrangler-api/src/main/java/io/cdap/wrangler/api/lineage/Lineage.java
```diff
@@ -0,0 +1,87 @@
...
+ * @see Many
+ */
+@Beta
+public interface Lineage {
+  /**
+   * Returns a Mutation that can be used to generate lineage.
+   *
+   * @return a instance of {@link Mutation} object used to generate lineage.
+   */
+  Mutation lineage();
```
CuriousVini, line 86: Would it be a good idea to call it `getMutation()`?

wrangler-api/src/main/java/io/cdap/wrangler/api/lineage/Many.java
```diff
@@ -0,0 +1,147 @@
...
+    return new Many(columns);
+  }
+
+  /**
+   * Set of columns as source or target.
+   *
+   * @param columns list source or target columns of type {@linkn Set} of {@link String}.
+   * @return a instance of {@link Many}.
+   */
+  public static Many of(Set<String> columns) {
```
CuriousVini: we should use `Collection` instead

wrangler-api/src/main/java/io/cdap/wrangler/api/lineage/Many.java
```diff
@@ -0,0 +1,147 @@
...
+    return new Many(columns.stream().map(ColumnName::value).collect(Collectors.toList()));
+  }
+
+  /**
+   * Set of {@link ColumnName} specified as source or target.
+   *
+   * @param columns list source or target columns of type {@link Set} of {@link ColumnName}.
+   * @return a instance of {@link Many}.
+   */
+  public static Many columns(Set<ColumnName> columns) {
```
CuriousVini: same here. We should use `Collection`

wrangler-api/src/main/java/io/cdap/wrangler/api/lineage/Many.java
```diff
@@ -0,0 +1,147 @@
...
+ * @see Relation
+ */
+@Beta
+public final class Many implements Serializable {
+  private List<String> columns = new ArrayList<>();
+
+  /**
+   * @return a instance of <pre>List<String></pre> that returns columns associated with source or target.
+   */
+  public List<String> columns() {
```
CuriousVini, line 59: this method returns list of strings where as other methods return `Many`. Since this is a getter, how about calling this `getColumns` or `getColumnNames`.

wrangler-api/src/main/java/io/cdap/wrangler/api/lineage/Many.java
```diff
@@ -0,0 +1,147 @@
...
+   * @return a instance of {@link Many}.
+   */
+  public static Many of(List<String> columns) {
+    return new Many(columns);
+  }
+
+  /**
+   * Set of columns as source or target.
+   *
+   * @param columns list source or target columns of type {@linkn Set} of {@link String}.
```
CuriousVini: typo: @link

wrangler-api/src/main/java/io/cdap/wrangler/api/lineage/Mutation.java
```diff
@@ -16,60 +16,242 @@
...
+    this.relations = new ArrayList<>();
+  }
 
-  public Mutation(String column, MutationType type, String description) {
-    this.column = column;
-    this.type = type;
-    this.description = description;
+  private Mutation(String readable, List<Relation> relations) {
+    this.readable = readable;
+    this.relations = relations;
```
CuriousVini: should do a deep copy of relations since the class is immutable.

wrangler-api/src/main/java/io/cdap/wrangler/api/lineage/Mutation.java
```diff
@@ -16,60 +16,242 @@
...
   }
 
   /**
-   * @return the type of mutation being applied on the <code>column</code>.
+   * @return a {@link List} of {@link Relation} associated with the transformation.
    */
-  public MutationType type() {
-    return type;
+  public List<Relation> relations() {
+    return relations;
```
CuriousVini, line 68: should return a deep copy of list of relations.

wrangler-api/src/main/java/io/cdap/wrangler/api/lineage/Relation.java
```diff
@@ -0,0 +1,123 @@
...
+  /**
+   * Specifies the type of relations that can exist between source and target.
+   */
+  public enum Type {
+    DROP,
+    ALL,
+    CREATE,
+    STANDARD,
```
CuriousVini, line 63: why do we have `ALL` and `STANDARD` types?

wrangler-api/src/main/java/io/cdap/wrangler/api/lineage/Mutation.java
```diff
@@ -16,60 +16,242 @@
...
+      return this;
+    }
+
+    /**
+     * Specifies a relation that has no cause but effect.
+     *
+     * @param targets a list of targets.
+     * @return a instance of {@link Mutation.Builder}.
+     */
+    public Mutation.Builder generate(Many targets) {
```
CuriousVini: Is there any specific reason to use method name `generate`(equivalent of CREATE) and `relation` (equivalent of STANDARD). Would it be a good idea to use create and standard method names?

wrangler-core/src/main/java/io/cdap/wrangler/lineage/LineageOperations.java
yaojiefeng: Changed the implementation to not modify the hashset and make all the fields in the constructor immutable

wrangler-api/src/main/java/io/cdap/wrangler/api/lineage/Lineage.java
yaojiefeng: Here since the class is named `Lineage`, I think it is good to call this `lineage`. Also, in wrangler, we don't have `get` as prefix for the method name. So Nitin made all the names in this pr not prefixed with `get`

wrangler-api/src/main/java/io/cdap/wrangler/api/lineage/Many.java
yaojiefeng: In wrangler, we don't have `get` as prefix for the method name. So Nitin made all the names in this pr not prefixed with `get`

wrangler-api/src/main/java/io/cdap/wrangler/api/lineage/Many.java
yaojiefeng: I removed this method with the change for the parameter to Collection

wrangler-api/src/main/java/io/cdap/wrangler/api/lineage/Relation.java
yaojiefeng: STANDARD means the relation has to clear-defined, the source and target has to be specified. ALL means the directive will apply changes to all the output fields, which we don't know in the directive, so we will have special logic when we generate lineage for ALL operations

wrangler-api/src/main/java/io/cdap/wrangler/api/lineage/Mutation.java
yaojiefeng: I think `relation` is fine for STANDARD, since most of the directives will provide clear relation, will change `generate` to `create`

CuriousVini (approved):

PR #372, opened 2019-12-16
Data Model Schema Viewer

bebinu:
* data model viewer schema definition
* data model schema loader
* data model api handler


nitinmotgi:
@bebinu Could we please not have 'utils' package. Over a period of time this gets misused heavily. Could we name the package as what it's supposed to be.

nitinmotgi:
Take back my comment :)

nitinmotgi:
Also, I had built a schema registry which supported a lot of functionality. We can bring that back.

bebinu:
> Also, I had built a schema registry which supported a lot of functionality. We can bring that back.
@nitinmotgi once we design and implement the semantic business glossary, then the SchemaRegistry will be enhanced to read data from that system. In addition, in this incremental update, the data model schema definitions are static so preferring to implement a dedicated handler to isolate scope.

PR #388, opened 2020-03-25
Bump version to 4.1.5

ajainarayanan:
This does not bump CDAP version and plugin parent versions yet. Those will be bumped once CDAP artifacts are pushed to maven central.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -19,7 +19,7 @@
 
   <groupId>io.cdap.wrangler</groupId>
   <artifactId>wrangler</artifactId>
-  <version>4.1.4</version>
+  <version>4.1.5</version>
   <name>Wrangler</name>
   <packaging>pom</packaging>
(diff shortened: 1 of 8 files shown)


albertshau (dismissed):

pom.xml
```diff
@@ -83,7 +83,7 @@
     <aws.sdk.version>1.11.133</aws.sdk.version>
     <bigquery.connector.hadoop2.version>0.10.2-hadoop2</bigquery.connector.hadoop2.version>
     <bouncycastle.version>1.56</bouncycastle.version>
-    <cdap.version>6.1.1</cdap.version>
+    <cdap.version>6.1.2</cdap.version>
```
albertshau: this should not be updated if it doesn't need to be

albertshau (approved):

PR #392, opened 2020-04-09
CDAP-16465 list tables correctly

albertshau:
Fixed the database handler to pass the catalog correctly so that
only tables within the catalog are returned.

PR #395, opened 2020-04-14
CDAP-16465 list tables correctly

albertshau:
Fixed the database handler to pass the catalog correctly so that
only tables within the catalog are returned.


albertshau:
cherry-pick of #392

CuriousVini (approved):

PR #397, opened 2020-04-21
[CDAP-16633] Added option to generate scoped GoogleCredentials with BQ and Drive scope for all BQ requests

vravish:


albertshau (approved):
lgtm. please be sure to cherrypick to release branches as well

PR #405, opened 2020-05-02
Minor fixes in testing and handling crlf correctly

nitinmotgi:


wrangler-core/src/main/java/io/cdap/wrangler/parser/MigrateToV2.java
```diff
@@ -56,7 +56,7 @@ public MigrateToV2(@Nullable String recipe, String delimiter) {
   }
 
   public MigrateToV2(String recipe) {
-    this(recipe, "\n");
+    this(recipe, "\\r?\\n");
```
albertshau, line 59: I've seen recipes with `\r\n` in the past and they seemed to work. Were only some directives stripping the `\r` or something?

wrangler-core/src/test/java/io/cdap/wrangler/parser/RecipeCompilerTest.java
```diff
@@ -183,26 +182,31 @@ public void test() throws Exception {
...
-    Assert.assertTrue(true);
+    TestingRig.compileSuccess(recipe);
+  }
+
+  @Test
+  public void testHandlingCRLF() throws Exception {
+    String[] recipe = new String[] {
+      "parse-as-csv :body '\t' true;\r\ndrop :body;\r\n"
+    };
+    TestingRig.compileSuccess(recipe);
```
albertshau, line 201: shouldn't just test that it compiles (didn't it compile successfully before?) Should test that the directives are parsed into the exact strings that are expected.

albertshau (commented):
Can you open a jira for this so that we have a record of the fix?

PR #406, opened 2020-05-02
[CDAP-16739] Added hard limit for Avro data file, proper error messag…

nitinmotgi:
…e redirecting user to use file source connector with format avro. Fixed parse-as-json to highlight error as well as handling integers correctly now

- [x] test case added
- [x] test file added
- [x] error message reviewed.


wrangler-core/src/main/java/io/cdap/directives/parser/JsParser.java
```diff
@@ -143,8 +167,19 @@ public void destroy() {
               row.add(column, getValue(element.getAsJsonPrimitive()));
             }
           }
-        } catch (JSONException e) {
-          throw new ErrorRowException(NAME, e.getMessage(), 1);
+        } catch (Exception e) {
+          String msg = e.getCause() != null ? e.getCause().getMessage() : e.getMessage();
+          // In case there are more rows being handled, we attempt to surface the Json that is
+          // causing an issue to make it easier for users to identify the problem quickly.
+          if (rows.size() > 1) {
```
albertshau, line 174: isn't rows always size 1? Otherwise throwing an ErrorRowException would incorrectly mark every Row as an error?

wrangler-core/src/main/java/io/cdap/directives/parser/ParseAvroFile.java
```diff
@@ -105,7 +115,10 @@ public void destroy() {
           }
         } else {
           throw new DirectiveExecutionException(
-            NAME, String.format("Column '%s' is of invalid type. It should be of type 'byte array'.", column));
+            NAME, String.format("Avro data file parsing directive requires '%s' to be of byte. " +
```
albertshau: to be of byte -> to be a bytes field.

wrangler-core/src/main/java/io/cdap/directives/parser/ParseAvroFile.java
```diff
@@ -105,7 +115,10 @@ public void destroy() {
...
+                                  "Change type in input as byte and make sure the format in File source is " +
```
albertshau: as byte -> to bytes

wrangler-core/src/main/java/io/cdap/directives/parser/ParseAvroFile.java
```diff
@@ -105,7 +115,10 @@ public void destroy() {
...
+                                  "set as 'blob'. We highly recommend to use File source connector to read Avro files.",
```
albertshau: We recommend using the avro format in the file source instead of using this directive.

wrangler-core/src/test/java/io/cdap/directives/parser/ParseAvroFileTest.java
```diff
@@ -52,6 +52,40 @@ public void testParseAsAvroFile() throws Exception {
     Assert.assertEquals(1495194308245L, results.get(1688).getValue("timestamp"));
   }
 
+  @Test
+  public void testAvroWithJsonFieldParsingWithParseJsonIgnoreError() throws Exception {
```
albertshau, line 56: these new tests are testing two directives. More specifically, these are not testing any change in behavior in the avro directive, they are just testing the changes in json directive.

Unit tests should be testing isolated components when possible. It would be better to enhance the parse json test. For example, if we remove the parse avro file directive, we could easily remove test coverage of the parse json directive accidentally.

wrangler-core/src/test/java/io/cdap/directives/parser/ParseAvroFileTest.java
```diff
@@ -52,6 +52,40 @@ public void testParseAsAvroFile() throws Exception {
...
+    String[] directives = new String[] {
+      "parse-as-avro-file body",
+      "parse-as-json :message_contents 2 true"
+    };
+
+    List<Row> rows = new ArrayList<>();
+    rows.add(new Row("body", data));
+
+    List<Row> results = TestingRig.execute(directives, rows);
+    Assert.assertEquals(6693, results.size()); // total 6670, 7 records are bad.
```
albertshau, line 69: these types of asserts are not specific enough. For example, how can we be sure the right rows are filtered? It would be better to just send a single row in as input.

wrangler-core/src/main/java/io/cdap/directives/parser/JsParser.java
nitinmotgi: Not always. If we parsing AvroFile. Parse-as-avro-file will generate records > 1.

wrangler-core/src/test/java/io/cdap/directives/parser/ParseAvroFileTest.java
nitinmotgi: This was a problem customer had. We have other tests for catching parse-as-json problems. This was when a json in avro file have. Using redacted user data as test.

wrangler-core/src/test/java/io/cdap/directives/parser/ParseAvroFileTest.java
nitinmotgi: Using customer test data to make sure we don't break it.

nitinmotgi:
@albertshau PTAL

wrangler-core/src/main/java/io/cdap/directives/parser/JsParser.java
albertshau: yes, but the RecipeExecutor will break that into multiple lists of size 1. Either this is always size 1, or the entire ErrorRowException contract is wrong.

wrangler-core/src/test/java/io/cdap/directives/parser/ParseAvroFileTest.java
albertshau: this doesn't make sure of anything besides the size

nitinmotgi:
@albertshau PTAL

wrangler-core/src/test/java/io/cdap/directives/parser/ParseAvroFileTest.java
albertshau: and this is a unit test, meaning it should be testing very specific parts of the code. With the ignore-error changed introduced, it should be testing that the json directive returns empty results when given bad data and ignore-error is true.

PR #407, opened 2020-05-03
[CDAP-16740] Centralizing logic for EL Context and improve tests

nitinmotgi:
It looks like a lot of files, but those are from refactoring of changing Row.length() to Row.width(). Internal function.

### DIFF
--- a/wrangler-api/src/main/java/io/cdap/wrangler/api/ErrorRowException.java
+++ b/wrangler-api/src/main/java/io/cdap/wrangler/api/ErrorRowException.java
@@ -16,9 +16,6 @@
 
 package io.cdap.wrangler.api;
 
-import java.util.Collections;
-import java.util.List;
-
 /**
  * Exception throw when the record needs to emitted to error collector.
  */
@@ -42,7 +39,7 @@ public class ErrorRowException extends Exception {
   }
 
   public ErrorRowException(String directiveName, String errorMessage, int code) {
-    this(String.format("Error encountered while executing '%s' : %s", directiveName, errorMessage), code);
+    this(String.format("%s (ecode: %d, directive: %s)", errorMessage, code, directiveName), code);
   }
 
   /**
--- a/wrangler-api/src/main/java/io/cdap/wrangler/api/Row.java
+++ b/wrangler-api/src/main/java/io/cdap/wrangler/api/Row.java
@@ -172,6 +172,14 @@ public final class Row implements Serializable {
   /**
    * @return  width of the row.
    */
+  @Deprecated
+  public int length() {
+    return columns.size();
+  }
+
+  /**
+   * @return  width of the row.
+   */
   public int width() {
     return columns.size();
   }
--- a/wrangler-api/src/main/java/io/cdap/wrangler/api/Row.java
+++ b/wrangler-api/src/main/java/io/cdap/wrangler/api/Row.java
@@ -170,9 +170,9 @@ public final class Row implements Serializable {
   }
 
   /**
-   * @return  Length of the row.
+   * @return  width of the row.
    */
-  public int length() {
+  public int width() {
     return columns.size();
   }
 
--- a/wrangler-core/src/main/java/io/cdap/directives/aggregates/SetTransientVariable.java
+++ b/wrangler-core/src/main/java/io/cdap/directives/aggregates/SetTransientVariable.java
@@ -82,18 +82,7 @@ public class SetTransientVariable implements Directive {
   public List<Row> execute(List<Row> rows, ExecutorContext context) throws DirectiveExecutionException {
     for (Row row : rows) {
       // Move the fields from the row into the context.
-      ELContext ctx = new ELContext(context);
-      ctx.set("this", row);
-      for (String var : el.variables()) {
-        ctx.set(var, row.getValue(var));
-      }
-
-      // Transient variables are added.
-      if (context != null) {
-        for (String variable : context.getTransientStore().getVariables()) {
-          ctx.set(variable, context.getTransientStore().get(variable));
-        }
-      }
+      ELContext ctx = new ELContext(context, el, row);
 
       // Execution of the script / expression based on the row data
       // mapped into context.
--- a/wrangler-core/src/main/java/io/cdap/directives/row/RecordConditionFilter.java
+++ b/wrangler-core/src/main/java/io/cdap/directives/row/RecordConditionFilter.java
@@ -91,19 +91,7 @@ public class RecordConditionFilter implements Directive, Lineage {
     List<Row> results = new ArrayList<>();
     for (Row row : rows) {
       // Move the fields from the row into the context.
-      ELContext ctx = new ELContext(context);
-      for (String var : el.variables()) {
-        Object value = row.getValue(var);
-        String strValue;
-        // support numeric values by converting them to string
-        if (value instanceof Number) {
-          Number number = (Number) value;
-          strValue = number.toString();
-          ctx.set(var, strValue);
-        } else {
-          ctx.set(var, row.getValue(var));
-        }
-      }
+      ELContext ctx = new ELContext(context, el, row);
       if (context != null) {
         for (String variable : context.getTransientStore().getVariables()) {
           ctx.set(variable, context.getTransientStore().get(variable));
--- a/wrangler-core/src/main/java/io/cdap/wrangler/executor/RecipePipelineExecutor.java
+++ b/wrangler-core/src/main/java/io/cdap/wrangler/executor/RecipePipelineExecutor.java
@@ -122,28 +122,31 @@ public final class RecipePipelineExecutor implements RecipePipeline<Row, Structu
         if (context != null) {
           context.getTransientStore().reset(TransientVariableScope.LOCAL);
         }
-        List<Row> newRows = rows.subList(i, i + 1);
+
+        List<Row> cumulativeRows = rows.subList(i, i + 1);
         try {
           for (Executor<List<Row>, List<Row>> directive : directives) {
             try {
-              newRows = directive.execute(newRows, context);
-              if (newRows.size() < 1) {
+              cumulativeRows = directive.execute(cumulativeRows, context);
+              if (cumulativeRows.size() < 1) {
                 break;
               }
             } catch (ReportErrorAndProceed e) {
-              messages.add(String.format("%d:%s", e.getCode(), e.getMessage()));
+              messages.add(String.format("%s (ecode: %d)", e.getMessage(), e.getCode()));
+              collector
+                .add(new ErrorRecord(rows.subList(i, i + 1).get(0), String.join(",", messages), e.getCode(), true));
+              cumulativeRows = new ArrayList<>();
+              break;
             }
           }
-          if (newRows.size() > 0) {
-            results.addAll(newRows);
-          }
+          results.addAll(cumulativeRows);
         } catch (ErrorRowException e) {
           messages.add(String.format("%s", e.getMessage()));
           collector
-            .add(new ErrorRecord(newRows.get(0), String.join(",", messages), e.getCode(),
+            .add(new ErrorRecord(rows.subList(i, i + 1).get(0), String.join(",", messages), e.getCode(),
               e.isShownInWrangler()));
         }
-        i++;
+        ++i;
       }
     } catch (DirectiveExecutionException e) {
       throw new RecipeException(e.getMessage(), e);
--- a/wrangler-core/src/main/java/io/cdap/wrangler/expression/EL.java
+++ b/wrangler-core/src/main/java/io/cdap/wrangler/expression/EL.java
@@ -74,7 +74,7 @@ public final class EL {
       throw new ELException(
         String.format("Error encountered while executing '%s' at line '%d' and column '%d'. " +
                         "Make sure a valid jexl transformation is provided.",
-                      info.getDetail().toString(), info.getLine(), info.getColumn()), e);
+                      info.toString(), info.getLine(), info.getColumn()), e);
     } catch (Exception e) {
       throw new ELException(e);
     }
--- a/wrangler-core/src/main/java/io/cdap/wrangler/expression/EL.java
+++ b/wrangler-core/src/main/java/io/cdap/wrangler/expression/EL.java
@@ -74,7 +74,7 @@ public final class EL {
       throw new ELException(
         String.format("Error encountered while executing '%s' at line '%d' and column '%d'. " +
                         "Make sure a valid jexl transformation is provided.",
-                      info.toString(), info.getLine(), info.getColumn()), e);
+                      info.getDetail().toString(), info.getLine(), info.getColumn()), e);
     } catch (Exception e) {
       throw new ELException(e);
     }
--- a/wrangler-core/src/main/java/io/cdap/wrangler/expression/ELContext.java
+++ b/wrangler-core/src/main/java/io/cdap/wrangler/expression/ELContext.java
@@ -43,11 +43,12 @@ public class ELContext implements JexlContext {
    * All properties of this class are public to ensure they can be accessed with dot(.) format.
    */
   public static class Context {
-    public String environment;
-    public String name;
-    public long nano;
-    public long millis;
-    public Context (String environment, String name) {
+    private final String environment;
+    private final String name;
+    private final long nano;
+    private final long millis;
+
+    public Context(String environment, String name) {
       this.environment = environment;
       this.name = name;
       this.nano = System.nanoTime();
--- a/wrangler-core/src/main/java/io/cdap/wrangler/expression/ELContext.java
+++ b/wrangler-core/src/main/java/io/cdap/wrangler/expression/ELContext.java
@@ -22,6 +22,7 @@ import org.apache.commons.jexl3.JexlContext;
 
 import java.util.HashMap;
 import java.util.Map;
+import javax.annotation.Nullable;
 
 /**
  * Manages variables which can be referenced in a JEXL expression.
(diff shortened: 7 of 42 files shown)


wrangler-api/src/main/java/io/cdap/wrangler/api/Row.java
```diff
@@ -170,9 +170,9 @@ public int find(String col) {
   }
 
   /**
-   * @return  Length of the row.
+   * @return  width of the row.
    */
-  public int length() {
+  public int width() {
```
CuriousVini, line 183: This is a change in api. Is this change backwards compatible? would this require redeployment of existing pipelines?

wrangler-api/src/main/java/io/cdap/wrangler/api/ErrorRowException.java
```diff
@@ -42,7 +39,7 @@ public ErrorRowException(String message, int code) {
   }
 
   public ErrorRowException(String directiveName, String errorMessage, int code) {
-    this(String.format("Error encountered while executing '%s' : %s", directiveName, errorMessage), code);
+    this(String.format("%s (ecode: %d, directive: %s)", errorMessage, code, directiveName), code);
```
CuriousVini, line 42: how about if we first have directive name, error message and then error code?

wrangler-core/src/main/java/io/cdap/wrangler/expression/EL.java
```diff
@@ -74,7 +74,7 @@ public void compile(String expression) throws ELException {
       throw new ELException(
         String.format("Error encountered while executing '%s' at line '%d' and column '%d'. " +
                         "Make sure a valid jexl transformation is provided.",
-                      info.getDetail().toString(), info.getLine(), info.getColumn()), e);
+                      info.toString(), info.getLine(), info.getColumn()), e);
```
CuriousVini: As per the javadoc, getDetail() provides actual part of the code that triggered the error. why do we need this change?

wrangler-core/src/main/java/io/cdap/wrangler/executor/RecipePipelineExecutor.java
```diff
@@ -122,28 +122,31 @@ public void destroy() {
...
+          results.addAll(cumulativeRows);
         } catch (ErrorRowException e) {
           messages.add(String.format("%s", e.getMessage()));
           collector
-            .add(new ErrorRecord(newRows.get(0), String.join(",", messages), e.getCode(),
+            .add(new ErrorRecord(rows.subList(i, i + 1).get(0), String.join(",", messages), e.getCode(),
               e.isShownInWrangler()));
         }
-        i++;
+        ++i;
```
CuriousVini, line 149: why is this change needed?

wrangler-core/src/test/java/io/cdap/wrangler/ExperimentTests.java
```diff
@@ -65,13 +61,4 @@ public void testIdCreationFromName() throws Exception {
     }
   }
 
-  @Test
-  public void testCurrencyParsing() throws Exception {
```
CuriousVini, line 69: why is this removed?

wrangler-core/src/test/java/io/cdap/directives/row/RowRegexFilterTest.java
```diff
@@ -91,7 +91,7 @@ public void testFilterOnNumericValues() throws Exception {
     directives.add("set-type :long long");
     directives.add("set-type :float float");
     directives.add("set-type :double double");
-    directives.add("filter-rows-on condition-false int =^ \"5\"");
+    directives.add("filter-rows-on condition-false int == 5005");
```
CuriousVini, line 94: why is this change needed?

wrangler-core/src/main/java/io/cdap/wrangler/expression/ELContext.java
```diff
@@ -36,6 +37,23 @@
 public class ELContext implements JexlContext {
   private final Map<String, Object> values = new HashMap<>();
 
+  /**
+   * Context object passed to every expresson evaluation.
```
CuriousVini: typo: `expression`

wrangler-core/src/main/java/io/cdap/wrangler/expression/ELContext.java
```diff
@@ -36,6 +37,23 @@
...
+   * All properties of this class are public to ensure they can be accessed with dot(.) format.
+   */
+  public static class Context {
+    public String environment;
```
CuriousVini: why not have these in the ELContext itself?

wrangler-core/src/main/java/io/cdap/wrangler/expression/ELContext.java
```diff
@@ -45,24 +63,40 @@ public ELContext() {
...
+    // because, el variables might not be present in a row.
+    init(context);
+    set("this", row);
+  }
 
-    values.put("runtime", context.getProperties());
-    values.put("environment", context.getEnvironment().name());
-    values.put("context", context.getContextName());
+  private void init(ExecutorContext context) {
+    if (context != null) {
```
CuriousVini, line 96: when can context be null? If context can be null it should be marked as nullable in caller as well as in this method

wrangler-core/src/main/java/io/cdap/wrangler/expression/ELContext.java
```diff
@@ -36,6 +37,23 @@
 public class ELContext implements JexlContext {
   private final Map<String, Object> values = new HashMap<>();
 
+  /**
+   * Context object passed to every expresson evaluation.
+   * All properties of this class are public to ensure they can be accessed with dot(.) format.
+   */
+  public static class Context {
```
CuriousVini, line 45: If this class is accessed by other classes outside this class, why not make it top level class? why should it be nested?

wrangler-core/src/main/java/io/cdap/directives/aggregates/SetTransientVariable.java
```diff
@@ -82,18 +82,7 @@ public void destroy() {
   public List<Row> execute(List<Row> rows, ExecutorContext context) throws DirectiveExecutionException {
     for (Row row : rows) {
       // Move the fields from the row into the context.
-      ELContext ctx = new ELContext(context);
```
CuriousVini, line 85: If this is moved to EL class, does it mean custom directives will need to remove this if they are using it and recompile?

CuriousVini (commented):
A few questions.

wrangler-api/src/main/java/io/cdap/wrangler/api/Row.java
nitinmotgi: no, this is internal only. Now, we are exposing it.

wrangler-api/src/main/java/io/cdap/wrangler/api/ErrorRowException.java
nitinmotgi: Directive is secondary information, I would like to know what the error is first and then find out where it's coming from. That's why I changed the order.

wrangler-core/src/main/java/io/cdap/wrangler/expression/EL.java
nitinmotgi: Depending on the situation `getDetail` can be null sometimes. So u can hit `NulPointerException`.

wrangler-core/src/main/java/io/cdap/wrangler/executor/RecipePipelineExecutor.java
nitinmotgi: it's just a habit from old days.

wrangler-core/src/test/java/io/cdap/wrangler/ExperimentTests.java
nitinmotgi: Was wrong test.

wrangler-core/src/test/java/io/cdap/directives/row/RowRegexFilterTest.java
nitinmotgi: Was a bad test so fixed it.

wrangler-core/src/main/java/io/cdap/wrangler/expression/ELContext.java
nitinmotgi: done

wrangler-core/src/main/java/io/cdap/wrangler/expression/ELContext.java
nitinmotgi: during test

nitinmotgi: done

wrangler-core/src/main/java/io/cdap/directives/aggregates/SetTransientVariable.java
nitinmotgi: These are advanced, I haven't see any one us these type. EL is not exposed outside.

nitinmotgi:
@CuriousVini PTAL

wrangler-api/src/main/java/io/cdap/wrangler/api/Row.java
albertshau: This is a backwards incompatible change and will break any existing UDD. We cannot remove any existing method without adding its replacement and deprecating it for a full major release.

In this case, this rename seems unnecessary and should not be done. If it's unclear, just add clearer javadocs.

wrangler-core/src/main/java/io/cdap/wrangler/expression/EL.java
albertshau: I looked at JexlInfo.toString(), this will result in a weird error message with duplicate information:

```
Error encountered while executing '@<line>:<column> ![<detail.start>,<detail.end>]: '<detail.toString>'' at line '<line>' and column '<column>'
```

Is this really what you intended?

wrangler-core/src/main/java/io/cdap/wrangler/expression/ELContext.java
```diff
@@ -36,6 +38,23 @@
...
 
+  /**
+   * Context object passed to every expression evaluation.
+   * All properties of this class are public to ensure they can be accessed with dot(.) format.
+   */
+  public static class Context {
+    public String environment;
+    public String name;
+    public long nano;
+    public long millis;
```
albertshau: don't think it makes sense for an expression to modify any of these right? If so, make these final otherwise directives can mess each other up.

wrangler-core/src/main/java/io/cdap/wrangler/expression/ELContext.java
```diff
@@ -36,6 +38,23 @@
...
+    public Context (String environment, String name) {
```
albertshau: style: remove extra space after Context

wrangler-core/src/test/java/io/cdap/directives/aggregates/SetTransientVariableTest.java
```diff
@@ -50,6 +50,7 @@ public void testSettingOfVariable() throws Exception {
     rows.add(new Row("fwd", 1).add("A", 2));
 
     final Map<String, Object> s = new HashMap<>();
+    s.put("test", 1);
```
albertshau: is this intended? don't see any change to the rest of the test

wrangler-core/src/test/java/io/cdap/directives/parser/CsvParserTest.java
```diff
@@ -67,4 +67,61 @@ public void testHeaders() throws Exception {
...
+      new Row("body", "14J456,65765566,Granger,Hermione,,,"),
+      new Row("body", "14J456,13233121,Diggory,Cedric,,,"),
+      new Row("body", "14J456,98786868,Weasley,George,,,"),
+      new Row("body", "14J456,78977876,Weasley,Fred,,,")
+    );
+
+    rows = TestingRig.execute(directives, rows);
+    Assert.assertEquals(6, rows.size());
+    for (int i = 0; i < rows.size(); ++i) {
+      Assert.assertEquals(4, rows.get(i).width());
```
albertshau, line 109: tests should basically never just assert on a size. They should always verify the correct values are being generated.

wrangler-core/src/test/java/io/cdap/directives/row/SendToErrorAndContinueTest.java
```diff
@@ -31,6 +31,36 @@
...
+    );
+
+    RecipePipeline pipeline = TestingRig.execute(directives);
+    List<Row> results = pipeline.execute(rows);
+    List<ErrorRecord> errors = pipeline.errors();
+
+    Assert.assertEquals(2, errors.size());
+    Assert.assertEquals(1, results.size());
+    Assert.assertEquals("1", errors.get(0).getRow().getValue("C"));
+    Assert.assertEquals("filtering records that dont have 4 columns (ecode: 1)", errors.get(0).getMessage());
```
albertshau: these types of assertions are very brittle and easily broken. For example, "dont" should actually be "don't" or "do not". Fixing it would break this test.

wrangler-core/src/main/java/io/cdap/directives/row/RecordConditionFilter.java
```diff
@@ -91,19 +91,7 @@ public void destroy() {
     List<Row> results = new ArrayList<>();
     for (Row row : rows) {
       // Move the fields from the row into the context.
-      ELContext ctx = new ELContext(context);
-      for (String var : el.variables()) {
-        Object value = row.getValue(var);
-        String strValue;
-        // support numeric values by converting them to string
-        if (value instanceof Number) {
-          Number number = (Number) value;
```
albertshau, line 100: Looks like this number conversion will no longer happen.  I don't know why numbers had to be converted to strings to begin with, but I imagine this change will break existing pipelines.

albertshau: is there test coverage for number input to this directive?

wrangler-core/src/main/java/io/cdap/wrangler/expression/ELContext.java
nitinmotgi: done!

wrangler-core/src/test/java/io/cdap/directives/aggregates/SetTransientVariableTest.java
nitinmotgi: unintended.

wrangler-core/src/test/java/io/cdap/directives/parser/CsvParserTest.java
nitinmotgi: This is testing if all columns are correct or not. The intent of test was to check for right columns. anyway added each column test.

wrangler-core/src/test/java/io/cdap/directives/row/SendToErrorAndContinueTest.java
nitinmotgi: removed

wrangler-core/src/main/java/io/cdap/directives/row/RecordConditionFilter.java
nitinmotgi: Was not there, There was no way to convert something to an number, so this path never happened. So, removed it.

nitinmotgi:
@albertshau @CuriousVini PTAL.

wrangler-api/src/main/java/io/cdap/wrangler/api/Row.java
```diff
@@ -170,12 +170,20 @@ public int find(String col) {
   }
 
   /**
-   * @return  Length of the row.
+   * @return  width of the row.
```
albertshau, line 173: add a `@deprecated` to javadoc and mention what should be used instead.

wrangler-core/src/test/java/io/cdap/directives/parser/CsvParserTest.java
```diff
@@ -67,4 +67,79 @@ public void testHeaders() throws Exception {
...
+      new Row("school_id", "14J456").add("student_id", "13233121").add("last_name", "Diggory")
+        .add("first_name", "Cedric"),
+      new Row("school_id", "14J456").add("student_id", "98786868").add("last_name", "Weasley")
+        .add("first_name", "George"),
+      new Row("school_id", "14J456").add("student_id", "78977876").add("last_name", "Weasley")
+        .add("first_name", "Fred")
+    );
+
+    rows = TestingRig.execute(directives, rows);
+    Assert.assertEquals(6, rows.size());
```
albertshau, line 107: can you just do `Assert.assertEquals(expected, rows)`?

wrangler-core/src/test/java/io/cdap/directives/parser/CsvParserTest.java
```diff
@@ -67,4 +67,79 @@ public void testHeaders() throws Exception {
...
+    };
+
+    List<Row> rows = Arrays.asList(
+      new Row("body", "school_id, student_id, last_name, first_name"),
+      new Row("body", "14J456,0033445566,Potter,Jr,Harry"),
+      new Row("body", "14J456,0044333433,Weasley,Ron")
+    );
+
+    rows = TestingRig.execute(directives, rows);
+    Assert.assertEquals(2, rows.size());
```
albertshau, line 135: similar comment here

albertshau (approved):
minor comments, otherwise lgtm. please fix, squash, then merge

PR #410, opened 2020-05-06
CDAP-16724 fix bad sampling in GCS handler

albertshau:
Fixed the GCS handler to create less duplicate data. Instead of
copying data multiple times, try to go through the file contents
once and just maintain a single list of Rows.

Also limited the number of rows in the sample to 5000. This is
because the Row class uses a lot of memory, so the handler was
going out of memory reading just 10mb of an input file, where that
10mb contained a million rows.

Also fixed a bug in Row that made one of the constructors useless
and changed another constructor to set the initial the array size
to reduce memory usage.

### DIFF
--- a/wrangler-service/src/main/java/io/cdap/wrangler/service/gcs/GCSHandler.java
+++ b/wrangler-service/src/main/java/io/cdap/wrangler/service/gcs/GCSHandler.java
@@ -24,6 +24,7 @@ import com.google.cloud.storage.BlobId;
 import com.google.cloud.storage.Bucket;
 import com.google.cloud.storage.Storage;
 import com.google.cloud.storage.StorageException;
+import com.google.common.base.Splitter;
 import io.cdap.cdap.api.annotation.TransactionControl;
 import io.cdap.cdap.api.annotation.TransactionPolicy;
 import io.cdap.cdap.api.service.http.HttpServiceRequest;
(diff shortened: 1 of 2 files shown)


wrangler-service/src/main/java/io/cdap/wrangler/service/gcs/GCSHandler.java
```diff
@@ -88,6 +89,7 @@
  */
 public class GCSHandler extends AbstractWranglerHandler {
   private static final Logger LOG = LoggerFactory.getLogger(GCSHandler.class);
+  private static final int MAX_SAMPLE_ROWS = 5000;
```
CuriousVini: if we make this change does this mean users who were able to browse more than 5000 records will see different behavior in case first 5000 records had null values in columns (the column types will be unknown)?

CuriousVini (approved):
just one question. lgtm 👍

wrangler-service/src/main/java/io/cdap/wrangler/service/gcs/GCSHandler.java
albertshau: It's an unlikely scenario, as they would have to small enough data that it wasn't going OOM. 

But it's a good point, we should have a way to change it. I think I'll change this to a default and look in the runtime arguments for a max value as an override.

PR #411, opened 2020-05-06
CDAP-16724 fix bad sampling in GCS handler

albertshau:
Fixed the GCS handler to create less duplicate data. Instead of
copying data multiple times, try to go through the file contents
once and just maintain a single list of Rows.

Also limited the number of rows in the sample to 5000. This is
because the Row class uses a lot of memory, so the handler was
going out of memory reading just 10mb of an input file, where that
10mb contained a million rows.

Also fixed a bug in Row that made one of the constructors useless
and changed another constructor to set the initial the array size
to reduce memory usage.


albertshau:
cherrypick of #410

mikkin (approved):

PR #414, opened 2020-05-19
remove -SNAPSHOT from release/4.2

elfenheart:


albertshau (commented):
lower bound of parents in wrangler-transform/pom.xml should be bumped up to 6.2.0

albertshau (approved):

PR #423, opened 2020-07-25
Treat JsonNull as null value in FillNullOrEmpty directive

DougHamil:
I noticed that using the parse-as-json directive can result in cells with JsonNull values that cannot be altered through the fill-null-or-empty directive. This change checks for JsonNull objects and treats them as a null value.


shwei (approved):

shwei (approved):
The addition of checking for JsonNull has been tested and verified.

PR #430, opened 2020-08-31
[PLUGIN-386] Displaying BigQuery MaterializedViews in Wrangler

MEseifan:
Wrangler was unable to load BigQuery datasets that contained MaterializedViews. This was caused by a bug in the BigQuery java library which was fixed in March 2020. This PR updates the Google dependencies to resolve this bug and ensure there are no version conflicts.


albertshau:
The source doesn't support views either. Are we planning to add support in the same release? Otherwise it will break later when you go from wrangler to a pipeline.

albertshau (approved):
changes lgtm, assuming we integrate this with BQ view support in the source.

PR #431, opened 2020-09-10
 [PLUGIN-386] Displaying BigQuery MaterializedViews in Wrangler

MEseifan:
Wrangler was unable to load BigQuery datasets that contained MaterializedViews. This was caused by a bug in the BigQuery java library which was fixed in March 2020. This PR updates the Google dependencies to resolve this bug and ensure there are no version conflicts.

This PR also contains changes to integrate with the new version of google-cloud plugins that supports querying Views and Materialized Views

### DIFF
--- a/wrangler-service/pom.xml
+++ b/wrangler-service/pom.xml
@@ -29,7 +29,7 @@
     <app.main.class>io.cdap.wrangler.DataPrep</app.main.class>
     <tika.version>1.14</tika.version>
     <hadoop.version>2.3.0</hadoop.version>
-    <google.cloud.spanner.version>1.20.0</google.cloud.spanner.version>
+    <google.cloud.spanner.version>1.51.0</google.cloud.spanner.version>
   </properties>
 
   <dependencies>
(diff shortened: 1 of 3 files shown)


mikkin (approved):

wrangler-service/pom.xml
```diff
@@ -29,7 +29,7 @@
     <app.main.class>io.cdap.wrangler.DataPrep</app.main.class>
     <tika.version>1.14</tika.version>
     <hadoop.version>2.3.0</hadoop.version>
-    <google.cloud.spanner.version>1.20.0</google.cloud.spanner.version>
+    <google.cloud.spanner.version>1.51.0</google.cloud.spanner.version>
```
CuriousVini, line 32: why do we need to change spanner version?

MEseifan: To resolve dependency version conflicts, if we only update the bigquery version it causes the spanner connection to break due to a conflict in the protobuf library

CuriousVini (approved):
lgtm

PR #442, opened 2020-10-12
add null handling functions

yaojiefeng:


wrangler-core/src/main/java/io/cdap/functions/Global.java
```diff
@@ -213,4 +213,74 @@ public static String unquote(String string) {
...
+
+  /**
+   * Returns an empty string if the input column is null, otherwise returns the input column value.
+   *
+   * @param value to be evaluated.
+   * @return Empty string if null, else 'value'
+   */
+  public static Object NullToEmpty(Object value) {
+    if (IsNull(value)) {
+      return new String("");
```
albertshau: return ""

wrangler-core/src/main/java/io/cdap/functions/Global.java
```diff
@@ -213,4 +213,74 @@ public static String unquote(String string) {
...
+    if (IsNull(value)) {
+      return replace;
+    }
+    return value;
+  }
+
+  /**
+   * @return Assigns a null value to the target column.
+   */
+  public static Object SetNull() {
```
albertshau: I don't see a test for this, but would the user use this like `set-column col SetNull()`? Seems weird, couldn't you just do `set-column col null`?

wrangler-docs/functions/null-handling.md
```diff
@@ -0,0 +1,125 @@
...
+If the input column does contain a `null`, then the output column contains the string `null`. In the below
+examples, `col1` is `0`, `col2` is `0`, and `col3` is `999`.
+
+```
+set-column col1 NullToZero(a)
+set-column col2 NullToZero(b == 'value' ? a : b)
+set-column col3 NullToZero(c)
+```
+
+## NullToValue
```
albertshau, line 85: this section is copy-pasted with wrong information

yaojiefeng: Fixed

wrangler-core/src/main/java/io/cdap/functions/Global.java
yaojiefeng: Yea `set-column col null` will have same effect. This is just to simulate what data stage has. I have added a test for it.

albertshau: let's remove this, I don't think it makes sense to have it.

wrangler-docs/functions/null-handling.md
```diff
@@ -0,0 +1,125 @@
...
+boolean (true/false)
+
+#### Example
+If `a` contains the `null`, `b` contains string `value` and `c` contains number `999` then
+If the input column does not contain a `null`, the output column contains the value of the input column.
+If the input column does contain a `null`, then the output column contains the string `null`. In the below
+examples, `col1` is `null`, `col2` is `value`, `col3` is `999` and `col4` is `null`.
+
+```
+set-column col1 IsNotNull(a) ? a : null
```
albertshau: we should replace this with a more useful example, as this is just a no-op. The example description is also confusing because of this. Use just a single expression like

set-column col1 IsNotNull(a) ? a : 0

Then explain what col1 evaluates to if a is null and if a is some value like 999.

wrangler-docs/functions/null-handling.md
```diff
@@ -0,0 +1,125 @@
...
+boolean (true/false)
+
+#### Example
+If `a` contains the `null`, `b` contains string `value` and `c` contains number `999` then
+If the input column does not contain a `null`, the output column contains the value of the input column.
+If the input column does contain a `null`, then the output column contains the string `null`. In the below
+examples, `col1` is `null`, `col2` is `null`, `col3` is `null` and `col4` is `value`.
+
+```
+set-column col1 IsNull(a) ? a : null
```
albertshau: similar comment here about using a simpler, more easily understood example.

wrangler-docs/functions/null-handling.md
```diff
@@ -0,0 +1,125 @@
...
+input value or string
+
+#### Example
+If `a` contains the `null`, `b` contains string `value` and `c` contains number `999` then
+If the input column does not contain a `null`, the output column contains the value of the input column.
+If the input column does contain a `null`, then the output column contains the string `null`. In the below
+examples, `col1` is `''`, `col2` is `value`, and `col3` is `999`.
+
+```
+set-column col1 NullToEmpty(a)
```
albertshau, line 62: similar comment here. Single directive, with a description of what happen if a is null and if a is some other value.

wrangler-docs/functions/null-handling.md
```diff
@@ -0,0 +1,125 @@
...
+column value or zero.
+
+#### Example
+If `a` contains the `null`, `b` contains string `value` and `c` contains number `999` then
+If the input column does not contain a `null`, the output column contains the value of the input column.
+If the input column does contain a `null`, then the output column contains the string `null`. In the below
+examples, `col1` is `0`, `col2` is `0`, and `col3` is `999`.
+
+```
+set-column col1 NullToZero(a)
```
albertshau, line 82: similar comment here

wrangler-docs/functions/null-handling.md
yaojiefeng: Replaced all methods with more understandable example.

albertshau (approved):

PR #443, opened 2020-10-12
add logical

yaojiefeng:


wrangler-core/src/main/java/io/cdap/functions/Logical.java
```diff
@@ -0,0 +1,147 @@
...
+   * @return a result of bitwise 'XOR'
+   */
+  public static long BitXor(long num1, long num2) {
+    return num1 ^ num2;
+  }
+
+  /**
+   * Returns the complement of the logical value of an expression. If the value of expression is true,
+   * the Not function returns a value of false (0). If the value of expression is false, the NOT function
+   * returns a value of true (1). A numeric expression that evaluates to 0 is a logical value of false.
```
albertshau: this description doesn't match the behavior

wrangler-core/src/main/java/io/cdap/functions/Logical.java
```diff
@@ -0,0 +1,147 @@
...
+    }
+    return 1;
+  }
+
+  /**
+   * Returns the complement of the logical value of an expression. If the value of expression is true,
+   * the Not function returns a value of false (0). If the value of expression is false, the NOT function
+   * returns a value of true (1). A numeric expression that evaluates to 0 is a logical value of false.
+   * A numeric expression that evaluates to anything else, other than the null value, is a logical true.
+   * An empty string is logically false. All other string expressions, including strings that include
```
albertshau, line 111: behavior doesn't match the description (empty string is not checked)

wrangler-core/src/main/java/io/cdap/functions/Logical.java
```diff
@@ -0,0 +1,147 @@
...
+   */
+  public static short Not(String val) {
+    if (val == null) {
+      return 0;
+    }
+    return 1;
+  }
+
+  /**
+   * Returns the integer made from the string argument, which contains a binary representation of "1"s and "0"s.
```
albertshau: is it supposed to return a long or an int?

wrangler-core/src/main/java/io/cdap/functions/Logical.java
```diff
@@ -0,0 +1,147 @@
...
+   *
+   * @param value to be compressed.
+   * @return long value of compressed binary string.
+   */
+  public static long BitCompress(String value) {
+    long result = Long.parseUnsignedLong(value, 2);
```
albertshau: can return directly, don't need variable

wrangler-core/src/main/java/io/cdap/functions/Logical.java
```diff
@@ -0,0 +1,147 @@
...
+   * the Not function returns a value of false (0). If the value of expression is false, the NOT function
+   * returns a value of true (1). A numeric expression that evaluates to 0 is a logical value of false.
+   * A numeric expression that evaluates to anything else, other than the null value, is a logical true.
+   * An empty string is logically false. All other string expressions, including strings that include
+   * an empty string, spaces, or the number 0 and spaces, are logically true.
+   *
+   * @param exp expression result.
+   * @return 1 or 0.
+   */
+  public static short Not(double exp) {
```
albertshau: I'm confused by the method signatures. Why does this return a short and not a boolean? How come there is just one method that takes a double? Does jexl cast things automatically, or will the directive break if the wrong type is passed in?

wrangler-core/src/main/java/io/cdap/functions/Logical.java
yaojiefeng: Fixed

wrangler-core/src/main/java/io/cdap/functions/Logical.java
yaojiefeng: Fixed

wrangler-core/src/main/java/io/cdap/functions/Logical.java
yaojiefeng: Should be long, fixed

wrangler-core/src/main/java/io/cdap/functions/Logical.java
yaojiefeng: Changed it to int based on our discussion

wrangler-core/src/main/java/io/cdap/functions/Logical.java
```diff
@@ -0,0 +1,218 @@
...
+   */
+  public static long BitXor(int num1, int num2) {
+    return num1 ^ num2;
+  }
+
+  /**
+   * Returns the complement of the logical value of an expression. If the value of expression is true,
+   * the Not function returns a value of false (0). If the value of expression is false, the NOT function
+   * returns a value of true (1). A numeric expression that evaluates to 0 is a logical value of false.
+   * A numeric expression that evaluates to anything else, other than the null value, is a logical true.
```
albertshau: this talks about null, but the function can't take nulls

wrangler-core/src/main/java/io/cdap/functions/Logical.java
```diff
@@ -0,0 +1,218 @@
...
+   * An empty string is logically false. All other string expressions, including strings that include
```
albertshau: shouldn't mention other unrelated types

wrangler-core/src/main/java/io/cdap/functions/Logical.java
```diff
@@ -0,0 +1,218 @@
...
+   * @param val expression result.
+   * @return 1 or 0.
+   */
+  public static int Not(double val) {
+    return val == 0d ? 1 : 0;
+  }
+
+  /**
+   * Returns the complement of the logical value of an expression. If the value of expression is true,
+   * the Not function returns a value of false (0). If the value of expression is false, the NOT function
```
albertshau: javadoc should be different depending on the input type

wrangler-core/src/main/java/io/cdap/functions/Logical.java
```diff
@@ -0,0 +1,218 @@
...
+   *
+   * @param value to be expanded into '1' and '0' binary representation.
+   * @return a binary string representation of value.
+   */
+  public static String BitExpand(long value) {
+    return Long.toBinaryString(value);
+  }
+
+  /**
+   * Returns an long with specific bits set to a specific state.
```
albertshau: nit: an long -> a long

wrangler-core/src/main/java/io/cdap/functions/Logical.java
```diff
@@ -0,0 +1,218 @@
...
+    if (bit > 0) {
+      bit = 1;
+    } else {
+      bit = 0;
+    }
+
+    long result = value;
+    for (int position : positions) {
+      position = position - 1;
+      if (bit == 1) {
```
albertshau: better for this to be a boolean comparison rather than check for 1.

wrangler-core/src/main/java/io/cdap/functions/Logical.java
```diff
@@ -0,0 +1,218 @@
...
+  }
+
+  /**
+   * Bitwise 'AND' operation of two numbers. The inputs must be non-null.
+   *
+   * @param num1 first number.
+   * @param num2 second number.
+   * @return a result of bitwise 'AND'.
+   */
+  public static long BitAnd(int num1, int num2) {
```
albertshau: shouldn't this return an int?

wrangler-core/src/main/java/io/cdap/functions/Logical.java
```diff
@@ -0,0 +1,218 @@
...
+  }
+
+  /**
+   * Bitwise 'OR' operation of two numbers. The inputs must be non-null.
+   *
+   * @param num1 first number.
+   * @param num2 second number.
+   * @return a result of bitwise 'OR'
+   */
+  public static long BitOr(int num1, int num2) {
```
albertshau: shouldn't this return an int?

wrangler-core/src/main/java/io/cdap/functions/Logical.java
```diff
@@ -0,0 +1,218 @@
...
+  }
+
+  /**
+   * Bitwise 'XOR' operation of two numbers. The inputs must be non-null.
+   *
+   * @param num1 first number.
+   * @param num2 second number.
+   * @return a result of bitwise 'XOR'
+   */
+  public static long BitXor(int num1, int num2) {
```
albertshau: shouldn't this return an int?

wrangler-docs/functions/logical.md
```diff
@@ -0,0 +1,161 @@
...
+`logical` namespace. To use the functions defined in this namespace prefix function invocation with `logical:<func>`.
+
+## BitAnd
+Returns the bitwise `AND` of the two arguments. The inputs must be non-null.
+
+#### Namespace
+`logical`
+
+#### Input
+number1(long, int), number2(long, int)
```
albertshau: Is this the standard for how the other docs describe arguments? This way of describing the arguments seems much harder to understand than just saying it supports two ints or two longs.

wrangler-docs/functions/logical.md
```diff
@@ -0,0 +1,161 @@
...
+
+#### Output
+long
```
albertshau: int inputs should return an int

wrangler-docs/functions/logical.md
```diff
@@ -0,0 +1,161 @@
...
+which contains a binary representation of `"1"s` and `"0"s`.
+
+#### Namespace
+`logical`
+
+#### Input
+String
+
+#### Output
+number(long)
```
albertshau, line 80: should this just be long?

wrangler-docs/functions/logical.md
```diff
@@ -0,0 +1,161 @@
...
+If `a` contains the number `352` and `b` contains the number `400`,
+then the following two functions are equivalent, and return the value `240`:
+
+```
+set-column val logical:BitXOr(352,400)
+set-column val logical:BitXOr(a,b)
+```
+
+## BitCompress
+Returns the `integer` made from the string argument,
```
albertshau: this does not match what's documented for output

wrangler-docs/functions/logical.md
```diff
@@ -0,0 +1,161 @@
...
+```
+set-column val logical:BitExpand(352)
+set-column val logical:BitExpand(mynumber)
+```
+
+## Not
+Returns the Not of the logical value of an expression. If the value of expression is `true`, the Not function
+returns a value of `false (0)`. If the value of expression is `false`, the `NOT` function returns a value of `true (1)`.
+A numeric expression that evaluates to `0` is a logical value of `false`. A numeric expression that evaluates to
+anything else, other than the `null` value, is a logical `true`. An empty string is logically `false`. All other string
```
albertshau: The numeric Not functions don't allow null. what happens if a null is passed in? Does it automatically get sent to the Not(string)?

albertshau: it's confusing to say something like 'An empty string is logically false' because you have to invert it to understand that the function actually returns true if you pass in an empty string. It would be more clear to just say what the function returns given various input.

wrangler-docs/functions/logical.md
```diff
@@ -0,0 +1,161 @@
...
+expressions, including strings that include an empty string, spaces, or the number `0` and spaces, are logically `true`.
+
+#### Namespace
+`logical`
+
+#### Input
+expression(double, float, int, long, String)
+
+#### Output
+number (int)
```
albertshau: should this just be int? Or should the other functions say number(type)? The docs need to be consistent everywhere.

wrangler-docs/functions/logical.md
```diff
@@ -0,0 +1,161 @@
...
+If `mynumber` contains the number `352`, then the following two functions are equivalent,
+and return the string `"101100000"`
+```
+set-column val logical:Not(5-5)
+set-column val logical:Not(5+5)
+set-column val logical:Not('value')
+```
+
+## SetBit
+Returns an `long` with specific bits set to a specific state, where `'origValue'` is the input value to perform
```
albertshau: an long -> a long

wrangler-docs/functions/logical.md
```diff
@@ -0,0 +1,161 @@
...
+the action on, `'bitArray'` is a array containing a list bit numbers (starting from 1) to set the state of,
+and `'bitState'` is either `1` or `0`, indicating which state to set those bits.
+
+#### Namespace
+`logical`
+
+#### Input
+
+ - origValue (long)
```
albertshau: Here the inputs are on different lines, previously they are on the same line. Please make sure we use one consistent style.

wrangler-docs/functions/logical.md
yaojiefeng: Modified this to explicitly say the inputs has to be both long or both int

wrangler-docs/functions/logical.md
yaojiefeng: Fixed

wrangler-docs/functions/logical.md
yaojiefeng: Made the doc to say number(type), the datastage doc basically has this structure, for number related input/output, the type is in the parentheses.

wrangler-docs/functions/logical.md
yaojiefeng: null will get sent to  Not(string)

yaojiefeng: Fixed the description

albertshau (approved):

PR #444, opened 2020-10-12
add new json functions

yaojiefeng:


wrangler-core/src/main/java/io/cdap/functions/JsonFunctions.java
```diff
@@ -197,12 +218,121 @@ public static JsonElement parse(String json) {
...
     return element;
   }
+
+  /**
+   * Parses a column or string to JSON. This is equivalent to <code>JSON.parse()</code>
+   *
+   * @param json string representation of json.
+   * @return parsed json else throws an exception.
+   */
+  public static JsonElement Parse(String json) {
```
albertshau, line 236: this behaves differently than the old parse() implementation. Is that intended?

wrangler-core/src/main/java/io/cdap/functions/JsonFunctions.java
```diff
@@ -197,12 +218,121 @@ public static JsonElement parse(String json) {
...
+    try {
+      JsonElement element = PARSER.parse(json);
+      return element;
+    } catch (JsonSyntaxException e) {
+      return JsonNull.INSTANCE;
```
albertshau, line 241: is this the intended behavior? Seems odd to return null for an invalid value.

wrangler-core/src/main/java/io/cdap/functions/JsonFunctions.java
```diff
@@ -197,12 +218,121 @@ public static JsonElement parse(String json) {
...
+   * Checks if a Json is {@code JsonNull}.
+   *
+   * @param element to be inspected for null.
+   * @return true if null, false otherwise.
+   */
+  public static boolean IsNull(JsonElement element) {
+    if (element != null && element.isJsonNull()) {
+      return true;
+    }
+    return false;
```
albertshau: if the element itself is a null this returns false? Is that intended?

wrangler-core/src/main/java/io/cdap/functions/JsonFunctions.java
```diff
@@ -197,12 +218,121 @@ public static JsonElement parse(String json) {
...
+    return element != null && element.isJsonArray();
+  }
+
+  /**
+   * Selects part of JSON using JSON DSL specified as json path.
+   *
+   * @param element json to be inspected.
+   * @param path to be searched for in the element.
+   * @param paths other paths.
+   * @return A json element containing the results of all json paths.
```
albertshau: json array?

wrangler-core/src/test/java/io/cdap/functions/JsonFunctionsTest.java
```diff
@@ -0,0 +1,94 @@
...
+    String[] directives = new String[]{
+      "set-column mayo json:Parse(body)",
+      "drop boody",
+      "set-column entries json:Select(mayo, '$.list[*].set.*')"
+    };
+    List<Row> rows = Arrays.asList(
+      new Row("body", JSON_SELECTION_EG1)
+    );
+    rows = TestingRig.execute(directives, rows);
+    Assert.assertTrue(rows.size() == 1);
```
albertshau, line 76: The directives could be doing all sorts of wrong things and this test would still pass. It should be checking the output.

wrangler-docs/functions/json.md
```diff
@@ -0,0 +1,207 @@
...
+    {
+      a : 1,
+      b : 2.0,
+      c : test
+      d : true
+    }
+```
+
+## Parse
+Parses a json into `json-element`. If there are any issues in parsing, the function would return
```
albertshau: would return -> returns

albertshau: same for other descriptions.

wrangler-docs/functions/json.md
```diff
@@ -0,0 +1,207 @@
...
+
+if `json` has the input json specified, then the result of the operation would return `true`.
+
+```
+  set-column parsedjson json:Parse(json)
+  set-column object json:IsObject(parsedjson)
+```
+
+## Select
+Returns part of json specified by json path.
```
albertshau: Is json path a standard? I would not know how to use this from the doc.

wrangler-docs/functions/json.md
```diff
@@ -0,0 +1,207 @@
...
+
+### Input
+`json-element`
+
+### Output
+boolean(`true`|`false`)
+
+### Example
+
+if `json` has the input json specified, then the result of the operation would return `true`.
```
albertshau: should be consistent with the examples. Some have 'json' as the field, others have 'body'. 'body' seems slightly better, as the term 'json' is mentioned a bunch of times.

albertshau:
I'm concerned about adding two copies of what look the same functions (parse vs Parse), except they behave differently at edge conditions.

I also don't know why it's important to change things to upper case. It looks like there are a bunch of existing lowercase functions, why are we making things inconsistent?

wrangler-core/src/main/java/io/cdap/functions/JsonFunctions.java
yaojiefeng: I think the new implementation will not throw exception if the input is invalid json, instead it treats it as a json null, I think that is intended, though I am unsure if it is correct to do this

wrangler-core/src/main/java/io/cdap/functions/JsonFunctions.java
yaojiefeng: Same above

wrangler-core/src/main/java/io/cdap/functions/JsonFunctions.java
yaojiefeng: Fixed

wrangler-docs/functions/json.md
yaojiefeng: Fixed

wrangler-docs/functions/json.md
```diff
@@ -0,0 +1,207 @@
...
+
+## Parse
+Parses a json into `json-element`. If there are any issues in parsing, the function returns
+`json-null`
+
+### Namespace
+`json`
+
+### Input
+`json-string`
```
albertshau: the convention here looks different than in other PRs. Please ensure they are consistent

wrangler-docs/functions/json.md
```diff
@@ -0,0 +1,207 @@
...
+
+### Input
+`json-element`
+
+### Output
+boolean(`true`|`false`)
+
+### Example
+
+if `body` has the input json specified, then the result of the operation returns `false`
```
albertshau: if -> If

albertshau (approved):

PR #445, opened 2020-10-12
modfiy geo functions

yaojiefeng:


wrangler-core/src/main/java/io/cdap/functions/GeoFences.java
```diff
@@ -47,7 +47,7 @@
    * @param geofences GeoJson representation of the fence area
    * @return true if location is inside any of the given geofences, else false
    */
-  public static Boolean inFence(double latitude, double longitude, String geofences) {
+  public static Boolean InFence(double latitude, double longitude, String geofences) {
```
albertshau: this is a backwards incompatible change, unless this method didn't exist before

yaojiefeng: Added back old methods and deprecated it.

wrangler-docs/functions/geo.md
```diff
@@ -0,0 +1,83 @@
...
+
+### Input
+latitude(`double`), longitude(`double`), json fence (`string`)
+
+### Output
+true/false
+
+### Example
+if `latitude` contains `37.378990156513105` and `longitude` contains `-122.05076694488525` and
+`fence` contains a json `string`, then resulting operation is `true`
```
albertshau: contains a json string -> contains the previously mentioned json string

albertshau (approved):

PR #446, opened 2020-10-12
add number functions

yaojiefeng:


wrangler-core/src/main/java/io/cdap/functions/NumberFunctions.java
```diff
@@ -0,0 +1,63 @@
...
+   */
+  public static double MantissaFromDFloat(Number value) {
+    return getMantissa(value.toString());
+  }
+
+  private static double getMantissa(String number) {
+    if (number.indexOf('.') < 0) {
+      return 0d;
+    }
+    return Double.valueOf(number.substring(number.indexOf('.') + 1));
```
albertshau: this doesn't seem correct. In the example given at https://www.wolframalpha.com/input/?i=mantissa, the mantissa for 3.14159 is 0.14159, not 14159.

This particular implementation also has a weird behavior around leading zeros. mantissa(3.05) == mantissa(3.0005), which seems bad.

wrangler-core/src/main/java/io/cdap/functions/NumberFunctions.java
```diff
@@ -0,0 +1,63 @@
...
+ */
+
+package io.cdap.functions;
+
+public final class NumberFunctions {
+
+  /**
+   * @return The number as double.
+   */
+  public static double AsDouble(Number value) {
```
albertshau: if a null is passed in, this will error. What is the expected behavior for nulls?

wrangler-core/src/main/java/io/cdap/functions/NumberFunctions.java
yaojiefeng: Changed this to match definition, also renamed to Mantissa

wrangler-core/src/main/java/io/cdap/functions/NumberFunctions.java
yaojiefeng: Added handling of null, the method will now return null

wrangler-core/src/main/java/io/cdap/functions/NumberFunctions.java
```diff
@@ -0,0 +1,86 @@
...
+   * @return Returns the mantissa from the given number. Mantissa definition is used from
+   * https://mathworld.wolfram.com/Mantissa.html
+   */
+  public static double Mantissa(double value) {
+    return getMantissa(String.valueOf(value));
+  }
+
+  private static double getMantissa(String value) {
+    // use string for calculation here since Math.floor will lose some precision.
+    if (value.indexOf('.') < 0) {
```
albertshau: seems like there should be a better way of doing this than converting to a string and back to a double.

wrangler-core/src/main/java/io/cdap/functions/NumberFunctions.java
```diff
@@ -0,0 +1,86 @@
...
+   */
+  public static double Mantissa(float value) {
+    return getMantissa(String.valueOf(value));
+  }
+
+  /**
+   * @return Returns the mantissa from the given number. Mantissa definition is used from
+   * https://mathworld.wolfram.com/Mantissa.html
+   */
+  public static double Mantissa(double value) {
```
albertshau, line 76: should we have a BigDecimal version? floating point math can get this close but wrong.

wrangler-core/src/main/java/io/cdap/functions/NumberFunctions.java
yaojiefeng: Due to the way how double is represented, if we use any math related methods to get the decimal part, it will only give the java representation of the decimal part. I have changed this to use BigDecimal to compute, but that still converts value to string. There are multiple stackoverflow threads about this and converting to String seems to be the easiest way to do it.

wrangler-core/src/main/java/io/cdap/functions/NumberFunctions.java
yaojiefeng: Added the bigdecimal version

albertshau (approved):

PR #447, opened 2020-10-12
Add DateTime functionality

yaojiefeng:


wrangler-core/src/main/java/io/cdap/functions/DateAndTime.java
```diff
@@ -0,0 +1,1096 @@
...
+   * @return return current time in milliseconds.
+   */
+  public static long CurrentTimeMS() {
+    return CurrentTime().getNano() / 1000000L;
+  }
+
+  /**
+   * @return return current data & time.
+   */
+  public static LocalDateTime CurrentTimestamp() {
```
albertshau: should this be a ZonedDateTime? A timestamp usually refers to an exact point in time

wrangler-docs/functions/date-and-time-functions.md
```diff
@@ -0,0 +1,961 @@
...
+Returns `time` in milliseconds when the function in invoked.
+
+#### Namespace
+`datetime`
+
+#### Input
+None
+
+#### Output
+time (`time`)
```
albertshau: these docs should all mention the type of the output and the type(s) of input. Should also mention what happens to null input.

wrangler-core/src/main/java/io/cdap/functions/DateAndTime.java
albertshau: still have this question.

wrangler-core/src/main/java/io/cdap/functions/DateAndTime.java
```diff
@@ -0,0 +1,663 @@
...
+    return time == null ? null : time.plusSeconds(seconds == null ? 0 : seconds);
+  }
+
+  /**
+   * Returns a timestamp from the given epoch time specified in seconds. If timestamp is null, null will be returned.
+   *
+   * @param epochTimestamp specified in seconds.
+   * @return a instance {@link LocalDateTime}
+   */
+  public static LocalDateTime TimestampFromEpoch(Long epochTimestamp) {
```
albertshau: any timestamp seems like it should be using ZonedDateTime, as it represents an exact point in time. That is what StructuredRecord uses too.

wrangler-core/src/main/java/io/cdap/functions/DateAndTime.java
```diff
@@ -0,0 +1,663 @@
...
+  }
+
+  /**
+   * Returns a timestamp from the given date and time. If any input is null, null will be returned.
+   *
+   * @param date a instance of {@link LocalDate}
+   * @param time a instance of {@link LocalTime}
+   * @return a instance of {@link LocalDateTime}
+   */
+  public static LocalDateTime TimestampFromDateTime(LocalDate date, LocalTime time) {
```
albertshau: A timestamp normally represents an actual point in time, which different than what these functions do.

wrangler-core/src/test/java/io/cdap/functions/DateAndTimeTest.java
```diff
@@ -0,0 +1,521 @@
...
+  public void testDateFromDaysSince() throws Exception {
+    String[] directives = new String[]{
+      "set-column DateFromDaysSince1 datetime:DateFromDaysSince(18250, datetime:GetDate('1958-08-18'))", // 2008-08-05
+      "set-column DateFromDaysSince2 datetime:DateFromDaysSince(-1, datetime:GetDate('1958-08-18'))", // 1958-08-17
+    };
+
+    List<Row> rows = Arrays.asList(new Row());
+    rows = TestingRig.execute(directives, rows);
+    Assert.assertTrue(rows.size() == 1);
+    Assert.assertEquals(DateAndTime.GetDate("2008-08-05"), rows.get(0).getValue("DateFromDaysSince1"));
```
albertshau: should create the expected object directly instead of calling the library that is being tested.

wrangler-docs/functions/date-and-time-functions.md
```diff
@@ -0,0 +1,1010 @@
...
+## DateFromDaysSince
+Returns a date object by adding an integer to a baseline date.
+The integer can be negative to return a date that is earlier than the baseline date.
+If baseline is null, null will be returned. If the number is null, it will be treated as 0. 
+
+#### Namespace
+`datetime`
+
+#### Input
+number (`Integer`), \[baseline date(`LocalDate`)\]
```
albertshau, line 221: baseline date shouldn't be in brackets should it? From the description it seems like it's a required argument. If it's optional the description should be updated.

wrangler-docs/functions/date-and-time-functions.md
```diff
@@ -0,0 +1,1010 @@
...
+date(`LocalDate`)
+
+#### Output
+week(`Integer`)
+
+#### Example
+If `basedate` contains the date `2008-08-18`, then the two following functions are equivalent,
+and return the value `33`
+```
+set-column result datetime:YearweekFromDate('2008-08-18')
```
albertshau: these examples all need to be updated since the functions take LocalDate objects and not strings.

wrangler-docs/functions/date-and-time-functions.md
yaojiefeng: This is because DateFromDaysSince has two methods, one of them does not take the LocalDate, I have updated the description to match the behavior

wrangler-core/src/main/java/io/cdap/functions/DateAndTime.java
yaojiefeng: The DataStage date and time functions basically do not operate on a ZonedDateTime, here is their doc: https://www.ibm.com/support/knowledgecenter/SSZJPZ_11.7.0/com.ibm.swg.im.iis.ds.parjob.dev.doc/topics/r_deeref_Date_and_Time_Functions.html. 
We have also heard about use cases about these functions used in LocalDateTime. I feel these functions are more like the Datetime type in Bigquery.

wrangler-core/src/main/java/io/cdap/functions/DateAndTime.java
yaojiefeng: The DataStage date and time functions basically do not operate on a ZonedDateTime, here is their doc: https://www.ibm.com/support/knowledgecenter/SSZJPZ_11.7.0/com.ibm.swg.im.iis.ds.parjob.dev.doc/topics/r_deeref_Date_and_Time_Functions.html.
We have also heard about use cases about these functions used in LocalDateTime. I feel these functions are more like the Datetime type in Bigquery.

wrangler-core/src/main/java/io/cdap/functions/DateAndTime.java
yaojiefeng: The DataStage date and time functions basically do not operate on a ZonedDateTime, here is their doc: https://www.ibm.com/support/knowledgecenter/SSZJPZ_11.7.0/com.ibm.swg.im.iis.ds.parjob.dev.doc/topics/r_deeref_Date_and_Time_Functions.html.
We have also heard about use cases about these functions used in LocalDateTime. I feel these functions are more like the Datetime type in Bigquery.

wrangler-core/src/main/java/io/cdap/functions/DateAndTime.java
albertshau: This is poor naming though, seeing how we have a 'timestamp' type that is an actual timestamp and not a date time. I think it's better to have a `ZonedDateTime CurrentTimestamp()` and a `LocalDateTime CurrentDateTime()` instead.

wrangler-core/src/main/java/io/cdap/functions/DateAndTime.java
```diff
@@ -0,0 +1,664 @@
...
+   * Return the {@link LocalTime} the string represents. If the string is null, this method will return null.
+   */
+  public static LocalTime GetTime(String time) {
+    return time == null ? null : LocalTime.parse(time, TIME_FORMAT);
+  }
+
+  /**
+   * Return the {@link LocalDateTime} the string represents. If the string is null, this method will return null.
+   */
+  public static LocalDateTime GetTimestamp(String date) {
```
albertshau: rename to GetDateTime()

wrangler-core/src/main/java/io/cdap/functions/DateAndTime.java
```diff
@@ -0,0 +1,664 @@
...
+   *
+   * @param hours component of {@link LocalTime}
+   * @param mins component of {@link LocalTime}
+   * @param seconds component of {@link LocalTime}
+   * @param microseconds component of {@link LocalTime}
+   * @return a instance of {@link LocalTime}
+   */
+  public static LocalTime TimeFromComponents(Integer hours, Integer mins, Integer seconds, Integer microseconds) {
+    return hours == null || mins == null || seconds == null || microseconds == null ? null :
+             LocalTime.of(hours, mins, seconds, microseconds * 1000);
```
albertshau: nit: indentation

wrangler-core/src/main/java/io/cdap/functions/DateAndTime.java
albertshau: let's rename to `GetDateTime(LocalDate date, LocalTime time)`

wrangler-core/src/main/java/io/cdap/functions/DateAndTime.java
```diff
@@ -0,0 +1,664 @@
...
+
+  /**
+   * Returns a timestamp that is derived from the number of seconds from the base timestamp object.
+   * If time is null, null will be returned. If seconds is null, time will be returned.
+   *
+   * @param seconds to be added to base timestamp.
+   * @param time base timestamp.
+   * @return timestamp with seconds added to base time.
+   */
+  public static LocalDateTime TimestampFromSecondsSince(Integer seconds, LocalDateTime time) {
```
albertshau: DateTimeFromSecondsSince

wrangler-core/src/main/java/io/cdap/functions/DateAndTime.java
albertshau: This is still weird to me, as it would be returning a LocalDateTime assuming it is UTC, but losing the fact that it is UTC. I don't see a TimestampFromEpoch function in the link you gave.

I think it makes more sense to have this return a ZonedDateTime, and have another `ToDateTime(ZonedDateTime timestamp, String timeZone)` function used to convert a real timestamp to a date time.

wrangler-core/src/main/java/io/cdap/functions/DateAndTime.java
```diff
@@ -0,0 +1,664 @@
...
+   * Returns a timestamp from the given time and timestamp objects.
+   * The value in the time object overwrites the time value in the timestamp object
+   * so that only the date part is used from the timestamp.
+   * If timestamp is null, null will be returned. If time is null, timestamp will be returned.
+   *
+   * @param time to replace in timestamp
+   * @param timestamp who's timestamp has to replaced.
+   * @return modified timestamp.
+   */
+  public static LocalDateTime TimestampFromTime(LocalTime time, LocalDateTime timestamp) {
```
albertshau: DateTimeFromTime()

wrangler-core/src/main/java/io/cdap/functions/DateAndTime.java
```diff
@@ -0,0 +1,664 @@
...
+                                                                    .withNano(time.getNano()));
+  }
+
+  /**
+   * Returns a timestamp from the given time and the offsets.
+   * If timestamp is null, null will be returned. If offset is null, it will treated as 0.
+   *
+   * @return Returns the timestamp, with offsets applied from the base timestamp.
+   */
+  public static LocalDateTime TimestampOffsetByComponents(LocalDateTime timestamp, Integer yearOffset,
```
albertshau: DateTimeOffsetByComponents()

wrangler-core/src/main/java/io/cdap/functions/DateAndTime.java
```diff
@@ -0,0 +1,664 @@
...
+                                        .plusHours(hourOffset == null ? 0 : hourOffset)
+                                        .plusMinutes(minOffset == null ? 0 : minOffset)
+                                        .plusSeconds(secondOffset == null ? 0 : secondOffset);
+  }
+
+  /**
+   * @return Returns the timestamp, with offsets applied from the base timestamp with seconds offset.
+   * If timestamp is null, null will be returned. If seconds is null, it will be treated as 0.
+   */
+  public static LocalDateTime TimestampOffsetBySeconds(LocalDateTime timestamp, Integer secondsOffset) {
```
albertshau: DateTimeOffsetBySeconds

wrangler-core/src/main/java/io/cdap/functions/DateAndTime.java
```diff
@@ -0,0 +1,664 @@
...
+    return timestamp == null ? null : timestamp.plusSeconds(secondsOffset == null ? 0 : secondsOffset);
+  }
+
+  /**
+   * @return Returns a UNIX epoch time from the given timestamp.
+   * If timestamp is null, null will be returned.
+   */
+  public static Long EpochFromTimestamp(LocalDateTime timestamp) {
```
albertshau: EpochFromDateTime

wrangler-core/src/main/java/io/cdap/functions/DateAndTime.java
```diff
@@ -0,0 +1,664 @@
...
+  /**
+   * @return Returns the timestamp, with offsets applied from the base timestamp with seconds offset.
+   * If timestamp is null, null will be returned. If seconds is null, it will be treated as 0.
+   */
+  public static LocalDateTime TimestampOffsetBySeconds(LocalDateTime timestamp, Integer secondsOffset) {
+    return timestamp == null ? null : timestamp.plusSeconds(secondsOffset == null ? 0 : secondsOffset);
+  }
+
+  /**
+   * @return Returns a UNIX epoch time from the given timestamp.
```
albertshau: doc needs to mention that this assumes the input is a datetime in UTC

wrangler-core/src/main/java/io/cdap/functions/DateAndTime.java
albertshau: Actually, it's called TimestampFromTimet in the doc you shared. Can have a DateTimeFromEpoch(), but it should be documented that it will be a DateTime in UTC.

wrangler-core/src/main/java/io/cdap/functions/DateAndTime.java
```diff
@@ -527,89 +535,90 @@ public static LocalTime TimeOffsetByComponents(LocalTime time, Integer hour, Int
...
+    return datetime == null ? null : datetime.plusSeconds(secondsOffset == null ? 0 : secondsOffset);
   }
 
   /**
-   * @return Returns a UNIX epoch time from the given timestamp.
-   * If timestamp is null, null will be returned.
+   * @return Returns a UNIX epoch time from the given datetime in UTC time.
+   * If datetime is null, null will be returned.
    */
-  public static Long EpochFromTimestamp(LocalDateTime timestamp) {
```
albertshau: for symmetry, also have an EpochFromTimestamp() that takes a ZonedDateTime.

albertshau (approved):

PR #463, opened 2020-12-15
PLUGIN-464 fix flatten to be a no-op on empty lists

albertshau:
Fixed a bug where input rows with an empty list would get
filtered out by the flatten directive.


albertshau:
cherry-pick of #462

elfenheart (approved):

supriyapatil-google (approved):
Thanks!

PR #466, opened 2021-02-03
Support datetime datatype and datetime directives in Wrangler

greeshmaswaminathan:
- Changes for supporting datetime datatype in Wrangler
New directives
1. parse-as-datetime  - Parses string as datetime with given format
2. format-datetime - Formats datetime to a string of given format
3. current-datetime - Generates current datetime with given zone
4. timestamp-to-datetime - Converts a timestamp to datetime
5. timestamp - Creates a timestamp from datetime and given zone

### DIFF
--- a/wrangler-core/src/main/java/io/cdap/directives/datetime/CurrentDateTime.java
+++ b/wrangler-core/src/main/java/io/cdap/directives/datetime/CurrentDateTime.java
@@ -1,0 +1,101 @@
+/*
+ *  Copyright © 2021 Cask Data, Inc.
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ *  use this file except in compliance with the License. You may obtain a copy of
+ *  the License at
+ *
+ *  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ *  WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ *  License for the specific language governing permissions and limitations under
+ *  the License.
+ */
+package io.cdap.directives.datetime;
+
+import io.cdap.cdap.api.annotation.Description;
+import io.cdap.cdap.api.annotation.Name;
+import io.cdap.cdap.api.annotation.Plugin;
+import io.cdap.wrangler.api.Arguments;
+import io.cdap.wrangler.api.Directive;
+import io.cdap.wrangler.api.DirectiveParseException;
+import io.cdap.wrangler.api.ExecutorContext;
+import io.cdap.wrangler.api.Optional;
+import io.cdap.wrangler.api.Row;
+import io.cdap.wrangler.api.annotations.Categories;
+import io.cdap.wrangler.api.lineage.Lineage;
+import io.cdap.wrangler.api.lineage.Mutation;
+import io.cdap.wrangler.api.parser.ColumnName;
+import io.cdap.wrangler.api.parser.TokenType;
+import io.cdap.wrangler.api.parser.UsageDefinition;
+
+import java.time.LocalDateTime;
+import java.time.ZoneId;
+import java.time.zone.ZoneRulesException;
+import java.util.List;
+
+/**
+ * Directive for generating current datetime with the specified zone
+ */
+@Plugin(type = Directive.TYPE)
+@Name("current-datetime")
+@Categories(categories = {"datetime"})
+@Description("Generates current datetime using the given zone")
+public class CurrentDateTime implements Directive, Lineage {
+
+  public static final String NAME = "current-datetime";
+  private static final String COLUMN = "column";
+  private static final String ZONE = "timezone";
+  private static final String UTC = "UTC";
+  private String column;
+  private String zone;
+  private ZoneId zoneId;
+
+  @Override
+  public UsageDefinition define() {
+    UsageDefinition.Builder builder = UsageDefinition.builder(NAME);
+    builder.define(COLUMN, TokenType.COLUMN_NAME);
+    builder.define(ZONE, TokenType.TEXT, Optional.TRUE);
+    return builder.build();
+  }
+
+  @Override
+  public void initialize(Arguments args) throws DirectiveParseException {
+    this.column = ((ColumnName) args.value(COLUMN)).value();
+    if (args.value(ZONE) == null) {
+      this.zone = UTC;
+      this.zoneId = ZoneId.of(UTC);
+      return;
+    }
+
+    this.zone = args.value(ZONE).value().toString();
+    try {
+      this.zoneId = ZoneId.of(this.zone);
+    } catch (IllegalArgumentException | ZoneRulesException exception) {
+      throw new DirectiveParseException(NAME, String.format("Zone '%s' is invalid.", this.zone), exception);
+    }
+  }
+
+  @Override
+  public List<Row> execute(List<Row> rows, ExecutorContext context) {
+    for (Row row : rows) {
+      row.addOrSet(column, LocalDateTime.now(zoneId));
+    }
+    return rows;
+  }
+
+  @Override
+  public void destroy() {
+    //no op
+  }
+
+  @Override
+  public Mutation lineage() {
+    return Mutation.builder()
+      .readable("Generated current datetime for column '%s' with zone '%s'", column, zone)
+      .relation(column, column)
+      .build();
+  }
+}
--- a/wrangler-core/src/main/java/io/cdap/directives/datetime/DateTimeToTimeStamp.java
+++ b/wrangler-core/src/main/java/io/cdap/directives/datetime/DateTimeToTimeStamp.java
@@ -1,0 +1,117 @@
+/*
+ *  Copyright © 2021 Cask Data, Inc.
+ *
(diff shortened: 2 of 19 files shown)


wrangler-core/src/main/java/io/cdap/directives/datetime/CurrentDateTime.java
```diff
@@ -0,0 +1,107 @@
...
+                                        String.format("Zone '%s' is invalid.", this.zone),
+                                        exception);
+    }
+  }
+
+  @Override
+  public List<Row> execute(List<Row> rows, ExecutorContext context) {
+    for (Row row : rows) {
+      int idx = row.find(column);
+      if (idx == -1) {
```
albertshau: seems like this should add the column if it doesn't exist?

wrangler-core/src/main/java/io/cdap/directives/datetime/DateTimeToTimeStamp.java
```diff
@@ -0,0 +1,116 @@
...
+import java.time.ZoneId;
+import java.time.ZonedDateTime;
+import java.time.zone.ZoneRulesException;
+import java.util.List;
+
+/**
+ * Directive for converting a datetime column to timestamp with the specified zone
+ */
+@Plugin(type = Directive.TYPE)
+@Name("timestamp")
```
albertshau: seems like a weird directive name? Should it be something like datetime-to-timestamp since there is one for timestamp-to-datetime?

wrangler-core/src/main/java/io/cdap/directives/datetime/DateTimeToTimeStamp.java
```diff
@@ -0,0 +1,116 @@
...
+        continue;
+      }
+      Object value = row.getValue(idx);
+      // If the data in the cell is null, then skip this row.
+      if (value == null) {
+        continue;
+      }
+
+      if (!(value instanceof LocalDateTime)) {
+        throw new ErrorRowException(NAME, String.format("Column %s is not of DateTime type", value.toString()), 2);
```
albertshau: Is this supposed to be the column name instead of the value?

Also would be good to mention what the type actually is, something like:

Expected column x to be a DateTime but found a y

wrangler-core/src/main/java/io/cdap/directives/datetime/FormatDateTime.java
```diff
@@ -0,0 +1,116 @@
...
+        continue;
+      }
+      Object value = row.getValue(idx);
+      // If the data in the cell is null, then skip this row.
+      if (value == null) {
+        continue;
+      }
+
+      if (!(value instanceof LocalDateTime)) {
+        throw new ErrorRowException(NAME, String.format("Column %s is not of DateTime type", value.toString()), 2);
```
albertshau: Is this supposed to be the column name instead of the value?

Also would be good to mention what the type actually is, something like:

Expected column x to be a DateTime but found a y

wrangler-core/src/main/java/io/cdap/directives/datetime/FormatDateTime.java
```diff
@@ -0,0 +1,116 @@
...
+      }
+
+      try {
+        final LocalDateTime localDateTime = (LocalDateTime) value;
```
albertshau: don't need the final

wrangler-core/src/main/java/io/cdap/directives/datetime/FormatDateTime.java
```diff
@@ -0,0 +1,116 @@
...
+        row.setValue(idx, localDateTime.format(formatter));
+      } catch (DateTimeException exception) {
+        throw new ErrorRowException(NAME, String.format("Error during converting datetime %s to string with format %s",
```
albertshau: include the cause, otherwise it will be hard to debug.

`throw new ErrorRowException(..., exception)`

wrangler-core/src/main/java/io/cdap/directives/parser/ParseDateTime.java
```diff
@@ -0,0 +1,111 @@
...
+        continue;
+      }
+      Object value = row.getValue(idx);
+      // If the data in the cell is null or is already Datetime , then skip this row.
+      if (value == null || value instanceof LocalDateTime) {
+        continue;
+      }
+
+      try {
+        final LocalDateTime localDateTime = LocalDateTime.parse(value.toString(), formatter);
```
albertshau: don't need the final

wrangler-core/src/main/java/io/cdap/directives/parser/ParseDateTime.java
```diff
@@ -0,0 +1,111 @@
...
+        row.setValue(idx, localDateTime);
+      } catch (DateTimeParseException exception) {
+        throw new ErrorRowException(NAME, String.format("Value %s not in the format %s", value.toString(), format), 2);
```
albertshau: include cause in the wrapped exception

albertshau: also good to include the column name in the error message

wrangler-core/src/main/java/io/cdap/wrangler/utils/RecordConvertor.java
```diff
@@ -91,6 +92,8 @@ public StructuredRecord decodeRecord(Row row, Schema schema) throws RecordConver
           builder.setTimestamp(name, (ZonedDateTime) decodedObj);
         } else if (decodedObj instanceof BigDecimal) {
           builder.setDecimal(name, (BigDecimal) decodedObj);
+        } else if (decodedObj instanceof LocalDateTime) {
```
albertshau, line 95: is there corresponding logic somewhere when converting StructuredRecord to a Row to create a LocalDateTime instead of the underlying string?

greeshmaswaminathan: wrangler-transform/src/main/java/io/cdap/wrangler/Wrangler.java getValue() method returns LocalDateTime to set to Row (In this PR). Applicable directives will also be setting LocalDateTime to Row.

wrangler-core/src/main/java/io/cdap/directives/datetime/FormatDateTime.java
greeshmaswaminathan: Will modify ErrorRowException to take cause.

wrangler-core/src/main/java/io/cdap/directives/datetime/DateTimeToTimeStamp.java
greeshmaswaminathan: The idea was to have a constructor style directive - Timestamp(datetime_expr, [timezone]) , but agree that datetime-to-timestamp may be better to avoid confusion

wrangler-core/src/main/java/io/cdap/directives/datetime/CurrentDateTime.java
greeshmaswaminathan: Done

wrangler-core/src/main/java/io/cdap/directives/datetime/DateTimeToTimeStamp.java
greeshmaswaminathan: changed to datetime-to-timestamp

wrangler-core/src/main/java/io/cdap/directives/datetime/FormatDateTime.java
greeshmaswaminathan: Including both column name and value in all such errors.

wrangler-core/src/main/java/io/cdap/directives/datetime/FormatDateTime.java
greeshmaswaminathan: Done

wrangler-core/src/main/java/io/cdap/directives/datetime/FormatDateTime.java
greeshmaswaminathan: Modified and logging ErrorRowException in RecipePipelineExecutor

wrangler-core/src/main/java/io/cdap/directives/parser/ParseDateTime.java
greeshmaswaminathan: Done

wrangler-core/src/main/java/io/cdap/directives/parser/ParseDateTime.java
greeshmaswaminathan: Done

wrangler-core/src/main/java/io/cdap/directives/datetime/DateTimeToTimeStamp.java
greeshmaswaminathan: Done

wrangler-core/src/main/java/io/cdap/directives/datetime/FormatDateTime.java
albertshau: nit: Error during converting -> Error converting

albertshau (approved):
in the future, please wait for an approval before squashing commits so that it is easier to review just the changes.

PR #467, opened 2021-02-09
CDAP-15979 do not generate bad reference names

yaojiefeng:


wrangler-core/src/main/java/io/cdap/wrangler/utils/ReferenceNameUtils.java
```diff
@@ -0,0 +1,56 @@
+/*
+ * Copyright © 2020 Cask Data, Inc.
```
albertshau: 2021

wrangler-core/src/main/java/io/cdap/wrangler/utils/ReferenceNameUtils.java
```diff
@@ -0,0 +1,56 @@
...
+ */
+
+package io.cdap.wrangler.utils;
+
+import java.util.regex.Pattern;
+
+/**
+ * Utility methods for reference name
+ */
+public class ReferenceNameUtils {
```
albertshau: nit: our utility classes don't normally have 'Util' in the name. This would be ReferenceNames

wrangler-core/src/main/java/io/cdap/wrangler/utils/ReferenceNameUtils.java
```diff
@@ -0,0 +1,56 @@
...
+
+  /**
+   * Cleanse the given reference name. This method will remove all the disallowed characters in the given reference
+   * name. For example, 111-22-33(1).csv will get convert to 111-22-331.csv. If no valid characters, the method will
+   * return "sample".
+   *
+   * @param referenceName the old reference name
+   * @return the reference name with only allowed characters
+   */
+  public static String cleanseReferenceName(String referenceName) {
```
albertshau: this can still return a name that is too long, should substring it if it's over max length

albertshau (approved):
minor comments

wrangler-core/src/main/java/io/cdap/wrangler/utils/ReferenceNameUtils.java
yaojiefeng: Pipeline deployed with 2M+ length reference name, so don't think there is a limit

PR #469, opened 2021-02-09
CDAO-15509 fix logical type

yaojiefeng:
For now only change the response for the logical type, for other type it still return the corresponding java type to preserve compatibility

### DIFF
--- a/wrangler-service/src/main/java/io/cdap/wrangler/service/directive/DirectivesHandler.java
+++ b/wrangler-service/src/main/java/io/cdap/wrangler/service/directive/DirectivesHandler.java
@@ -96,12 +96,11 @@ import io.cdap.wrangler.registry.UserDirectiveRegistry;
 import io.cdap.wrangler.service.common.AbstractWranglerHandler;
 import io.cdap.wrangler.statistics.BasicStatistics;
 import io.cdap.wrangler.statistics.Statistics;
-import io.cdap.wrangler.utils.Json2Schema;
 import io.cdap.wrangler.utils.ObjectSerDe;
+import io.cdap.wrangler.utils.SchemaConverter;
 import io.cdap.wrangler.validator.ColumnNameValidator;
 import io.cdap.wrangler.validator.Validator;
 import io.cdap.wrangler.validator.ValidatorException;
-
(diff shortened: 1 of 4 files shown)


wrangler-core/src/main/java/io/cdap/wrangler/utils/SchemaConvertor.java
```diff
@@ -48,18 +48,9 @@
...
+ * This class provides functionality to retrieve  {@link Schema} from a row or an object.
  */
-public final class Json2Schema {
-  // Parser to parse json if provided as string.
-  private final JsonParser parser;
-
-  public Json2Schema() {
-    this.parser = new JsonParser();
-  }
+public final class SchemaConvertor {
```
albertshau: nit: Convertor -> Converter

wrangler-service/src/main/java/io/cdap/wrangler/service/directive/DirectivesHandler.java
```diff
@@ -625,7 +625,14 @@ public void execute(HttpServiceRequest request, HttpServiceResponder responder,
             Object object = field.getSecond();
 
             if (object != null) {
-              types.put(fieldName, object.getClass().getSimpleName());
+              Schema schema = convertor.getSchema(object, fieldName);
+              String type = object.getClass().getSimpleName();
+              if (schema != null) {
+                schema = schema.isNullable() ? schema.getNonNullable() : schema;
+                // here we only change the type if this is a logical type to preserve backward compatibility
+                type = schema.getLogicalType() == null ? type : schema.getLogicalType().name();
```
albertshau: better to use schema.getType().name() instead of the classname. Arrays, maps, and nested records will work better that way.

albertshau: would be good to check with @bdmogal on the casing here. This would give it in all caps, which might not be desired.

yaojiefeng: I made the change so the first character is upper case, the rest is lower case

albertshau (approved):

PR #470, opened 2021-02-09
CDAP-16316 fix case insensitive find

yaojiefeng:
This is pretty risky and backward incompatible changes. 
But it is in need IMO because:
1. Our CDAP schema is case sensitive, so the row should be consistent with it

Therefore, it is good we make this change to make everything consistent.

### DIFF
--- a/wrangler-core/src/main/java/io/cdap/wrangler/utils/ColumnConverter.java
+++ b/wrangler-core/src/main/java/io/cdap/wrangler/utils/ColumnConverter.java
@@ -28,26 +28,33 @@ public final class ColumnConverter {
   }
 
   /**
-   * Renames a column.
+   * Renames a column. The renamed column must not exist in the row. The comparision is case insensitive.
    *
    * @param row source record to be modified.
    * @param column name of the column within source record.
    * @param toName the target name of the column.
-   * @throws DirectiveExecutionException when a column matching the target name already exists.
+   * @throws DirectiveExecutionException when a column matching the target name already exists or the given column
+   * does not exist
    */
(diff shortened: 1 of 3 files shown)


yaojiefeng:
Since this change will for sure break things for user upgrades the pipeline, will just make it special for rename directive since rename doesn't allow it in the first place so it will be safe

wrangler-core/src/main/java/io/cdap/wrangler/utils/ColumnConverter.java
```diff
@@ -28,26 +28,47 @@ private ColumnConverter() {
...
+   * @param row source record to be modified.
+   * @param column name of the column within source record.
+   * @param toName the target name of the column.
+   * @param exactMatch true for case sensitive col name, false for case insensitive
+   * @throws DirectiveExecutionException when a column matching the target name already exists or the given column
+   * does not exist
+   */
+  public static void rename(String directiveName, Row row, String column, String toName, boolean exactMatch)
+    throws DirectiveExecutionException {
+    int idx = exactMatch ? row.findExact(column) : row.find(column);
```
albertshau: this is also backwards incompatible. The checks should all be case insensitive, with the exception of if before and after are just different casing.

albertshau: In other words, if there is a row <a:0, b:1>

rename A a2 // should not fail, result in <a2:0, b:1>
rename a B // should behave the same (failure)
rename a A // should no longer fail, result in <A:0, b:1>

albertshau (approved):

PR #477, opened 2021-03-10
Validation and docs for datetime

greeshmaswaminathan:
- Added validation while setting a column value to datetime
- Added documentation for all new datetime directives


fernst (approved):
LGTM

PR #481, opened 2021-03-11
CDAP-17736 - Set-type directive fixes for datetime

greeshmaswaminathan:
Set-type directive fixes for datetime
- Allow datetime type to be converted to string
- Throw exception for incompatible types

### DIFF
--- a/wrangler-core/src/main/java/io/cdap/wrangler/utils/ColumnConverter.java
+++ b/wrangler-core/src/main/java/io/cdap/wrangler/utils/ColumnConverter.java
@@ -18,9 +18,7 @@ package io.cdap.wrangler.utils;
 import io.cdap.cdap.api.common.Bytes;
 import io.cdap.wrangler.api.DirectiveExecutionException;
 import io.cdap.wrangler.api.Row;
-
 import java.math.BigDecimal;
-import java.math.BigInteger;
 
 /**
  * Utility class that converts a {@link Row} column into another column.
@@ -112,7 +110,7 @@ public final class ColumnConverter {
         } else if (object instanceof byte[]) {
           return Bytes.toInt((byte[]) object);
         } else {
-          return object;
+          break;
         }
       }
 
@@ -135,7 +133,7 @@ public final class ColumnConverter {
         } else if (object instanceof byte[]) {
           return Bytes.toShort((byte[]) object);
         } else {
-          return object;
+          break;
         }
       }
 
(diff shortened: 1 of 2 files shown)


wrangler-core/src/main/java/io/cdap/wrangler/utils/ColumnConverter.java
```diff
@@ -203,8 +205,10 @@ private static Object convertType(String col, String toType, Object object) thro
           return Bytes.toString((byte[]) object);
         } else if (object instanceof Boolean) {
           return Boolean.toString((Boolean) object);
+        } else if (object instanceof LocalDateTime) {
```
albertshau: while we're doing this, should also add one for timestamps (ZonedDateTime)

wrangler-core/src/main/java/io/cdap/wrangler/utils/ColumnConverter.java
```diff
@@ -203,8 +205,10 @@ private static Object convertType(String col, String toType, Object object) thro
...
+          return ((LocalDateTime) object).format(DateTimeFormatter.ISO_DATE_TIME);
         } else {
-          return object;
+          break;
```
albertshau: I wonder if string should just return object.toString()? I think we would be able to get rid of a bunch of these instanceof checks, unless Double.toString() behaves differently?

Also, I don't see a null check anywhere, is null checked outside of this method? Or does it throw an exception?

wrangler-core/src/main/java/io/cdap/wrangler/utils/ColumnConverter.java
greeshmaswaminathan: Is a toString() enough on ZonedDateTime since we don't have a specified format?

wrangler-core/src/main/java/io/cdap/wrangler/utils/ColumnConverter.java
greeshmaswaminathan: Null check is done before calling this. Yes, I think a toString() should be fine except for byte[].  Double and other wrappers toString() calls the static toString(), so it should be fine. For Datetime toString() gives a ISO-8601 format. So will allow all types to be converted to string.

greeshmaswaminathan: Done.

albertshau (approved):

PR #485, opened 2021-03-22
Revert to error collection for parse-as-datetime directive.

greeshmaswaminathan:
Reverting change in https://github.com/data-integrations/wrangler/pull/483


avirmani (approved):
LGTM

PR #489, opened 2021-04-12
[CDAP-17693] Introduce spark 3 for tests, drop spark 1, create spark …

tivv:
…3 runtime for cdap sandbox and standalone

### DIFF
--- a/wrangler-transform/pom.xml
+++ b/wrangler-transform/pom.xml
@@ -103,8 +103,8 @@
         <version>1.1.0</version>
         <configuration>
           <cdapArtifacts>
-            <parent>system:cdap-data-pipeline[6.4.0-SNAPSHOT,7.0.0-SNAPSHOT)</parent>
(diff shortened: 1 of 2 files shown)


wrangler-transform/pom.xml
```diff
@@ -103,8 +103,8 @@
         <version>1.1.0</version>
         <configuration>
           <cdapArtifacts>
-            <parent>system:cdap-data-pipeline[6.4.0-SNAPSHOT,7.0.0-SNAPSHOT)</parent>
-            <parent>system:cdap-data-streams[6.4.0-SNAPSHOT,7.0.0-SNAPSHOT)</parent>
+            <parent>system:cdap-data-pipeline2_2.11[6.5.0-SNAPSHOT,7.0.0-SNAPSHOT)</parent>
```
albertshau: don't need to change the artifact name.

Also, spark doesn't look to be used anywhere in wrangler, can leave this PR out so the jira history doesn't have an unrelated PR or git commit.

PR #492, opened 2021-05-17
CDAP-17995 change wrangler on error default

yaojiefeng:
Change the wrangler default back to fail pipeline. Make backend and UI defaults to the same.
Adds a hidden flag to preserve 6.3 compatibility in case any customer wants to keep 6.3 defaults, this flag will be hidden in UI.

Tried to add a unit test https://github.com/data-integrations/wrangler/blob/CDAP-17995-change-wrangler-defaults/wrangler-transform/src/test/java/io/cdap/wrangler/WranglerTest.java.
But due to https://cdap.atlassian.net/browse/CDAP-11465, wrangler uses guava 20.0, which is a higher version than the platform, the unit test cannot work.

### DIFF
--- a/wrangler-transform/src/main/java/io/cdap/wrangler/Wrangler.java
+++ b/wrangler-transform/src/main/java/io/cdap/wrangler/Wrangler.java
@@ -90,7 +90,8 @@ public class Wrangler extends Transform<StructuredRecord, StructuredRecord> {
   private static final String APPLICATION_NAME = "dataprep";
   private static final String SERVICE_NAME = "service";
   private static final String CONFIG_METHOD = "config";
-  private static final String ON_ERROR_DEFAULT = "send-to-error-port";
+  private static final String ON_ERROR_DEFAULT = "fail-pipeline";
+  private static final String ERROR_STRATEGY_DEFAULT = "wrangler.error.strategy.default";
 
   // Plugin configuration.
   private final Config config;
@@ -112,6 +113,9 @@ public class Wrangler extends Transform<StructuredRecord, StructuredRecord> {
 
   // Directive registry.
   private DirectiveRegistry registry;
+
+  // on error strategy
+  private String onErrorStrategy;
 
   // This is used only for tests, otherwise this is being injected by the ingestion framework.
   public Wrangler(Config config) {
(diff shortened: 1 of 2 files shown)


wrangler-transform/src/main/java/io/cdap/wrangler/Wrangler.java
```diff
@@ -555,21 +556,31 @@ private RecipeParser getRecipeParser(StageContext context)
     @Nullable
     private final String onError;
 
+    // this is a hidden property which preserves some sort of 6.3 backward compatibility,
+    // if any user wants the "send-to-error-port" default, they can use this flag, but it won't be shown in UI.
+    @Nullable
+    private final Boolean keepOldErrorDefault;
```
fernst: Would it be better to make this an instance-level configuration? This forces users to always have to modify existing pipelines in order to enable this behavior.

tivv: I agree. We should do it from runtime properties. Having it as a config does not solve a problem as user have to change every pipeline. If they go change every pipeline, they can set onError  directly. And runtime properly is something they can set on namespace/instance level

tivv: Also, does it syncs up the UI? We need to ensure backend default and UI default are in sync

yaojiefeng: Added the config as instance level. 
It won't sync with UI, the UI default is inside the widget file, like this, https://github.com/data-integrations/wrangler/pull/492/files#diff-6b579e1ba0e29fd41b29cceb044c8227f2c4491be4bedf3410b4945fd3425055R58

wrangler-transform/src/main/java/io/cdap/wrangler/Wrangler.java
```diff
@@ -320,6 +325,9 @@ public void initialize(TransformContext context) throws Exception {
       );
     }
 
+    boolean keepOld = Boolean.valueOf(context.getArguments().get(KEEP_OLD_DEFAULT_KEY));
```
fernst: nit: I know this probably doesn't cause issues, but I would like to have an explicit null check here before invoking the valueOf method.

fernst (approved):
LGTM

wrangler-transform/src/main/java/io/cdap/wrangler/Wrangler.java
tivv: Can we have string `wrangler.error.strategy.default` instead? It would be more flexible and look as a feature, not as a patch. This way one can set any default he/she wants, not "old" or "new"

tivv (approved):

PR #499, opened 2021-05-26
new workspace store

yaojiefeng:
This pr adds a new workspace store for the workspace.

PR #500, opened 2021-05-26
Add new handler and deprecate all the old one

yaojiefeng:
This is depending on #499 and https://github.com/cdapio/cdap/pull/13430

This pr has quite a lot of changes but lots of code cannot be unit tested due to https://cdap.atlassian.net/browse/CDAP-11465.

Basically it has following major changes:
1. Deprecated most of the unused endpoints and dataset
2. Introduce a new set of v2 endpoints for workspace to make the code cleaner, for some of the endpoints still preserve the old logic as these logic would need a lot of effort to clean up.

All the changes have been tested manually

### DIFF
--- a/wrangler-core/src/main/java/io/cdap/wrangler/registry/CompositeDirectiveRegistry.java
+++ b/wrangler-core/src/main/java/io/cdap/wrangler/registry/CompositeDirectiveRegistry.java
@@ -17,6 +17,7 @@
 package io.cdap.wrangler.registry;
 
 import com.google.common.collect.Iterables;
+import io.cdap.cdap.api.artifact.ArtifactSummary;
 import io.cdap.wrangler.api.DirectiveLoadException;
 
 import java.io.IOException;
@@ -66,6 +67,24 @@ public final class CompositeDirectiveRegistry implements DirectiveRegistry {
     for (int idx = 0; idx < registries.length; ++idx) {
       registries[idx].reload(namespace);
     }
+  }
+
+  @Nullable
+  @Override
+  public ArtifactSummary getLatestWranglerArtifact() {
+    ArtifactSummary latestArtifact = null;
+    for (DirectiveRegistry registry : registries) {
+      ArtifactSummary artifact = registry.getLatestWranglerArtifact();
+      if (artifact == null) {
+        continue;
+      }
+      if (latestArtifact == null) {
+        latestArtifact = artifact;
+      } else {
+        latestArtifact = DirectiveRegistry.pickLatest(latestArtifact, artifact);
+      }
+    }
+    return latestArtifact;
   }
 
   /**
--- a/wrangler-core/src/main/java/io/cdap/wrangler/registry/UserDirectiveRegistry.java
+++ b/wrangler-core/src/main/java/io/cdap/wrangler/registry/UserDirectiveRegistry.java
@@ -20,12 +20,14 @@ import com.google.common.collect.MapDifference;
 import com.google.common.collect.Maps;
 import io.cdap.cdap.api.artifact.ArtifactInfo;
 import io.cdap.cdap.api.artifact.ArtifactManager;
+import io.cdap.cdap.api.artifact.ArtifactSummary;
 import io.cdap.cdap.api.artifact.CloseableClassLoader;
 import io.cdap.cdap.api.plugin.PluginClass;
 import io.cdap.cdap.api.plugin.PluginConfigurer;
 import io.cdap.cdap.api.plugin.PluginProperties;
 import io.cdap.cdap.api.service.http.HttpServiceContext;
 import io.cdap.cdap.etl.api.StageContext;
+import io.cdap.cdap.etl.api.Transform;
 import io.cdap.wrangler.api.Directive;
 import io.cdap.wrangler.api.DirectiveLoadException;
 
@@ -63,10 +65,13 @@ import javax.annotation.Nullable;
  * @see CompositeDirectiveRegistry
  */
 public final class UserDirectiveRegistry implements DirectiveRegistry {
+  private static final String WRANGLER_TRANSFORM = "wrangler-transform";
+  private static final String WRANGLER_PLUGIN = "Wrangler";
   private final Map<String, Map<String, DirectiveInfo>> registry = new ConcurrentSkipListMap<>();
   private final List<CloseableClassLoader> classLoaders = new ArrayList<>();
   private StageContext context = null;
   private HttpServiceContext manager = null;
+  private ArtifactSummary wranglerArtifact;
 
   /**
    * This constructor should be used when initializing the registry from <tt>Service</tt>.
@@ -156,7 +161,9 @@ public final class UserDirectiveRegistry implements DirectiveRegistry {
     if (manager != null) {
       try {
         List<ArtifactInfo> artifacts = manager.listArtifacts(namespace);
+        ArtifactSummary latestWrangler = null;
         for (ArtifactInfo artifact : artifacts) {
+          boolean isWranglerArtifact = artifact.getName().equalsIgnoreCase(WRANGLER_TRANSFORM);
           Set<PluginClass> plugins = artifact.getClasses().getPlugins();
           for (PluginClass plugin : plugins) {
             if (Directive.TYPE.equalsIgnoreCase(plugin.getType())) {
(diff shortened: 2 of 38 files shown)


wrangler-service/src/main/java/io/cdap/wrangler/service/directive/AbstractDirectiveHandler.java
```diff
@@ -0,0 +1,268 @@
...
+import java.util.Map;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+/**
+ * Abstract handler which contains common logic for v1 and v2 endpoints
+ *
+ * TODO: CDAP-18015 Refactor and add unit test for methods in this class
+ */
+public class AbstractDirectiveHandler extends AbstractWranglerHandler {
```
albertshau, line 72: I'm assuming the code here is just moved from existing places. Please point out changes to logic if there are any.

wrangler-service/src/main/java/io/cdap/wrangler/service/directive/ConnectionDiscoverer.java
```diff
@@ -0,0 +1,175 @@
...
+          TimeUnit.MILLISECONDS.sleep(delay);
+          delay = (long) (delay * (minMultiplier + Math.random() * (maxMultiplier - minMultiplier + 1)));
+          delay = Math.min(delay, RETRY_MAX_DELAY_MILLIS);
+        } catch (IOException e) {
+          throw new IOException(String.format("Failed to retrieve sample for connection '%s' in namespace '%s'.",
+                                              connectionName, namespace), e);
+        }
+      }
+    } catch (InterruptedException e) {
+      throw new RuntimeException(String.format("Thread interrupted while trying to retrieve sample for " +
```
albertshau: can this be propagated up instead of wrapped in a RuntimeException?

wrangler-service/src/main/java/io/cdap/wrangler/service/directive/ConnectionDiscoverer.java
```diff
@@ -0,0 +1,175 @@
...
+                                                 "connection '%s' in namespace '%s'.", connectionName, namespace), e);
+    }
+    throw new IllegalStateException(String.format("Timed out when trying to retrieve sample for " +
```
albertshau: this is not really an IllegalState, as it is a normal situation that could happen (like if the connection service is down). It would be better to throw the IOException that caused the latest failure.

wrangler-service/src/main/java/io/cdap/wrangler/service/directive/WorkspaceHandler.java
```diff
@@ -0,0 +1,353 @@
...
+  }
+
+  /**
+   * Get source specs, contains some hacky way on dealing with the csv parser
+   */
+  private Set<StageSpec> getSourceSpecs(WorkspaceDetail detail, List<String> directives) {
+    SampleSpec sampleSpec = detail.getWorkspace().getSampleSpec();
+    Set<StageSpec> srcSpecs = sampleSpec == null ? Collections.emptySet() : sampleSpec.getRelatedPlugins();
+
+    // really hacky way for the parse-as-csv directive, should get removed once we have support to provide the
```
albertshau, line 319: I believe there are other hacks in the handlers. I seem to remember sometimes the 'blob' format gets used for the source to read the entire content as a single record, maybe if the file is xml or avro or something like that?

wrangler-service/src/main/java/io/cdap/wrangler/service/directive/WorkspaceHandler.java
```diff
@@ -0,0 +1,353 @@
...
+      Schema schema = schemaConvertor.toSchema("record", createUberRecord(result));
+      Map<String, String> properties = ImmutableMap.of("directives", String.join("\n", directives));
+
+      Set<StageSpec> srcSpecs = getSourceSpecs(detail, directives);
+
+      responder.sendString(GSON.toJson(new WorkspaceSpec(
+        srcSpecs, new StageSpec(
+          schema, new Plugin("Wrangler", "transform", properties,
+                             // TODO: find a better way to get the scope and the artifact name
+                             new Artifact("wrangler-transform", ProjectInfo.getVersion(), "system"))))));
```
albertshau: hardcoding this isn't really ideal.

Looks like ArtifactManager.list() is already called (through DirectiveRegistry's reload() method when executing directives). Could we also keep track of the most recent Wrangler when that happens and use it here? Then we could also remove that ProjectInfo stuff

wrangler-service/src/main/java/io/cdap/wrangler/service/directive/AbstractDirectiveHandler.java
yaojiefeng: Yes, these are just code in old handlers, move them to better share the code

wrangler-service/src/main/java/io/cdap/wrangler/service/directive/WorkspaceHandler.java
yaojiefeng: Yes, I feel this hack should happen in the connector instead of here. When it comes to workspace stage, the sample should already have this logic applied, i.e, using blob for avro and excel file. Therefore, the generate spec should already have that.

wrangler-service/src/main/java/io/cdap/wrangler/service/directive/WorkspaceHandler.java
yaojiefeng: Moved the logic to UserDirectiveRegistry to get the wrangler artifact. I keep `ProjectInfo` since we have an /info endpoint which returns the version information, previously it reread the file every time the method is called, now it is in ProjectInfo so the file won't be read each time the endpoint is called

wrangler-core/src/main/java/io/cdap/wrangler/registry/CompositeDirectiveRegistry.java
```diff
@@ -68,6 +69,18 @@ public void reload(String namespace) throws DirectiveLoadException {
     }
   }
 
+  @Nullable
+  @Override
+  public Artifact getLatestWranglerArtifact() {
+    for (DirectiveRegistry registry : registries) {
+      Artifact artifact = registry.getLatestWranglerArtifact();
+      if (artifact != null) {
+        return artifact;
```
albertshau: though this happens to behave correctly now because System registry always returns null, this should return the latest one from all of the sub-registries instead of returning whichever registry was non-null first.

wrangler-core/src/main/java/io/cdap/wrangler/registry/UserDirectiveRegistry.java
```diff
@@ -156,7 +165,9 @@ public void reload(String namespace) throws DirectiveLoadException {
     if (manager != null) {
       try {
         List<ArtifactInfo> artifacts = manager.listArtifacts(namespace);
```
albertshau, line 163: does this return system artifacts too, or only user artifacts in the namespace?

If it doesn't return system artifacts we'll have to initial latest as the current ProjectInfo version.

wrangler-core/src/main/java/io/cdap/wrangler/registry/UserDirectiveRegistry.java
```diff
@@ -212,4 +243,28 @@ public void close() throws IOException {
...
+    if (cmp > 0) {
+      return artifact1;
+    }
+
+    if (cmp < 0) {
+      return artifact2;
+    }
+
+    // if scope is also same, then they are equal
+    if (artifact1.getScope().equals(artifact2.getScope())) {
```
albertshau: don't actually need this comparison, the next one covers this case.

wrangler-service/src/main/java/io/cdap/wrangler/service/directive/WorkspaceHandler.java
```diff
@@ -283,6 +282,8 @@ public void specification(HttpServiceRequest request, HttpServiceResponder respo
       if (ns.getName().equalsIgnoreCase(NamespaceId.SYSTEM.getNamespace())) {
         throw new BadRequestException("Getting specification in system namespace is currently not supported");
       }
+      // reload to retrieve the wrangler transform
+      composite.reload(namespace);
```
albertshau, line 289: this isn't already called in the executeDirectives() method? If not, I'm not sure how executeDirectives() works with UDDs.

wrangler-core/src/main/java/io/cdap/wrangler/registry/UserDirectiveRegistry.java
yaojiefeng: Yes, this one returns artifacts in that namespace and system namespace, from javadoc of ArtifactManager `Get the list of artifacts in the repository in the specified namespace and system namespace`. I have verified on sandbox it did return all the artifacts in system and user scope

wrangler-service/src/main/java/io/cdap/wrangler/service/directive/WorkspaceHandler.java
yaojiefeng: Good catch I missed this when migrating the endpoints, added it to the execute

yaojiefeng:
@albertshau addressed comments, please take one more look, thank you!

albertshau (approved):

PR #506, opened 2021-06-10
Enable remote execution for Wrangler directives

greeshmaswaminathan:
- Refactored to extract common code for directive execution
- Added code for remote execution
- Needs code in https://github.com/cdapio/cdap/pull/13450 to compile.

### DIFF
--- a/wrangler-core/pom.xml
+++ b/wrangler-core/pom.xml
@@ -295,6 +295,11 @@
       <artifactId>hapi-structures-v26</artifactId>
       <version>${hl7.version}</version>
     </dependency>
+    <dependency>
+      <groupId>io.cdap.cdap</groupId>
+      <artifactId>cdap-system-app-api</artifactId>
+      <version>${cdap.version}</version>
+    </dependency>
   </dependencies>
 
 
--- a/wrangler-core/src/main/java/io/cdap/wrangler/registry/UserDirectiveRegistry.java
+++ b/wrangler-core/src/main/java/io/cdap/wrangler/registry/UserDirectiveRegistry.java
@@ -26,6 +26,7 @@ import io.cdap.cdap.api.plugin.PluginClass;
 import io.cdap.cdap.api.plugin.PluginConfigurer;
 import io.cdap.cdap.api.plugin.PluginProperties;
 import io.cdap.cdap.api.service.http.HttpServiceContext;
+import io.cdap.cdap.api.service.worker.SystemAppTaskContext;
 import io.cdap.cdap.etl.api.StageContext;
 import io.cdap.cdap.etl.api.Transform;
 import io.cdap.wrangler.api.Directive;
@@ -72,6 +73,7 @@ public final class UserDirectiveRegistry implements DirectiveRegistry {
   private StageContext context = null;
   private HttpServiceContext manager = null;
   private ArtifactSummary wranglerArtifact;
+  private SystemAppTaskContext systemAppTaskContext;
 
   /**
    * This constructor should be used when initializing the registry from <tt>Service</tt>.
@@ -88,6 +90,14 @@ public final class UserDirectiveRegistry implements DirectiveRegistry {
    */
   public UserDirectiveRegistry(HttpServiceContext manager) throws DirectiveLoadException {
     this.manager = manager;
+  }
+
+  /**
+   * This constructor is used when creating from remote task
+   * @param systemAppTaskContext {@link SystemAppTaskContext}
+   */
+  public UserDirectiveRegistry(SystemAppTaskContext systemAppTaskContext) {
+   this.systemAppTaskContext = systemAppTaskContext;
   }
 
   /**
@@ -123,15 +133,7 @@ public final class UserDirectiveRegistry implements DirectiveRegistry {
   public DirectiveInfo get(String namespace, String name) throws DirectiveLoadException {
     Class<? extends Directive> directive = null;
     try {
-      if (context != null) {
-        directive = context.loadPluginClass(name);
-      } else {
-        if (manager != null) {
-          PluginConfigurer configurer = manager.createPluginConfigurer(namespace);
-          directive = configurer.usePluginClass(Directive.TYPE, name, UUID.randomUUID().toString(),
-                                                PluginProperties.builder().build());
-        }
-      }
+      directive = getDirective(namespace, name);
       if (directive == null) {
         throw new DirectiveLoadException(
           String.format("10-5 - Unable to load the user defined directive '%s'. " +
@@ -153,14 +155,26 @@ public final class UserDirectiveRegistry implements DirectiveRegistry {
     }
   }
 
+  @Nullable
+  private Class<? extends Directive> getDirective(String namespace, String name) throws IOException {
+    if (context != null) {
+      return context.loadPluginClass(name);
+    }
+    PluginConfigurer configurer = manager != null ?
+      manager.createPluginConfigurer(namespace) : systemAppTaskContext.createPluginConfigurer(namespace);
+    return configurer.usePluginClass(Directive.TYPE, name, UUID.randomUUID().toString(),
+                                     PluginProperties.builder().build());
+  }
+
   @Override
   public void reload(String namespace) throws DirectiveLoadException {
     Map<String, DirectiveInfo> newRegistry = new TreeMap<>();
     Map<String, DirectiveInfo> currentRegistry = registry.computeIfAbsent(namespace, k -> new TreeMap<>());
 
-    if (manager != null) {
+    ArtifactManager artifactManager = getArtifactManager();
+    if (artifactManager != null) {
       try {
-        List<ArtifactInfo> artifacts = manager.listArtifacts(namespace);
+        List<ArtifactInfo> artifacts = artifactManager.listArtifacts(namespace);
         ArtifactSummary latestWrangler = null;
         for (ArtifactInfo artifact : artifacts) {
           boolean isWranglerArtifact = artifact.getName().equalsIgnoreCase(WRANGLER_TRANSFORM);
@@ -168,7 +182,7 @@ public final class UserDirectiveRegistry implements DirectiveRegistry {
           for (PluginClass plugin : plugins) {
             if (Directive.TYPE.equalsIgnoreCase(plugin.getType())) {
               CloseableClassLoader closeableClassLoader
-                    = manager.createClassLoader(namespace, artifact, getClass().getClassLoader());
+                    = artifactManager.createClassLoader(namespace, artifact, getClass().getClassLoader());
               Class<? extends Directive> directive =
                 (Class<? extends Directive>) closeableClassLoader.loadClass(plugin.getClassName());
               DirectiveInfo classz = new DirectiveInfo(DirectiveInfo.Scope.USER, directive);
@@ -211,6 +225,10 @@ public final class UserDirectiveRegistry implements DirectiveRegistry {
         throw new DirectiveLoadException(e.getMessage(), e);
       }
     }
+  }
+
+  private ArtifactManager getArtifactManager() {
+    return manager != null ? manager : systemAppTaskContext.getArtifactManager();
   }
 
   @Nullable
--- a/wrangler-proto/src/main/java/io/cdap/wrangler/proto/workspace/v2/WorkspaceDetail.java
+++ b/wrangler-proto/src/main/java/io/cdap/wrangler/proto/workspace/v2/WorkspaceDetail.java
@@ -19,6 +19,7 @@ package io.cdap.wrangler.proto.workspace.v2;
 
 import io.cdap.wrangler.api.Row;
 
+import java.util.Arrays;
 import java.util.List;
 import java.util.Objects;
 
@@ -28,10 +29,12 @@ import java.util.Objects;
 public class WorkspaceDetail {
   private final Workspace workspace;
   private final List<Row> sample;
+  private final byte[] sampleAsBytes;
 
-  public WorkspaceDetail(Workspace workspace, List<Row> sample) {
+  public WorkspaceDetail(Workspace workspace, List<Row> sample, byte[] sampleAsBytes) {
     this.workspace = workspace;
     this.sample = sample;
+    this.sampleAsBytes = sampleAsBytes;
   }
 
   public Workspace getWorkspace() {
@@ -40,6 +43,10 @@ public class WorkspaceDetail {
 
   public List<Row> getSample() {
     return sample;
+  }
+
+  public byte[] getSampleAsBytes() {
+    return sampleAsBytes;
   }
 
   @Override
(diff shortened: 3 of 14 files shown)


wrangler-core/src/main/java/io/cdap/wrangler/registry/UserDirectiveRegistry.java
```diff
@@ -90,6 +92,16 @@ public UserDirectiveRegistry(HttpServiceContext manager) throws DirectiveLoadExc
     this.manager = manager;
   }
 
+  /**
+   * This constructor is used when creating from remote task
+   * @param pluginConfigurer {@link PluginConfigurer} for configuring directives
+   * @param artifactManager {@link ArtifactManager} for fetching artifacts for directives
+   */
+  public UserDirectiveRegistry(PluginConfigurer pluginConfigurer, ArtifactManager artifactManager) {
```
yaojiefeng: Adding this constructor will cause a lot of confusion, because either `HttpServiceContext` or `PluginConfiguer` and `ArtifactManager` will be null. I see this constructor is used when using `SystemAppTaskContext`. Our system service uses `SystemHttpServiceContext`, what is the difference between these two contexts, any difference on the plugin configurer and artifact manager they get created?

yaojiefeng: I feel it is better we can have a consistent way of loading artifacts and instantiate plugins

wrangler-service/src/main/java/io/cdap/wrangler/service/common/AbstractWranglerHandler.java
```diff
@@ -205,12 +207,22 @@ protected void respond(HttpServiceResponder responder, String namespace,
...
       responder.sendJson(HttpURLConnection.HTTP_INTERNAL_ERROR,
                          new io.cdap.wrangler.proto.workspace.v2.ServiceResponse<>((t.getMessage())));
     }
   }
 
+  private int getErrorCode(String exceptionName) {
+    boolean isBadRequest = Arrays
+      .asList(BadRequestException.class, ErrorRecordsException.class, JsonSyntaxException.class).
+        stream().anyMatch(e -> e.getName().equals(exceptionName));
+    return isBadRequest ? HttpURLConnection.HTTP_BAD_REQUEST : HttpURLConnection.HTTP_INTERNAL_ERROR;
```
yaojiefeng, line 223: I don't think this is the only bad code returned, I believe 404 is returned if a workspace is not found, it is better to extend this to support more error codes

wrangler-proto/src/main/java/io/cdap/wrangler/proto/workspace/v2/WorkspaceDetail.java
```diff
@@ -21,17 +21,27 @@
...
+import javax.annotation.Nullable;
 
 /**
  * Full information about the workspace, including sample data
  */
 public class WorkspaceDetail {
   private final Workspace workspace;
   private final List<Row> sample;
+  @Nullable
+  private final byte[] sampleAsBytes;
```
yaojiefeng, line 32: better to not have this `Nullable`, you can move the logic to convert sample to byte here so the byte[] always contain same thing as sample - https://github.com/data-integrations/wrangler/blob/develop/wrangler-storage/src/main/java/io/cdap/wrangler/store/workspace/WorkspaceStore.java#L217-L222

wrangler-service/src/main/java/io/cdap/wrangler/service/directive/CommonExecutor.java
```diff
@@ -0,0 +1,128 @@
...
+import java.io.ObjectOutputStream;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.function.Function;
+import java.util.stream.Collectors;
+
+/**
+ * Common code for executing directives
+ */
+public class CommonExecutor {
```
yaojiefeng: Call this `CommonDirectiveExecutor`

wrangler-service/src/main/java/io/cdap/wrangler/service/directive/WorkspaceHandler.java
```diff
@@ -253,6 +261,25 @@ public void execute(HttpServiceRequest request, HttpServiceResponder responder,
...
+                                   WorkspaceDetail detail) throws DirectiveLoadException, DirectiveParseException {
+    // load the udd
+    composite.reload(namespace);
+    return executeDirectives(namespace, directives, new ArrayList<>(detail.getSample()));
+  }
+
+
+  private List<Row> executeRemotely(String namespace, List<String> directives,
+                                    WorkspaceDetail detail) throws Exception {
+    LOG.info("Executing remotely");
```
yaojiefeng: doesn't feel this should be a info, make it a DEBUG

yaojiefeng (commented):
Took a first pass

wrangler-core/src/main/java/io/cdap/wrangler/registry/UserDirectiveRegistry.java
greeshmaswaminathan: SystemAppTaskContext provides only a subset of the contract provided by SystemHttpServiceContext, so cant substitute SystemAppTaskContext in place of SystemHttpServiceContext. But I can use the SystemAppTaskContext directly here instead of assigning PluginConfigurer and ArtifactManager to make it consistent with the existing one. This will however require introducing a new dependency.

wrangler-service/src/main/java/io/cdap/wrangler/service/common/AbstractWranglerHandler.java
greeshmaswaminathan: The workspace code is still executed locally, so it will not be part of the RemoteExecutionException. Will test to verify this.

wrangler-service/src/main/java/io/cdap/wrangler/service/directive/WorkspaceHandler.java
greeshmaswaminathan: I will remove this, was added for testing.

wrangler-proto/src/main/java/io/cdap/wrangler/proto/workspace/v2/WorkspaceDetail.java
```diff
@@ -21,17 +21,27 @@
...
 
   public WorkspaceDetail(Workspace workspace, List<Row> sample) {
     this.workspace = workspace;
     this.sample = sample;
+    this.sampleAsBytes = null;
+  }
+
+  public WorkspaceDetail(Workspace workspace, List<Row> sample, byte[] sampleAsBytes) {
```
CuriousVini, line 34: mark the constructor argument as nullable.

wrangler-proto/src/main/java/io/cdap/wrangler/proto/workspace/v2/WorkspaceDetail.java
CuriousVini, line 32: I think we will have to re-generate equals and hashcode methods

wrangler-service/src/main/java/io/cdap/wrangler/service/directive/WorkspaceHandler.java
```diff
@@ -253,6 +261,25 @@ public void execute(HttpServiceRequest request, HttpServiceResponder responder,
...
+    RemoteDirectiveRequest directiveRequest = new RemoteDirectiveRequest(directives,
+                                                                         namespace, detail.getSampleAsBytes());
+    RunnableTaskRequest runnableTaskRequest = RunnableTaskRequest.getBuilder(RemoteExecutionTask.class.getName()).
+      withParam(GSON.toJson(directiveRequest)).build();
+    byte[] bytes = getContext().runTask(runnableTaskRequest);
+    return CommonExecutor.convertToRows(bytes);
```
CuriousVini: how is error returned from `runTask`? If error is returned, `CommonExecutor#convertToRows ` should make sure its being handled

wrangler-service/src/main/java/io/cdap/wrangler/service/directive/CommonExecutor.java
```diff
@@ -0,0 +1,128 @@
...
+   * @param data byte[] with data
+   * @return {@link List} of {@link Row}
+   * @throws IOException if an I/O error occurs
+   * @throws ClassNotFoundException if the class being read as is not found
+   */
+  public static List<Row> convertToRows(byte[] data) throws IOException, ClassNotFoundException {
+    List<Row> rows = new ArrayList<>();
+    if (data != null) {
+      try (ObjectInputStream ois = new ObjectInputStream(new ByteArrayInputStream(data))) {
+        rows = (List<Row>) ois.readObject();
```
CuriousVini: would this always be object of list of rows? Trying to see what is data in case of error is returned from `runTask`

wrangler-service/src/main/java/io/cdap/wrangler/service/directive/WorkspaceHandler.java
CuriousVini: Is this called everytime a new directive is added? should we reduce log level?

CuriousVini (commented):
took first pass. a few comments

wrangler-service/src/main/java/io/cdap/wrangler/service/directive/WorkspaceHandler.java
greeshmaswaminathan: Will remove this, was added for test

wrangler-service/src/main/java/io/cdap/wrangler/service/directive/WorkspaceHandler.java
greeshmaswaminathan: An exception is returned on non 200 responses - https://github.com/cdapio/cdap/blob/8452f58fe423624369573d8c001e4df3c102cd36/cdap-app-fabric/src/main/java/io/cdap/cdap/internal/app/RemoteTaskExecutor.java#L75

wrangler-service/src/main/java/io/cdap/wrangler/service/common/AbstractWranglerHandler.java
greeshmaswaminathan: Tested to verify that 404 is returned when workspace is not found.

wrangler-proto/src/main/java/io/cdap/wrangler/proto/workspace/v2/WorkspaceDetail.java
greeshmaswaminathan: Done. Reusing ObjectSerDe to convert instead of adding new methods.

wrangler-service/src/main/java/io/cdap/wrangler/service/directive/CommonExecutor.java
greeshmaswaminathan: Done.

wrangler-service/src/main/java/io/cdap/wrangler/service/directive/WorkspaceHandler.java
greeshmaswaminathan: Done.

wrangler-proto/src/main/java/io/cdap/wrangler/proto/workspace/v2/WorkspaceDetail.java
greeshmaswaminathan: With latest changes this is non nullable.

wrangler-proto/src/main/java/io/cdap/wrangler/proto/workspace/v2/WorkspaceDetail.java
greeshmaswaminathan: Done.

wrangler-service/src/main/java/io/cdap/wrangler/service/directive/CommonExecutor.java
greeshmaswaminathan: Yes, if there is an error an exception will be thrown but wrong data will not be returned. Reusing ObjectSerDe instead of this method now.

wrangler-core/pom.xml
```diff
@@ -295,6 +295,12 @@
       <artifactId>hapi-structures-v26</artifactId>
       <version>${hl7.version}</version>
     </dependency>
+    <dependency>
+      <groupId>io.cdap.cdap</groupId>
+      <artifactId>cdap-system-app-api</artifactId>
+      <version>${cdap.version}</version>
+      <scope>compile</scope>
```
CuriousVini: no need for `compile` scope

wrangler-service/src/main/java/io/cdap/wrangler/service/directive/RemoteExecutionTask.java
```diff
@@ -0,0 +1,61 @@
...
+    );
+    composite.reload(namespace);
+    Function<TransientStore, ExecutorContext> contextProvider =
+      store -> new ServicePipelineContext(namespace, ExecutorContext.Environment.SERVICE, systemAppContext, store);
+    CommonDirectiveExecutor commonDirectiveExecutor = new CommonDirectiveExecutor(contextProvider, composite);
+    ObjectSerDe<List<Row>> objectSerDe = new ObjectSerDe<>();
+    List<Row> inputRows = objectSerDe.toObject(directiveRequest.getData());
+    List<Row> rows = commonDirectiveExecutor.executeDirectives(namespace, directiveRequest.getDirectives(), inputRows);
+    runnableTaskContext.writeResult(objectSerDe.toByteArray(rows));
+    composite.close();
```
CuriousVini: should this be in finally?

CuriousVini (commented):
few minor comments.

wrangler-core/pom.xml
greeshmaswaminathan: Removed.

wrangler-service/src/main/java/io/cdap/wrangler/service/directive/RemoteExecutionTask.java
greeshmaswaminathan: Yes, added a try-with-resource for the registry.

wrangler-service/src/main/java/io/cdap/wrangler/service/directive/RemoteExecutionTask.java
```diff
@@ -0,0 +1,62 @@
...
+  @Override
+  public void run(RunnableTaskContext runnableTaskContext) throws Exception {
+    RemoteDirectiveRequest directiveRequest = GSON
+      .fromJson(runnableTaskContext.getParam(), RemoteDirectiveRequest.class);
+    SystemAppTaskContext systemAppContext = runnableTaskContext.getRunnableTaskSystemAppContext();
+    String namespace = directiveRequest.getPluginNameSpace();
+    try (DirectiveRegistry composite = new CompositeDirectiveRegistry(
+      new SystemDirectiveRegistry(),
+      new UserDirectiveRegistry(systemAppContext)
+    )) {
```
CuriousVini: nit: `)) {` could be moved to above line.

CuriousVini (approved):
one minor comment, LGTM 👍🏼

wrangler-service/src/main/java/io/cdap/wrangler/service/directive/RemoteExecutionTask.java
greeshmaswaminathan: Done

PR #510, opened 2021-06-17
CDAP-17904 support connection/workspace upgrade

yaojiefeng:
Contains some unavoided hardcoding to convert the connection type to our connector. 
The upgraded connection and workspaces should just behave the new connections and workspaces, i.e, workspace automatically fill out spec, plugin info and use connection in source when transforming into studio page.
For connections and workspaces that are unsupported for upgrade, i.e, ADLS, or connections already get deleted, transforming it to studio will only have the wrangler stage. 
For database connections and workspaces, it is possible that we are not able to get schema due to incompatibility of jdbc url string. The rest should look like the same

### DIFF
--- a/wrangler-core/src/main/java/io/cdap/wrangler/utils/ArtifactSummaryComparator.java
+++ b/wrangler-core/src/main/java/io/cdap/wrangler/utils/ArtifactSummaryComparator.java
@@ -1,0 +1,54 @@
+/*
+ * Copyright © 2021 Cask Data, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ *
+ */
+
+package io.cdap.wrangler.utils;
+
+import io.cdap.cdap.api.artifact.ArtifactScope;
+import io.cdap.cdap.api.artifact.ArtifactSummary;
+import io.cdap.cdap.api.artifact.ArtifactVersion;
+
+import java.util.Comparator;
+
+/**
+ * Comparator for artifact summary
+ */
+public class ArtifactSummaryComparator implements Comparator<ArtifactSummary> {
+  private static final ArtifactSummaryComparator COMPARATOR = new ArtifactSummaryComparator();
+
+  @Override
+  public int compare(ArtifactSummary summary1, ArtifactSummary summary2) {
+    // first compare the artifact
+    int cmp = new ArtifactVersion(summary1.getVersion()).compareTo(new ArtifactVersion(summary2.getVersion()));
+    if (cmp > 0) {
+      return 1;
+    }
+
+    if (cmp < 0) {
+      return -1;
+    }
+
+    // if scope is different, whoever has user scope is latest
+    return summary1.getScope().equals(ArtifactScope.USER) ? 1 : -1;
+  }
+
+  /**
+   * Pick up the latest artifact, this method assumes the name of the artifact is same
+   */
+  public static ArtifactSummary pickLatest(ArtifactSummary artifact1, ArtifactSummary artifact2) {
+    return COMPARATOR.compare(artifact1, artifact2) > 0 ? artifact1 : artifact2;
+  }
+}
--- a/wrangler-service/src/main/java/io/cdap/wrangler/service/DataPrepService.java
+++ b/wrangler-service/src/main/java/io/cdap/wrangler/service/DataPrepService.java
@@ -17,6 +17,7 @@
 package io.cdap.wrangler.service;
 
 import io.cdap.cdap.api.service.AbstractSystemService;
+import io.cdap.cdap.api.service.SystemServiceContext;
 import io.cdap.wrangler.dataset.connections.ConnectionStore;
 import io.cdap.wrangler.dataset.schema.SchemaRegistry;
 import io.cdap.wrangler.dataset.workspace.ConfigStore;
@@ -26,8 +27,10 @@ import io.cdap.wrangler.service.bigquery.BigQueryHandler;
 import io.cdap.wrangler.service.connections.ConnectionHandler;
 import io.cdap.wrangler.service.connections.ConnectionTypeConfig;
 import io.cdap.wrangler.service.database.DatabaseHandler;
+import io.cdap.wrangler.service.directive.ConnectionUpgrader;
 import io.cdap.wrangler.service.directive.DirectivesHandler;
 import io.cdap.wrangler.service.directive.WorkspaceHandler;
+import io.cdap.wrangler.service.directive.WorkspaceUpgrader;
 import io.cdap.wrangler.service.explorer.FilesystemExplorer;
 import io.cdap.wrangler.service.gcs.GCSHandler;
 import io.cdap.wrangler.service.kafka.KafkaHandler;
@@ -35,12 +38,22 @@ import io.cdap.wrangler.service.s3.S3Handler;
 import io.cdap.wrangler.service.schema.DataModelHandler;
 import io.cdap.wrangler.service.schema.SchemaRegistryHandler;
 import io.cdap.wrangler.service.spanner.SpannerHandler;
+import io.cdap.wrangler.store.upgrade.UpgradeEntityType;
+import io.cdap.wrangler.store.upgrade.UpgradeState;
+import io.cdap.wrangler.store.upgrade.UpgradeStore;
 import io.cdap.wrangler.store.workspace.WorkspaceStore;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.concurrent.TimeUnit;
 
 /**
  * Data prep service.
  */
 public class DataPrepService extends AbstractSystemService {
+  private static final Logger LOG = LoggerFactory.getLogger(DataPrepService.class);
+  private static final UpgradeState PRE_UPGRADE = new UpgradeState(0L);
+
   private final ConnectionTypeConfig config;
 
   public DataPrepService(ConnectionTypeConfig config) {
@@ -57,6 +70,7 @@ public class DataPrepService extends AbstractSystemService {
     createTable(SchemaRegistry.ENTRY_TABLE_SPEC);
     createTable(WorkspaceDataset.TABLE_SPEC);
     createTable(WorkspaceStore.WORKSPACE_TABLE_SPEC);
+    createTable(UpgradeStore.UPGRADE_TABLE_SPEC);
 
     addHandler(new DirectivesHandler());
     addHandler(new SchemaRegistryHandler());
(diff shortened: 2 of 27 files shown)


wrangler-core/src/main/java/io/cdap/wrangler/utils/ArtifactSelectorUtils.java
```diff
@@ -0,0 +1,55 @@
...
+package io.cdap.wrangler.utils;
+
+import io.cdap.cdap.api.artifact.ArtifactScope;
+import io.cdap.cdap.api.artifact.ArtifactSummary;
+import io.cdap.cdap.api.artifact.ArtifactVersion;
+
+/**
+ * Artifact selector util
+ */
+public class ArtifactSelectorUtils {
```
albertshau: would be more useful to have a class that implements `Comparator<ArtifactSummary>` rather than this utility class

wrangler-service/src/main/java/io/cdap/wrangler/service/directive/WorkspaceUpgrader.java
```diff
@@ -0,0 +1,140 @@
...
+      WorkspaceDataset wsDataset = WorkspaceDataset.get(ctx);
+      return wsDataset.listWorkspaces(namespace, upgradeBeforeTsSecs);
+    });
+
+    for (Workspace workspace : workspaces) {
+      List<Row> sample = Collections.emptyList();
+      try {
+        sample = DirectivesHandler.fromWorkspace(workspace);
+      } catch (Exception e) {
+        // this should not happen, but guard here to avoid failing the entire upgrade process
```
albertshau, line 93: why wouldn't this happen? Isn't the call performing some I/O?

Logging a warning and moving on means its possible some workspaces never get upgraded. Seems like this should get propagated up.

wrangler-storage/src/main/java/io/cdap/wrangler/store/upgrade/UpgradeStore.java
```diff
@@ -0,0 +1,181 @@
...
+  private static final String CONNECTION_UPGRADE_COMPLETE_FIELD = "connection_upgrade_complete";
+  private static final String WORKSPACE_UPGRADE_COMPLETE_FIELD = "workspace_upgrade_complete";
+  private static final String UPGRADE_TIMESTAMP = "upgrade_timestamp";
+
+  public static final StructuredTableSpecification UPGRADE_TABLE_SPEC =
+    new StructuredTableSpecification.Builder()
+      .withId(TABLE_ID)
+      .withFields(Fields.stringType(NAMESPACE_COL),
+                  Fields.longType(GENERATION_COL),
+                  Fields.stringType(CONNECTION_UPGRADE_COMPLETE_FIELD),
```
albertshau: We should treat the UpgradeStore as something that can be re-used for future upgrades without needing to update the table schema or creating yet another table.

I think it would be better to have a column for the entity type (and add it to the primary key), with 'workspace' and 'connection' as possible values today. Have another column that contains the current state of that entity, which could be a serialized UpgradeState object which includes a version or something like that.

This way, it could be re-used for other purposes, such as upgrading drafts or other application specific data.

wrangler-service/src/main/java/io/cdap/wrangler/service/directive/SpecificationUpgradeUtils.java
```diff
@@ -0,0 +1,130 @@
...
+  }
+
+  /**
+   * Get the connector name for the given connection type.
+   *
+   * @param connectionType the connection type
+   * @return the connector name for the connection type
+   */
+  @Nullable
+  public static String getConnectorName(String connectionType) {
```
albertshau: instead of this method, the CONNECTION_TYPE_MAP, and and static initialization block, you can add another property to the ConnectionType enum for the corresponding connector name.

wrangler-service/src/main/java/io/cdap/wrangler/service/directive/WorkspaceHandler.java
```diff
@@ -81,17 +85,50 @@
...
 
   @Override
   public void initialize(SystemHttpServiceContext context) throws Exception {
     super.initialize(context);
-    store = new WorkspaceStore(context);
+    wsStore = new WorkspaceStore(context);
     discoverer = new ConnectionDiscoverer(context);
+
+    UpgradeStore upgradeStore = new UpgradeStore(context);
+    if (upgradeStore.isUpgradeComplete()) {
```
albertshau: initialize gets called for every thread created by the handler, and will only get called once a request is made to the handler. It is entirely possible for the upgrade logic to run multiple times with this logic.

I thought the plan was to add an initialize method to the AbstractService class and do the upgrade there?

wrangler-service/src/main/java/io/cdap/wrangler/service/directive/WorkspaceUpgrader.java
albertshau: ok looked at the fromWorkspace code, it throws exceptions if it can't deserialize, which shouldn't happen.

In either case, it's better for the log warning to be more specfic, like

"Could not decode the sample data for workspace {}. This workspace will not be upgraded"

wrangler-service/src/main/java/io/cdap/wrangler/service/directive/WorkspaceUpgrader.java
```diff
@@ -0,0 +1,140 @@
...
+      long now = System.currentTimeMillis();
+      io.cdap.wrangler.proto.workspace.v2.Workspace.Builder ws =
+        io.cdap.wrangler.proto.workspace.v2.Workspace
+          .builder(workspace.getName(), workspace.getNamespacedId().getId()).setDirectives(directives)
+          .setCreatedTimeMillis(now).setUpdatedTimeMillis(now);
+
+      String connectorName = SpecificationUpgradeUtils.getConnectorName(
+        workspace.getProperties().get(PropertyIds.CONNECTION_TYPE).toLowerCase());
+      String path = SpecificationUpgradeUtils.getPath(workspace);
+      // if the connection type cannot be upgraded, create a workspace with empty sample spec, so there will be
```
albertshau: rather than use null for this special purpose, it's cleaner to have a method that specifically checks if a given connection type can be upgraded.

albertshau: is this just for ADLS? Should we just skip upgrade for these types of workspaces since you won't be able to see them in the new UX anyway?

wrangler-service/src/main/java/io/cdap/wrangler/service/directive/WorkspaceUpgrader.java
```diff
@@ -0,0 +1,140 @@
...
+              new Artifact(artifact.getName(), artifact.getVersion(), artifact.getScope()));
+            return new StageSpec(plugin.getSchema(), pluginSpec);
+          }).collect(Collectors.toSet()));
+        wsStore.saveWorkspace(workspaceId, new WorkspaceDetail(ws.setSampleSpec(spec).build(), sample));
+      } catch (NotFoundException e) {
+        LOG.warn("Connection {} related to workspace {} does not exist. " +
+                   "The workspace will be upgraded without that information", connection, workspace.getName());
+        wsStore.saveWorkspace(workspaceId, new WorkspaceDetail(ws.build(), sample));
+      } catch (Exception e) {
+        LOG.warn("Unable to get the spec from connection {} for workspace {}. " +
```
albertshau: when does this happen? Is it a transient exception? Does it make more sense to retry for some time and then give up?

upgrading this workspace without the spec would put it in a broken state that could never be fixed right? Seems better not to upgrade it at all rather than do that.

wrangler-service/src/main/java/io/cdap/wrangler/service/directive/ConnectionUpgrader.java
```diff
@@ -0,0 +1,112 @@
...
+    for (Connection connection : connections) {
+      // do not upgrade pre configured connection
+      if (connection.isPreconfigured()) {
+        continue;
+      }
+
+      // if it is not upgradable
+      String connectionType = connection.getType().name().toLowerCase();
+      String connectorName = SpecificationUpgradeUtils.getConnectorName(connectionType);
+      if (connectorName == null) {
```
albertshau: better to have a specific method for checking if a connection type can be upgraded, rather than making null having this special meaning.

wrangler-service/src/main/java/io/cdap/wrangler/service/directive/WorkspaceUpgrader.java
yaojiefeng: it is not just for ADLS, also for workspaces created with direct upload, I think we will not do upgrade for ADLS but still do upgrade for the upload file, as the UX will support upload file to create a workspace

wrangler-service/src/main/java/io/cdap/wrangler/service/directive/WorkspaceUpgrader.java
yaojiefeng: The get spec call in discoverer already have retry logic, it will only throw exception when a non-retryable exception happens or retry timed out. 
This can happen for database related connectors where the url does not include the database name, I think we should still upgrade them to let them show up in the new UX

wrangler-core/src/main/java/io/cdap/wrangler/utils/ArtifactSelectorUtils.java
yaojiefeng: Changed this class to a `Comparator` class but still preserve the `pickLatest` static method as it is useful on picking an artifact when the code does not really care which one is latest

wrangler-service/src/main/java/io/cdap/wrangler/service/directive/WorkspaceHandler.java
yaojiefeng: Have this https://github.com/cdapio/cdap/pull/13487 to have the program life cycle for service, changed this pr to do the upgrade on the first instance of the service initialize

wrangler-service/src/main/java/io/cdap/wrangler/service/directive/WorkspaceUpgrader.java
yaojiefeng: Actually if we are able to create the workspace for db relations, it is quite possible we are able to get spec back, changed this to not upgrade and just leave an error message

wrangler-service/src/main/java/io/cdap/wrangler/service/directive/WorkspaceUpgrader.java
yaojiefeng: Updated logic: now it will only upgrade workspaces in the allowed connection type list. Only UPLOAD type workspace will get created without a sample spec

wrangler-core/src/main/java/io/cdap/wrangler/utils/ArtifactSummaryComparator.java
```diff
@@ -22,34 +22,38 @@
...
     // if scope is different, whoever has user scope is latest
-    return artifact1.getScope().equals(ArtifactScope.USER);
+    return summary.getScope().equals(ArtifactScope.USER) ? 1 : -1;
+  }
+
+  /**
+   * Pick up the latest artifact, this method assumes the name of the artifact is same
+   */
+  public static ArtifactSummary pickLatest(ArtifactSummary artifact1, ArtifactSummary artifact2) {
+    ArtifactSummaryComparator comparator = new ArtifactSummaryComparator(artifact1);
```
albertshau: it's better not to create objects needlessly like this. You can have a static instance of a Comparator and use that here instead of always creating a new object.

wrangler-service/src/main/java/io/cdap/wrangler/service/DataPrepService.java
```diff
@@ -74,4 +84,49 @@ protected void configure() {
...
+  @Override
+  public void initialize(SystemServiceContext context) {
+    // only do the upgrade on first instance to avoid transaction conflict
+    if (context.getInstanceId() != 0) {
+      return;
+    }
+
+    UpgradeStore upgradeStore = new UpgradeStore(context);
+    WorkspaceStore wsStore = new WorkspaceStore(context);
+    boolean isConnDone = upgradeStore.isEntityUpgradeComplete(UpgradeEntityType.CONNECTION);
```
albertshau: this logic assumes one instance of the service, which is the normal setup but not necessarily the case for all setups. Does the ServiceContext expose the instance id? If so you can check that to make sure only instance 0 upgrades.

wrangler-service/src/main/java/io/cdap/wrangler/service/directive/WorkspaceUpgrader.java
```diff
@@ -98,21 +101,30 @@ private void upgradeWorkspacesInConnections(NamespaceSummary namespace) {
...
-        workspace.getProperties().get(PropertyIds.CONNECTION_TYPE).toLowerCase());
-      String path = SpecificationUpgradeUtils.getPath(workspace);
-      // if the connection type cannot be upgraded, create a workspace with empty sample spec, so there will be
-      // no sources created when converting to pipeline
-      if (path == null) {
+      ConnectionType connectionType =
+        ConnectionType.valueOf(workspace.getProperties().get(PropertyIds.CONNECTION_TYPE).toUpperCase());
+      // if it is not upgradable workspace types, just ignore and continue, i.e, ADLS
+      if (!ConnectionType.WORKSPACE_UPGRADABLE_TYPES.contains(connectionType)) {
+        LOG.warn("Workspace {} of type {} is not upgradable. This workspace will not be upgraded",
```
albertshau: don't need to warn, since this is expected behavior. Debug should be ok.

wrangler-storage/src/main/java/io/cdap/wrangler/store/upgrade/UpgradeState.java
```diff
@@ -0,0 +1,34 @@
...
+ */
+
+package io.cdap.wrangler.store.upgrade;
+
+/**
+ * Upgrade state to store any upgrade related information. Can be extended to contain more information about the
+ * upgrade information
+ */
+public class UpgradeState {
+  private final boolean isUpgradeComplete;
```
albertshau: It would be better for this to be a version number, so that if we need to upgrade anything in the future we can still use this. 

Similarly, better if the UpgradeStore returns this object and the caller makes the decision of whether or not upgrade needs to run, instead of having the UpgradeStore return a boolean on whether upgrade is done or not.

wrangler-service/src/main/java/io/cdap/wrangler/service/DataPrepService.java
yaojiefeng: I added the logic of checking instance id here - https://github.com/data-integrations/wrangler/pull/510/files/b1546431b6c791713cd71577d381e9a1a00dea0f..30c1ddf35ccd97a057ec1f3e09f5a42c256284de#diff-4a6d56260e39433bf54fa3ac1e682aa59ee6b1cd6218003d07a3b882cce14795R91-R93, it will only upgrade on instance 0

albertshau (approved):

PR #513, opened 2021-06-17
CDAP-17933 fix a regression where rename will fail on non-existing co…

yaojiefeng:
…lumns

### DIFF
--- a/wrangler-core/src/main/java/io/cdap/wrangler/utils/ColumnConverter.java
+++ b/wrangler-core/src/main/java/io/cdap/wrangler/utils/ColumnConverter.java
@@ -18,6 +18,7 @@ package io.cdap.wrangler.utils;
 import io.cdap.cdap.api.common.Bytes;
 import io.cdap.wrangler.api.DirectiveExecutionException;
 import io.cdap.wrangler.api.Row;
+
 import java.math.BigDecimal;
 
 /**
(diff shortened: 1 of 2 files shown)


wrangler-core/src/main/java/io/cdap/wrangler/utils/ColumnConverter.java
```diff
@@ -34,18 +35,14 @@ private ColumnConverter() {
...
-   * @throws DirectiveExecutionException when a column matching the target name already exists or the given column
-   * does not exist
+   * @throws DirectiveExecutionException when a column matching the target name already exists
    */
   public static void rename(String directiveName, Row row, String column, String toName)
     throws DirectiveExecutionException {
     int idx = row.find(column);
     int existingColumn = row.find(toName);
     if (idx == -1) {
-      throw new DirectiveExecutionException(
```
albertshau, line 45: I don't get how this fixes the bug described in CDAP-17933. Why would this behave differently on different rows?

This looks like an incompatible behavioral change. For example, 'rename XYZ xyz' becomes a no-op, and other renames silently do nothing.

yaojiefeng: The `find` method in row is a case insensitive find, so `rename XYZ xyz` will still rename since it will able to find `XYZ` if the column exists. 
In 6.3 and below, the behavior is actually do nothing if the column does not exist, https://github.com/data-integrations/wrangler/pull/470/files#diff-7db95e4e0141e52d5f85699269693ef20f5e1df2a47eb1000c9a3a5a618196aaL42-L50. In 6.4, we added this check, so customer is complaining the rename column needs to become non-null before applying the directive. I think it is fine we do nothing, since in order for UI to show the column, the whole sample must contain this column. 
The fundamental problem is that row is not like `StructuredRecord` which has a schema and a list of rows is not constructed consistently, a row does not necessarily to have all the columns in it. I added more tests on non-existing cols on non-existing cols

albertshau: I see, got it.

albertshau (approved):

PR #514, opened 2021-06-18
Fix : Handle null values for optional array fields

VictorD-Veolia:
With the current version of the wrangler, null fields would throw an exception during decoding, even if the field is defined as nullable in the schema. I added a check in the decode function.


bdmogal:
@VictorD-Veolia apologies for the delay on this. Thanks for looking into this issue. Could you please open a bug for this issue describing what is broken, and perhaps add a test as well?

VictorD-Veolia:
Hi @bdmogal. I added a unit test for the behavior fixed and created an issue in JIRA : https://cdap.atlassian.net/browse/PLUGIN-794

wrangler-core/src/main/java/io/cdap/wrangler/utils/RecordConvertor.java
```diff
@@ -111,16 +111,22 @@ public StructuredRecord decodeRecord(Row row, Schema schema) throws RecordConver
...
+
+    if (object == null && isNullable) {
+      return null;
+    }
+
     // Extract the type of the field.
     Schema.Type type = schema.getType();
     Schema.LogicalType logicalType =
-        schema.isNullable() ? schema.getNonNullable().getLogicalType() :
+            isNullable ? schema.getNonNullable().getLogicalType() :
```
yaojiefeng: nit - can this move to previous line?

wrangler-core/src/test/java/io/cdap/wrangler/utils/RecordConvertorTest.java
```diff
@@ -162,6 +162,31 @@ public void testNullableEmptyField() throws Exception {
...
+    Schema schema = Schema.recordOf("test",
+            Schema.Field.of("test_id", Schema.of(Schema.Type.STRING)),
+            Schema.Field.of("values", Schema.nullableOf(
+                    Schema.arrayOf(Schema.nullableOf(Schema.of(Schema.Type.INT))))));
+
+    String[] directives = new String[] {
+    };
+
+    List<Row> rows = Arrays.asList(
+            new Row().add("test_id", "a").add("values",ImmutableList.of(1)),
```
yaojiefeng: nit - indentation should be 2 spaces.

yaojiefeng (approved):
The indentation should be 2 spaces, please fix it, otherwise code LGTM

PR #517, opened 2021-06-23
CDAP-18097 - Add NPE check when getting ArtifactManager

greeshmaswaminathan:
CDAP-18097 - Add NPE check when getting ArtifactManager


fernst (approved):
LGTM

PR #518, opened 2021-06-23
Add NPE check when getting ArtifactManager

greeshmaswaminathan:
CDAP-18097 - Add NPE check when getting ArtifactManager


fernst (approved):
LGTM

wrangler-core/src/main/java/io/cdap/wrangler/registry/UserDirectiveRegistry.java
```diff
@@ -228,8 +228,10 @@ public void reload(String namespace) throws DirectiveLoadException {
     }
   }
 
+  @Nullable
```
yaojiefeng, line 231: The ArtifactManager should never be null in any case, if this is null, this means udd will never work in pipeline

PR #524, opened 2021-07-20
Wrangler performance optimizations

tivv:
* Cache of lowercase column names in Row for lookup. During Row->Structured records currently it's n^2 of toLowerCase, this should make it n for toLowerCase
 * javax.script.SimpleScriptContext constructor is pretty expensive as it unconditionally creates a number of Buffered readers/writer. This change caches it as ThreadLocal

### DIFF
--- a/wrangler-api/src/main/java/io/cdap/wrangler/api/Row.java
+++ b/wrangler-api/src/main/java/io/cdap/wrangler/api/Row.java
@@ -22,7 +22,10 @@ import org.slf4j.LoggerFactory;
 
 import java.io.Serializable;
 import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.List;
+import java.util.Locale;
+import java.util.Map;
 import java.util.Objects;
 
 /**
@@ -35,6 +38,9 @@ public final class Row implements Serializable {
 
   // Name of the columns held by the row.
   private List<String> columns = new ArrayList<>();
+
+  // Map of lower case column names to speed up lookup
+  private transient Map<String, Integer> nameIndex;
 
   // Values held by the row.
   private List<Object> values = new ArrayList<>();
@@ -50,6 +56,7 @@ public final class Row implements Serializable {
   public Row(Row row) {
     this.values = new ArrayList<>(row.values);
     this.columns = new ArrayList<>(row.columns);
+    this.nameIndex = row.nameIndex == null ? null : new HashMap<>(row.nameIndex);
   }
 
   /**
@@ -93,6 +100,7 @@ public final class Row implements Serializable {
    */
   public void setColumn(int idx, String name) {
     columns.set(idx, name);
+    nameIndex = null;
   }
 
   /**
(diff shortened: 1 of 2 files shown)


wrangler-api/src/main/java/io/cdap/wrangler/api/Row.java
```diff
@@ -141,6 +156,9 @@ public Row setValue(int idx, Object value) {
   public Row add(String name, Object value) {
     columns.add(name);
     values.add(value);
+    if (nameIndex != null && name != null) {
+      nameIndex.put(name.toLowerCase(Locale.ENGLISH), columns.size());
```
seanzhougoogle: `columns.size()-1` ?

wrangler-api/src/main/java/io/cdap/wrangler/api/Row.java
```diff
@@ -150,8 +168,11 @@ public Row add(String name, Object value) {
    * @param idx for which the value and column are removed.
    */
   public Row remove(int idx) {
-    columns.remove(idx);
+    String name = columns.remove(idx);
     values.remove(idx);
+    if (name != null && nameIndex != null) {
+      nameIndex.remove(name.toLowerCase(Locale.ENGLISH));
```
seanzhougoogle: remove from the middle will also change the index after it , right ?

wrangler-transform/src/main/java/io/cdap/wrangler/Precondition.java
```diff
@@ -45,14 +49,28 @@ public Precondition(String condition) throws PreconditionException {
     }
   }
 
+  public ScriptContext createContext() {
```
seanzhougoogle, line 52: Not familiar with ScriptConext, I saw this part is copied from `CompiledScript.eval`, is it implementation specific ? will it work with later java versions ?

wrangler-api/src/main/java/io/cdap/wrangler/api/Row.java
fernst: good catch, is this a 0-based or 1-based index?

wrangler-transform/src/main/java/io/cdap/wrangler/Precondition.java
tivv: It does not use any private API, so I don't expect problems here

tivv: Also it's not from specific implementation, but from CompiledScript itself

wrangler-api/src/main/java/io/cdap/wrangler/api/Row.java
tivv: Done

wrangler-api/src/main/java/io/cdap/wrangler/api/Row.java
tivv: Yes, good catch, fixed.

seanzhougoogle (approved):

PR #526, opened 2021-07-22
One more set of Wrangler Row performance optimizations.

tivv:
Looks like using a map does not speed things up, so let's just optimize a Row->StructuredRecord use case when wrangler field set is similar to output schema

### DIFF
--- a/wrangler-api/src/main/java/io/cdap/wrangler/api/Row.java
+++ b/wrangler-api/src/main/java/io/cdap/wrangler/api/Row.java
@@ -22,10 +22,7 @@ import org.slf4j.LoggerFactory;
 
 import java.io.Serializable;
 import java.util.ArrayList;
-import java.util.HashMap;
 import java.util.List;
-import java.util.Locale;
-import java.util.Map;
 import java.util.Objects;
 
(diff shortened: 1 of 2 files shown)


fernst (approved):
LGTM, we should add some tests but it can be done in a later PR.

wrangler-api/src/main/java/io/cdap/wrangler/api/Row.java
```diff
@@ -174,17 +162,17 @@ public Row remove(int idx) {
...
-      //We may have duplicate names, start from end to get the first index retained in the map
-      for (int i = columns.size() - 1; i >= 0; i--) {
-        String name = columns.get(i);
-        if (name != null) {
-          nameIndex.put(name.toLowerCase(Locale.ENGLISH), i);
-        }
+    return find(col, 0);
+  }
+
+  public int find(String col, int firstIdx) {
```
fernst, line 176: nit: add a comment for this method

tivv: Done!

PR #527, opened 2021-07-23
[Cherry-pick into 4.5] One more set of Wrangler Row performance optimizations.

tivv:
Looks like using a map does not speed things up, so let's just optimize a Row->StructuredRecord use case when wrangler field set is similar to output schema
Cherry-pick of https://github.com/data-integrations/wrangler/pull/526


fernst (approved):
LGTM

PR #528, opened 2021-07-23
CDAP-18262 do not finish computing summary if it is stuck for too long

yaojiefeng:
The library we use will take > 5s for a value of string > 1000 characters, add this workaround to timeout the computation.

### DIFF
--- a/wrangler-core/src/main/java/io/cdap/wrangler/statistics/BasicStatistics.java
+++ b/wrangler-core/src/main/java/io/cdap/wrangler/statistics/BasicStatistics.java
@@ -23,9 +23,11 @@ import java.util.List;
 import java.util.Map;
 
 /**
- * Created by nitin on 2/4/17.
+ * Basic class to compute summary from a list of rows
  */
 public class BasicStatistics implements Statistics {
+  // default time out be 10s
+  private static final long TIME_OUT_MILLIS = 10000;
   private final FinderEngine engine;
 
   public BasicStatistics() throws Exception {
@@ -37,6 +39,7 @@ public class BasicStatistics implements Statistics {
     ColumnMetric types = new ColumnMetric();
     ColumnMetric stats = new ColumnMetric();
 
+    long startTime = System.currentTimeMillis();
     Double count = new Double(0);
     for (Row row : rows) {
       ++count;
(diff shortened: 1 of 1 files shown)


wrangler-core/src/main/java/io/cdap/wrangler/statistics/BasicStatistics.java
```diff
@@ -55,10 +58,15 @@ public Row aggregate(List<Row> rows) {
           if (value.isEmpty()) {
             stats.increment(column, "empty");
           } else {
+            // this call is very expensive for string > 2000 characters, took seconds to return
             Map<String, List<String>> finds = engine.findWithType(value);
             for (String find : finds.keySet()) {
               types.increment(column, find);
             }
+            // TODO: this is a workaround for CDAP-18262, to proper fix we should revisit this computation logic
```
seanzhougoogle, line 66: file a TODO jira for this ? or just leave CDAP-18262 open and mark fix version as 6.6.0 ?

seanzhougoogle (approved):

wrangler-core/src/main/java/io/cdap/wrangler/statistics/BasicStatistics.java
yaojiefeng: I will file a new JIRA for this and close 18262

wrangler-core/src/main/java/io/cdap/wrangler/statistics/BasicStatistics.java
```diff
@@ -55,10 +58,15 @@ public Row aggregate(List<Row> rows) {
...
+            if (System.currentTimeMillis() - now > TIME_OUT_MILLIS) {
```
albertshau: now is a confusing variable name, better to name it start or something like that

wrangler-core/src/main/java/io/cdap/wrangler/statistics/BasicStatistics.java
yaojiefeng: https://cdap.atlassian.net/browse/CDAP-18277

wrangler-core/src/main/java/io/cdap/wrangler/statistics/BasicStatistics.java
yaojiefeng: Fixed

PR #533, opened 2021-07-31
[PLUGIN-768] Fix Flatten so that it correctly handles empty arrays

sechegaray:


wrangler-core/src/main/java/io/cdap/directives/row/Flatten.java
```diff
@@ -110,6 +110,11 @@ public void destroy() {
       }
 
       if (max == 0) {
+        for (int i = 0; i < count; ++i) {
```
tivv, line 114: could you please mutate a copy? I don't think it makes a difference here, but it's better not to have side effects.

sechegaray: Done

sechegaray: done

tivv (approved):

PR #534, opened 2021-07-31
Feature/add struct array datatype to wrangler

sechegaray:


wrangler-core/src/main/java/io/cdap/directives/row/Flatten.java
```diff
@@ -110,6 +110,11 @@ public void destroy() {
       }
 
       if (max == 0) {
+        for (int i = 0; i < count; ++i) {
```
tivv: I believe this is part of other PR

wrangler-core/src/main/java/io/cdap/wrangler/utils/RecordConvertor.java
```diff
@@ -69,6 +69,9 @@
    * @return Populated {@link StructuredRecord}
    */
   public StructuredRecord decodeRecord(Row row, Schema schema) throws RecordConvertorException {
+    if (row == null) {
```
tivv, line 74: Please annotate parameter and return value as `@Nullable`

wrangler-core/src/main/java/io/cdap/wrangler/utils/RecordConvertor.java
```diff
@@ -87,7 +90,7 @@ public StructuredRecord decodeRecord(Row row, Schema schema) throws RecordConver
       String name = field.getName();
       Object value = null;
       int idx = -1;
-      if (name.equals(row.getColumn(firstUnclaimedField))) {
+      if ((firstUnclaimedField < row.width()) && (name.equals(row.getColumn(firstUnclaimedField)))) {
```
tivv: Could you please make this a separate PR and cherry-pick into current release?

wrangler-core/src/main/java/io/cdap/wrangler/utils/RecordConvertor.java
```diff
@@ -196,8 +199,14 @@ private Object decode(String name, Object object, Schema schema) throws RecordCo
   }
 
   private StructuredRecord decodeRecord(String name, Object object, Schema schema) throws RecordConvertorException {
+    if (object == null) {
```
tivv, line 206: Same here. If you need to treat something as nullable, please annotate it

wrangler-core/src/main/java/io/cdap/wrangler/utils/SchemaConverter.java
```diff
@@ -152,6 +155,16 @@ public Schema getSchema(Object value, String name) throws RecordConvertorExcepti
...
     }
 
+    if (value instanceof Row) {
+      ArrayList<Field> fields = new ArrayList<>();
+      for (Pair<String, Object> field: ((Row) value).getFields()) {
+        if (field.getSecond() != null) {
+          fields.add(Field.of(field.getFirst(), getSchema(field.getSecond(), field.getFirst())));
+        }
+      }
+      return Schema.nullableOf(Schema.recordOf(UUID.randomUUID().toString().replace("-", ""), fields));
```
tivv: Please use stable names for schema as discussed

wrangler-core/src/main/java/io/cdap/wrangler/utils/SchemaConverter.java
```diff
@@ -225,7 +238,7 @@ private Schema toSchema(String name, JsonObject object) throws RecordConvertorEx
     if (fields.size() == 0) {
       return Schema.nullableOf(Schema.arrayOf(Schema.of(Schema.Type.NULL)));
     }
-    return Schema.recordOf(name, fields);
+    return Schema.recordOf(UUID.randomUUID().toString().replace("-", ""), fields);
```
tivv: Please use stable names as discussed, also make it backwards-compatible

wrangler-core/src/test/java/io/cdap/directives/row/FlattenTest.java
```diff
@@ -45,7 +45,7 @@ public void testEmptyList() throws Exception {
 
     Assert.assertEquals(1, output.size());
     Assert.assertEquals("y", output.get(0).getValue("y"));
-    Assert.assertEquals(0, ((List) output.get(0).getValue("x")).size());
+    Assert.assertNull(((List) output.get(0).getValue("x")));
```
tivv: I believe this is part of other PR

wrangler-service/src/main/java/io/cdap/wrangler/service/directive/AbstractDirectiveHandler.java
```diff
@@ -97,6 +107,76 @@ public void destroy() {
...
+    }
+
+    private TypeAdapter<Row> rowClassAdapter(Gson gson) {
+      return new TypeAdapter<Row>() {
+        @Override
+        public void write(JsonWriter out, Row value) throws IOException {
+          if (value == null) {
+            return;
+          }
+          final TypeAdapter<Object> elementAdapter = gson.getAdapter(Object.class);
```
tivv: nit: move this into `rowClassAdapter` method. This would ensure it's called once and not for each record, similar to `defaultClassAdapter`

wrangler-service/src/main/java/io/cdap/wrangler/service/directive/AbstractDirectiveHandler.java
```diff
@@ -97,6 +107,76 @@ public void destroy() {
...
+
+        @Override public
+        Row read(JsonReader in) throws IOException {
+          throw new UnsupportedOperationException("Reading Rows from Wrangler display format not implemented");
+        }
+      };
+    }
+
+    private TypeAdapter<Object> defaultClassAdapter(Gson gson) {
+      final TypeAdapter<Object> elementAdapter = gson.getAdapter(Object.class);
```
tivv: is this used at all?

wrangler-service/src/main/java/io/cdap/wrangler/service/directive/AbstractDirectiveHandler.java
```diff
@@ -97,6 +107,76 @@ public void destroy() {
...
+      return new TypeAdapter<Object>() {
+        @Override
+        public void write(JsonWriter out, Object value) throws IOException {
+          try {
+            if ((value.getClass().getMethod("toString").getDeclaringClass() != Object.class)) {
+              out.value(value.toString());
+            } else {
+              out.value("Non-displayable object");
```
tivv: Could you please confirm with product on what to say here. I'd prefer something shorter as those JSONs going to be displayed as 1 line in UI

wrangler-service/src/main/java/io/cdap/wrangler/service/directive/AbstractDirectiveHandler.java
```diff
@@ -146,6 +238,7 @@ protected DirectiveExecutionResponse generateExecutionResponse(
     return new DirectiveExecutionResponse(values, headers, types, getWorkspaceSummary(rows));
   }
 
+
```
tivv: Do you need this extra empty line?

tivv (commented):
Please add a feature flag to ensure we don't expose this feature until after it's fully implemented and tested

wrangler-core/src/main/java/io/cdap/directives/column/CreateRecord.java
```diff
@@ -0,0 +1,110 @@
...
+import io.cdap.wrangler.api.parser.ColumnNameList;
+import io.cdap.wrangler.api.parser.TokenType;
+import io.cdap.wrangler.api.parser.UsageDefinition;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+
+/**
+ * A directive that Creates a record from columns
```
yaojiefeng: Creates -> creates

wrangler-core/src/main/java/io/cdap/directives/column/ExplodeRecord.java
```diff
@@ -0,0 +1,123 @@
...
+@Name(ExplodeRecord.NAME)
+@Categories(categories = { "column"})
+@Description("Flattens a record into individual columns.")
+public class ExplodeRecord implements Directive, Lineage {
+  public static final String NAME = "explode-record";
+  // Column on which to apply mask.
+  private String[] columns;
+
+  // Map to Keep Track new Columns;
+  private ArrayList<String> newColumnNames = new ArrayList<>();
```
yaojiefeng: `List<String>`, no need to use concrete object type

wrangler-core/src/main/java/io/cdap/directives/column/ExplodeRecord.java
```diff
@@ -0,0 +1,123 @@
...
+        if (idx != -1) {
+          Object object = row.getValue(idx);
+
+          if (object == null) {
+            continue;
+          }
+
+          if (!(object instanceof Row)) {
+            throw new DirectiveExecutionException(
+                NAME,
```
yaojiefeng: nit - indentation should be 2 characters instead of 4.

wrangler-core/src/main/java/io/cdap/directives/column/ExplodeRecord.java
```diff
@@ -0,0 +1,123 @@
...
+      }
+      results.add(row);
+    }
+    return results;
+  }
+
+  @Override
+  public Mutation lineage() {
+    return Mutation.builder()
+        .readable("Flatten the record in columns %s into columns", Arrays.asList(columns))
```
yaojiefeng: nit - indentation should be 2 characters instead of 4.

wrangler-core/src/main/java/io/cdap/directives/column/ExplodeRecord.java
```diff
@@ -0,0 +1,123 @@
...
+        .all(Many.columns(columns), Many.columns(newColumnNames.toArray(new String[0]))
+        )
```
yaojiefeng: this doesn't need an extra line

wrangler-core/src/main/java/io/cdap/directives/column/ExplodeRecord.java
```diff
@@ -0,0 +1,123 @@
...
+  }
+
+  @Override
+  public List<Row> execute(List<Row> rows, ExecutorContext context) throws DirectiveExecutionException {
+    List<Row> results = new ArrayList<>();
+
+    for (Row row : rows) {
+      for (String column: columns) {
+        int idx = row.find(column);
+        if (idx != -1) {
```
yaojiefeng: better change it to 
```
if (idx == -1) {
  results.add(row); 
  continue;
}
...
```
to avoid the reduce the nested lines

wrangler-core/src/main/java/io/cdap/wrangler/utils/RecordConvertor.java
```diff
@@ -64,11 +65,15 @@
...
    *
-   * @param row defines a single {@link Row}
+   * @param row    defines a single {@link Row}
    * @param schema Schema associated with {@link StructuredRecord}
    * @return Populated {@link StructuredRecord}
    */
-  public StructuredRecord decodeRecord(Row row, Schema schema) throws RecordConvertorException {
+  @Nullable
+  public StructuredRecord decodeRecord(@Nullable Row row, Schema schema) throws RecordConvertorException {
+    if (row == null) {
```
yaojiefeng, line 74: Under what condition this will be null? I feel in Wrangler, the row will not be null right?

wrangler-core/src/main/java/io/cdap/directives/column/ExplodeRecord.java
yaojiefeng: This object seems only used to track the generated columns for lineage purpose, and is updated in the execute method. This won't work since the `lineage()` call is invoked in `prepareRun`, which is before the `execute` call

wrangler-core/src/main/java/io/cdap/directives/column/ExplodeRecord.java
```diff
@@ -0,0 +1,123 @@
...
+          }
+
+          if (!(object instanceof Row)) {
+            throw new DirectiveExecutionException(
+                NAME,
+                String.format("Column '%s' has invalid type '%s'. It should be of type 'Record'.",
+                    column, object.getClass().getSimpleName()));
+          }
+
+          Row rowColumns = ((Row) object);
```
yaojiefeng: ((Row) object) -> (Row) object

wrangler-core/src/main/java/io/cdap/directives/column/ExplodeRecord.java
```diff
@@ -0,0 +1,123 @@
...
+ * A directive for splitting a RowType ("Record") Column into multiple Columns.
+ */
+@Plugin(type = Directive.TYPE)
+@Name(ExplodeRecord.NAME)
+@Categories(categories = { "column"})
+@Description("Flattens a record into individual columns.")
+public class ExplodeRecord implements Directive, Lineage {
+  public static final String NAME = "explode-record";
+  // Column on which to apply mask.
+  private String[] columns;
```
yaojiefeng: I don't understand why this directive has to work on a list of columns, shouldn't it just operate on one column, and if a user wants to explode records on multiple columns, he can invoke this directive multiple times on different columns. Making this a list might result in a large change in the workspace which might be confusing to user

wrangler-core/src/main/java/io/cdap/directives/column/ExplodeRecord.java
```diff
@@ -0,0 +1,123 @@
...
+          }
+
+          Row rowColumns = ((Row) object);
+
+          String columnName;
+          newColumnNames.add((String) column);
+          for (Pair<String, Object> field : rowColumns.getFields()) {
+            columnName = String.format("%s_%s", column, field.getFirst());
+            newColumnNames.add(columnName);
+            row.add(columnName, field.getSecond());
```
yaojiefeng: Should we use `addOrSet` here? If there is a conflict in the column name, this change will result in two columns with same name, which can cause weird behavior in the following directives execution

wrangler-core/src/main/java/io/cdap/directives/column/ExplodeRecord.java
```diff
@@ -0,0 +1,123 @@
...
+import io.cdap.wrangler.api.parser.Text;
+import io.cdap.wrangler.api.parser.TokenType;
+import io.cdap.wrangler.api.parser.UsageDefinition;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+
+/**
+ * A directive for splitting a RowType ("Record") Column into multiple Columns.
```
yaojiefeng: We need much better documentation here, especially on what the resulting column names will look like(exploded col + "_" + original col name) will be after flatten, it should also mention what will happen if there is a conflict in the colume name after flatten, whether the existing column is replaced or a column with same name will be added(which should not be the case since we don't allow field with same name in StructuredRecord)

wrangler-core/src/main/java/io/cdap/directives/column/ExplodeRecord.java
```diff
@@ -0,0 +1,123 @@
...
+      results.add(row);
+    }
+    return results;
+  }
+
+  @Override
+  public Mutation lineage() {
+    return Mutation.builder()
+        .readable("Flatten the record in columns %s into columns", Arrays.asList(columns))
+        .all(Many.columns(columns), Many.columns(newColumnNames.toArray(new String[0]))
```
yaojiefeng: Our lineage does not work well for nested fields, I think it is fine if this directive does not extend from `Lineage`

wrangler-core/src/main/java/io/cdap/wrangler/utils/SchemaConverter.java
```diff
@@ -152,6 +168,23 @@ public Schema getSchema(Object value, String name) throws RecordConvertorExcepti
       }
     }
 
+    if (value instanceof Row) {
+      ArrayList<Field> fields = new ArrayList<>();
```
yaojiefeng: `List<Field>`

wrangler-core/src/main/java/io/cdap/wrangler/utils/SchemaConverter.java
```diff
@@ -208,13 +245,25 @@ private Schema toSchema(String name, JsonPrimitive primitive) throws RecordConve
   }
 
   private Schema toSchema(String name, JsonObject object) throws RecordConvertorException {
+    return toSchema(name, object, null);
+  }
+
+  private Schema toSchema(String name,
+                        JsonObject object, String recordPrefix) throws RecordConvertorException {
```
yaojiefeng: nit - indentation off

yaojiefeng: Mark `recordPrefix` as `@Nullable` if it can be null.

wrangler-core/src/main/java/io/cdap/wrangler/utils/SchemaConverter.java
```diff
@@ -81,14 +84,27 @@ public Schema toSchema(String id, Row row) throws RecordConvertorException {
...
+  /**
+   * Generates the CDAP Schema for the given object
+   *
+   * @param value the value to retrieve schema
+   * @param name name of the field
+   * @param recordPrefix prefix to append at the beginning of a custom record
+   * @return the schema of this object
+   */
+  @Nullable
+  public Schema getSchema(Object value, String name, String recordPrefix) throws RecordConvertorException {
```
yaojiefeng: Mark `recordPrefix` as `@Nullable` if it can be null.

wrangler-core/src/main/java/io/cdap/wrangler/utils/StructuredToRowTransformer.java
```diff
@@ -77,14 +78,36 @@ public static Object getValue(StructuredRecord input, String fieldName) {
...
+    switch(schema.getType()) {
+      // if the type is bytes, need to make sure the value is byte array since byte buffer is not serializable
+      case BYTES:
+        return val instanceof ByteBuffer ? Bytes.toBytes((ByteBuffer) val) : val;
+      // Recursively process structured records.
+      case RECORD:
+        return val instanceof StructuredRecord ? transform((StructuredRecord) val) : val;
+      case ARRAY:
+        if (val instanceof Iterable) {
+          ArrayList<Object> rowList = new ArrayList<>();
```
yaojiefeng: `List<Object>`

wrangler-core/src/main/java/io/cdap/wrangler/utils/StructuredToRowTransformer.java
```diff
@@ -77,14 +78,36 @@ public static Object getValue(StructuredRecord input, String fieldName) {
...
-    // simple type (int/long).
-    return input.get(fieldName);
+  private static Object processValue(Object val, Schema schema) {
+    switch(schema.getType()) {
+      // if the type is bytes, need to make sure the value is byte array since byte buffer is not serializable
+      case BYTES:
+        return val instanceof ByteBuffer ? Bytes.toBytes((ByteBuffer) val) : val;
+      // Recursively process structured records.
+      case RECORD:
+        return val instanceof StructuredRecord ? transform((StructuredRecord) val) : val;
```
yaojiefeng: what other type can `val` have here? Should we throw exception if it is not a StructuredRecord?

wrangler-core/src/main/java/io/cdap/wrangler/utils/StructuredToRowTransformer.java
```diff
@@ -77,14 +78,36 @@ public static Object getValue(StructuredRecord input, String fieldName) {
...
+      case ARRAY:
+        if (val instanceof Iterable) {
```
yaojiefeng, line 101: what other type can val have here? Should we throw exception if it is not a `Iterable`?

wrangler-core/src/test/java/io/cdap/wrangler/utils/StructuredToRowTransformerTest.java
```diff
@@ -53,10 +56,20 @@ public void testStructuredRecordToRow() {
       Schema.Field.of("f11", Schema.of(Schema.LogicalType.TIME_MICROS)),
       Schema.Field.of("f12", Schema.of(Schema.LogicalType.TIME_MILLIS)),
       Schema.Field.of("f13", Schema.decimalOf(3, 2)),
-      Schema.Field.of("f14", Schema.of(Schema.LogicalType.DATETIME)));
+      Schema.Field.of("f14", Schema.of(Schema.LogicalType.DATETIME))
```
yaojiefeng: Should also add `Array` type to the record for unit test.

wrangler-service/src/main/java/io/cdap/wrangler/service/directive/AbstractDirectiveHandler.java
```diff
@@ -146,6 +159,76 @@ public void destroy() {
...
+          for (Pair<String, Object> field : value.getFields()) {
+            out.name(field.getFirst());
+            elementAdapter.write(out, field.getSecond());
+          }
+          out.endObject();
+        }
+
+        @Override public
+        Row read(JsonReader in) throws IOException {
+          throw new UnsupportedOperationException("Reading Rows from Wrangler display format not implemented");
```
yaojiefeng: `not implemented` -> `not supported`

wrangler-service/src/main/java/io/cdap/wrangler/service/directive/AbstractDirectiveHandler.java
```diff
@@ -176,16 +262,25 @@ protected DirectiveExecutionResponse generateExecutionResponse(
           String type = object.getClass().getSimpleName();
           if (schema != null) {
             schema = schema.isNullable() ? schema.getNonNullable() : schema;
-            type = schema.getLogicalType() == null ? schema.getType().name() : schema.getLogicalType().name();
+            type = schema.getLogicalType() == null ? schema.getType().name()
+                : schema.getLogicalType().name();
```
yaojiefeng: this doesn't need a new line

wrangler-service/src/main/java/io/cdap/wrangler/service/directive/AbstractDirectiveHandler.java
```diff
@@ -146,6 +159,76 @@ public void destroy() {
     }
   }
 
+
+  /**
+   * Formats Structure and Array records in a "pseudojson" in which it collapses
+   * complex types by callint toString().  This is done so the complex types are shown
+   * in a similar way inside and outside the structure records.
+   */
+  private static class WranglerDisplaySerializer implements TypeAdapterFactory {
```
yaojiefeng: better separate this to a new class and have unit tests for it

yaojiefeng (commented):
Took a first pass, have some comments

wrangler-core/src/main/java/io/cdap/directives/column/CreateRecord.java
```diff
@@ -0,0 +1,110 @@
...
+    List<Row> results = new ArrayList<>();
+
+    // Iterate through the rows.
+    for (Row row : rows) {
+
+      // We create a new Row with values from the columns specified.
+      Row newRecord = new Row();
+      for (String columnName : columns) {
+        Object columnValue = row.getValue(columnName);
+        // Only set value y column is set.
```
rmstar: y -> if ?

wrangler-core/src/main/java/io/cdap/directives/column/CreateRecord.java
```diff
@@ -0,0 +1,110 @@
...
+    builder.define("columns", TokenType.COLUMN_NAME_LIST);
+    return builder.build();
+  }
+
+  @Override
+  public void initialize(Arguments args) throws DirectiveParseException {
+    List<String> cols = ((ColumnNameList) args.value("columns")).value();
+    targetColumn = args.value("target_column").value().toString();
+    columns = new String[cols.size()];
+    columns = cols.toArray(columns);
```
rmstar: nit: can be simplified - columns = cols.toArray(new String[0])

wrangler-service/src/main/java/io/cdap/wrangler/service/directive/AbstractDirectiveHandler.java
```diff
@@ -156,6 +239,9 @@ protected DirectiveExecutionResponse generateExecutionResponse(
     Set<String> headers = new LinkedHashSet<>();
     SchemaConverter convertor = new SchemaConverter();
 
+
+    Gson gson = new GsonBuilder().registerTypeAdapterFactory(new WranglerDisplaySerializer()).create();
```
rmstar: Can we create a static Gson instance instead of creating it here?

wrangler-core/src/main/java/io/cdap/wrangler/utils/StructuredToRowTransformer.java
sechegaray: Added the exception.

wrangler-core/src/test/java/io/cdap/wrangler/utils/StructuredToRowTransformerTest.java
sechegaray: Added

wrangler-core/src/main/java/io/cdap/directives/column/CreateRecord.java
sechegaray: changed

wrangler-core/src/main/java/io/cdap/directives/column/ExplodeRecord.java
sechegaray: There's already a directive called `Flatten` that works for arrays. That directive takes a list of columns. Making a different syntax for two similar directives might confuse the users even more.

wrangler-core/src/main/java/io/cdap/directives/column/ExplodeRecord.java
sechegaray: Good Point

wrangler-core/src/main/java/io/cdap/wrangler/utils/RecordConvertor.java
sechegaray: A Nested Record, is converted to A Row inside a Row in Wrangler.  It can be that the nested record is nullable.

wrangler-core/src/main/java/io/cdap/directives/column/ExplodeRecord.java
sechegaray: done

wrangler-core/src/main/java/io/cdap/directives/column/ExplodeRecord.java
sechegaray: Talked with @bdmogal about this. We will document it as a known issue. The same issue occurs with other directives that generate new columns dynamically.

wrangler-core/src/main/java/io/cdap/directives/column/ExplodeRecord.java
```diff
@@ -0,0 +1,141 @@
...
+@Name(ExplodeRecord.NAME)
+@Categories(categories = { "column"})
+@Description("Flattens a record into individual columns.")
+public class ExplodeRecord implements Directive, Lineage {
+  public static final String NAME = "explode-record";
+  // Column on which to apply mask.
+  private String[] columns;
+
+  // Map to Keep Track new Columns;
+  private List<String> newColumnNames = new ArrayList<>();
```
yaojiefeng: Can just remove this if we do not care about lineage

wrangler-core/src/main/java/io/cdap/directives/column/ExplodeRecord.java
```diff
@@ -0,0 +1,141 @@
...
+          row.addOrSet(columnName, field.getSecond());
+        }
+      }
+      results.add(row);
+    }
+    return results;
+  }
+
+  @Override
+  public Mutation lineage() {
```
yaojiefeng: Don't extend this class from `Lineage` since it is impossible to emit lineage correctly for this directive, better not support it rather than giving wrong information

wrangler-core/src/main/java/io/cdap/directives/column/ExplodeRecord.java
yaojiefeng: I still think this should just operate on a single column, it doesn't make sense for this to operate on multiple columns, otherwise all the directives related to single column should be extended to be multiple columns. We shouldn't use old wrong behavior as tradition. With new implementation, we should make sure things are correct

wrangler-core/src/main/java/io/cdap/wrangler/utils/SchemaConverter.java
```diff
@@ -31,6 +32,7 @@
 import io.cdap.wrangler.api.Row;
 import org.json.JSONException;
 
+
```
yaojiefeng: nit - remove extra line

wrangler-core/src/test/java/io/cdap/directives/column/CreateRecordTest.java
```diff
@@ -0,0 +1,67 @@
+/*
+ *  Copyright © 2017-2019 Cask Data, Inc.
```
yaojiefeng: should be 2021

wrangler-core/src/test/java/io/cdap/directives/column/ExplodeRecordTest.java
```diff
@@ -0,0 +1,71 @@
+/*
+ *  Copyright © 2017-2019 Cask Data, Inc.
```
yaojiefeng: Should be 2021

wrangler-core/src/test/java/io/cdap/directives/column/ExplodeRecordTest.java
```diff
@@ -0,0 +1,71 @@
...
+import org.junit.Assert;
+import org.junit.Test;
+
+import java.util.Arrays;
+import java.util.List;
+
+/**
+ * Tests {@link ExplodeRecordTest}
+ */
+public class ExplodeRecordTest {
```
yaojiefeng: This test is too simple, should have a test for a more nested scenario.

wrangler-service/src/main/java/io/cdap/wrangler/service/directive/AbstractDirectiveHandler.java
```diff
@@ -66,6 +75,8 @@
 import java.util.Set;
 import java.util.stream.Collectors;
 
+
```
yaojiefeng: nit - remove extra line

wrangler-service/src/main/java/io/cdap/wrangler/service/directive/AbstractDirectiveHandler.java
```diff
@@ -17,6 +17,14 @@
 
 package io.cdap.wrangler.service.directive;
 
+import com.google.gson.Gson;
+import com.google.gson.GsonBuilder;
+import com.google.gson.TypeAdapter;
```
yaojiefeng: these should be unused now right?

wrangler-service/src/main/java/io/cdap/wrangler/service/directive/AbstractDirectiveHandler.java
```diff
@@ -146,6 +158,9 @@ public void destroy() {
     }
   }
 
+
+
```
yaojiefeng: nit - remove extra lines

wrangler-service/src/main/java/io/cdap/wrangler/service/directive/AbstractDirectiveHandler.java
```diff
@@ -181,11 +198,19 @@ protected DirectiveExecutionResponse generateExecutionResponse(
             type = type.substring(0, 1).toUpperCase() + type.substring(1).toLowerCase();
           }
           types.put(fieldName, type);
-          if ((object.getClass().getMethod("toString").getDeclaringClass() != Object.class)) {
-            value.put(fieldName, object.toString());
+
+          if ((object instanceof StructuredRecord)
```
yaojiefeng: under what condition can this object be `StructuredRecord`? It is not possible that it is still a StructuredRecord right?

wrangler-service/src/main/java/io/cdap/wrangler/service/directive/WranglerDisplaySerializer.java
```diff
@@ -0,0 +1,126 @@
...
+      @Override public
+      Row read(JsonReader in) throws IOException {
+        throw new UnsupportedOperationException("Reading Rows from Wrangler display format not supported.");
+      }
+    };
+  }
+
+
+  private TypeAdapter<Row> rowClassAdapter(Gson gson) {
+    final TypeAdapter<Object> elementAdapter = gson.getAdapter(Object.class);
```
yaojiefeng, line 55: can be private

wrangler-service/src/main/java/io/cdap/wrangler/service/directive/WranglerDisplaySerializer.java
```diff
@@ -0,0 +1,126 @@
...
+          } else {
+            out.value(NONDISPLAYABLE_STRING);
+          }
+        } catch (NoSuchMethodException e) {
+          e.printStackTrace();
+          out.value(NONDISPLAYABLE_STRING);
+        }
+      }
+
+
```
yaojiefeng: nit - remove extra line

wrangler-service/src/main/java/io/cdap/wrangler/service/directive/WranglerDisplaySerializer.java
```diff
@@ -0,0 +1,126 @@
...
+      @Override
+      public void write(JsonWriter out, Object value) throws IOException {
+        try {
+          if ((value.getClass().getMethod("toString").getDeclaringClass() != Object.class)) {
+            out.value(value.toString());
+          } else {
+            out.value(NONDISPLAYABLE_STRING);
+          }
+        } catch (NoSuchMethodException e) {
+          e.printStackTrace();
```
yaojiefeng: remove this print stack trace

wrangler-service/src/main/java/io/cdap/wrangler/service/directive/WranglerDisplaySerializer.java
```diff
@@ -0,0 +1,126 @@
...
+    // Let GSON Handle primitives and collections
+    if (ClassUtils.isPrimitiveOrWrapper(type.getRawType())
+      || (Iterable.class.isAssignableFrom(type.getRawType()))) {
+      return null;
+    }
+    if (Row.class.isAssignableFrom(type.getRawType())) {
+      return (TypeAdapter<T>) rowClassAdapter(gson);
+    }
+
+    if (Buffer.class.isAssignableFrom(type.getRawType())) {
```
yaojiefeng: What is the current behavior of dealing Buffer, are we showing the actual bytes value? This change will be backward incompatible.

wrangler-service/src/main/java/io/cdap/wrangler/service/directive/WranglerDisplaySerializer.java
```diff
@@ -0,0 +1,126 @@
...
+      return (TypeAdapter<T>) bufferClassAdapter(gson);
```
yaojiefeng: This `Buffer` handling and the default object handling can be put together in one method, 
the `if` condition can check whether this is a Buffer or something that doesn't have a `toString` method, this will make the handling clearer

yaojiefeng (commented):
Some more comments

wrangler-core/src/main/java/io/cdap/directives/column/ExplodeRecord.java
```diff
@@ -0,0 +1,141 @@
...
+ * followed by an underscore. Existing columns with the same name will be replaced by the new values.
+ *
+ * E.G
+ *
+ * Input:
+ * |  A |
+ * {B: 1, C: 2}
+ *
+ * Directive
+ * exploder-record :A
```
rmstar: nit: explode-record

wrangler-core/src/test/java/io/cdap/directives/column/CreateRecordTest.java
rmstar: 2021

wrangler-service/src/main/java/io/cdap/wrangler/service/directive/AbstractDirectiveHandler.java
rmstar: remove extra newlines?

wrangler-core/src/main/java/io/cdap/directives/column/ExplodeRecord.java
bdmogal: I think we should actually support multiple columns. The key is that we don't support many transformations on nested data. You *have* flatten them. So providing an option to do it in a single directive does seem more user friendly.

wrangler-core/src/main/java/io/cdap/directives/column/ExplodeRecord.java
```diff
@@ -0,0 +1,141 @@
...
+ * | A          | A_B | A_C |
+ * {B: 1, C:2}  | 1   |  2 |
+ *
+ */
+@Plugin(type = Directive.TYPE)
+@Name(ExplodeRecord.NAME)
+@Categories(categories = { "column"})
+@Description("Flattens a record into individual columns.")
+public class ExplodeRecord implements Directive, Lineage {
+  public static final String NAME = "explode-record";
```
bdmogal: I know we have an ``explode`` directive. However, can we start using ``flatten-record`` instead, just to make it a less aggressive name?

wrangler-service/src/main/java/io/cdap/wrangler/service/directive/WranglerDisplaySerializer.java
sechegaray: As mentioned offline. This mirrors current behavior.

wrangler-service/src/main/java/io/cdap/wrangler/service/directive/WranglerDisplaySerializer.java
sechegaray: Having as separate methods show a more explicit approach. Having the default check for different classes that we might or not support might make it more complex.
Here we are explicitly mentioning that we are addressing Rows and Buffers explicitly, and everything else gets a toString call (if they have one).  Hiding the functionality in the default method would make it less clear.

wrangler-core/src/main/java/io/cdap/directives/column/FlattenRecord.java
```diff
@@ -0,0 +1,125 @@
...
+@Description("Flattens a record into individual columns.")
+public class FlattenRecord implements Directive {
+  public static final String NAME = "flatten-record";
+  // Column on which to apply mask.
+  private String[] columns;
+
+  @Override
+  public UsageDefinition define() {
+    UsageDefinition.Builder builder = UsageDefinition.builder(NAME);
+    builder.define("column", TokenType.COLUMN_NAME_LIST);
```
yaojiefeng: call this `columns` if it is for multiple columns

wrangler-core/src/main/java/io/cdap/wrangler/utils/SchemaConverter.java
```diff
@@ -45,6 +46,7 @@
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.UUID;
```
yaojiefeng: nit - unused?

wrangler-service/src/main/java/io/cdap/wrangler/service/directive/AbstractDirectiveHandler.java
```diff
@@ -73,11 +76,14 @@
...
 public class AbstractDirectiveHandler extends AbstractWranglerHandler {
   private static final Logger LOG = LoggerFactory.getLogger(AbstractDirectiveHandler.class);
+  private static final Gson  GSON = new GsonBuilder().registerTypeAdapterFactory(
+    new WranglerDisplaySerializer()).create();
 
   protected static final String COLUMN_NAME = "body";
   protected static final String RECORD_DELIMITER_HEADER = "recorddelimiter";
   protected static final String DELIMITER_HEADER = "delimiter";
 
+
```
yaojiefeng: nit - remove extra line

wrangler-service/src/main/java/io/cdap/wrangler/service/directive/AbstractDirectiveHandler.java
```diff
@@ -156,6 +162,8 @@ protected DirectiveExecutionResponse generateExecutionResponse(
     Set<String> headers = new LinkedHashSet<>();
     SchemaConverter convertor = new SchemaConverter();
 
+
```
yaojiefeng: nit - remove extra line

wrangler-service/src/main/java/io/cdap/wrangler/service/directive/WranglerDisplaySerializer.java
yaojiefeng: The original code does not have any special handling of the Bytes type and still show it non-displayable, I don't think we need this check for special handling of Buffer, since all Buffer will be converted to byte[] in https://github.com/data-integrations/wrangler/blob/develop/wrangler-core/src/main/java/io/cdap/wrangler/utils/StructuredToRowTransformer.java#L81-L83. Let's remove this check and add it back when we support other Buffer type in the future.

yaojiefeng (commented):
Just minor comments

yaojiefeng (approved):
LGTM, please squash the commits into 1 before merging

PR #536, opened 2021-08-03
Remote directives execution optimization

chtyim:
- This PR contains multiple code cleanup as suggested by IDE (mostly about `final` fields, unused methods/classes, and logic simplification).
- The following optimizations related to remote directives execution are made:
   - Fixed the logic in `UserDirectiveRegistry` to avoid creating multiple classloaders of the same artifact.
   - Avoid calling DirectiveRegistry.reload inside the remote task. This is achieved by:
      1. Gather information about system directives before submitting to the remote task.
      1. Load the system directive class directly, instead of using reflection to query for all available system directives in the task (i.e. avoid creating `SystemDirectiveRegistry` in the task).
      1. Any non system directives are considered as UDD, and they are loaded directly via the plugin manager API, instead of looping through all artifacts at the beginning (e.g. avoid calling `UserDirectiveRegistry.reload` in the task).
  - Avoid task executor to restart if no UDD was executed
- In the process, also fixed a bug about the `#pragma` directive is not added when UDD is used.
With all these changes, the wrangler `RemoteExecutionTask` execution time was reduced from 2 seconds to 200ms (in a hot container).


wrangler-api/src/main/java/io/cdap/wrangler/api/DirectiveParseException.java
```diff
@@ -31,6 +31,11 @@ public DirectiveParseException(String message, Iterator<SyntaxError> errors) {
     this.errors = errors;
   }
 
+  public DirectiveParseException(String message, Iterator<SyntaxError> errors, Throwable cause) {
+    super(message, cause);
+    this.errors = errors;
```
tivv, line 36: nit: Wow, iterator as a field. Not something that introduced here, but pretty unsafe as it's single-use object

wrangler-api/src/main/java/io/cdap/wrangler/api/RecipePipeline.java
```diff
@@ -66,7 +59,8 @@
   /**
    * Destroys the pipeline.
    */
-  void destroy();
+  @Override
+  void close();
```
tivv, line 63: Do we need any backwards API compatibility here to support custom directives?

wrangler-core/src/main/java/io/cdap/wrangler/executor/RecipePipelineExecutor.java
```diff
@@ -41,45 +40,39 @@
...
+  public void close() {
+    if (directives == null) {
+      return;
+    }
+    for (Directive directive : directives) {
       try {
         directive.destroy();
       } catch (Throwable t) {
-        LOG.warn(t.getMessage());
+        LOG.warn(t.getMessage(), t);
```
tivv, line 75: nit: Make a message that explains where it happened

wrangler-core/src/main/java/io/cdap/wrangler/executor/RecipePipelineExecutor.java
```diff
@@ -164,4 +155,15 @@ public void destroy() {
   public List<ErrorRecord> errors() {
     return collector.get();
   }
+
+  private List<Directive> getDirectives() throws RecipeException {
+    if (directives == null) {
+      try {
+        this.directives = recipeParser.parse();
+      } catch (DirectiveParseException | DirectiveNotFoundException | DirectiveLoadException e) {
+        throw new RecipeException(e.getMessage(), e);
```
tivv, line 164: nit: Usually for exception without own message it's just `new RecipeException(e)` or similar.

wrangler-core/src/main/java/io/cdap/wrangler/parser/GrammarWalker.java
```diff
@@ -0,0 +1,89 @@
...
+import io.cdap.wrangler.api.DirectiveContext;
+import io.cdap.wrangler.api.DirectiveParseException;
+import io.cdap.wrangler.api.TokenGroup;
+import io.cdap.wrangler.api.parser.DirectiveName;
+import io.cdap.wrangler.api.parser.SyntaxError;
+
+import java.util.Iterator;
+
+/**
+ *
```
tivv: Please add a description

wrangler-core/src/main/java/io/cdap/wrangler/registry/DirectiveInfo.java
```diff
@@ -116,21 +109,30 @@ public boolean deprecated() {
...
   }
 
   /**
    * @return a <code>String</code> type containing the usage information of the directive.
    */
   public String usage() {
+    if (usage != null) {
+      return usage;
+    }
+    UsageDefinition definition = definition();
```
tivv, line 130: Please be consistent in using "synchronized". It's better to have whole class either thread-safe or not thread-safe

wrangler-core/src/main/java/io/cdap/wrangler/parser/ConfigDirectiveContext.java
```diff
@@ -16,52 +16,21 @@
...
  */
 public class ConfigDirectiveContext implements DirectiveContext {
-  private DirectiveConfig config;
+  private final DirectiveConfig config;
 
   public ConfigDirectiveContext(DirectiveConfig config) {
     this.config = config;
   }
 
-  public ConfigDirectiveContext(URL url) throws IOException {
```
greeshmaswaminathan, line 51: Was this unused?

wrangler-core/src/main/java/io/cdap/wrangler/registry/SystemDirectiveRegistry.java
```diff
@@ -107,7 +122,7 @@ public ArtifactSummary getLatestWranglerArtifact() {
    */
   @Override
   public Iterable<DirectiveInfo> list(String namespace) {
-    return registry.values();
+    return Collections.unmodifiableCollection(registry.values());
```
tivv, line 138: nit: make it immutable in constructor.

wrangler-service/src/main/java/io/cdap/wrangler/service/directive/AbstractDirectiveHandler.java
```diff
@@ -85,16 +100,57 @@ public void destroy() {
...
+      Set<String> userDirectives = new HashSet<>();
+      GrammarWalker walker = new GrammarWalker(new RecipeCompiler(), new ConfigDirectiveContext(DirectiveConfig.EMPTY));
+      walker.walk(recipe, (command, tokenGroup) -> {
+        DirectiveInfo info = SystemDirectiveRegistry.INSTANCE.get(namespace, command);
+        if (info == null) {
+          userDirectives.add(command);
+        }
+      });
+      if (!userDirectives.isEmpty()) {
+        directives.add(0, "#pragma load-directives " + String.join(",", userDirectives) + ";");
```
tivv: Where is this guy used? I see you create a `recipe` above, but I don't see any `directives` usage below. Is it an expected side effect?

wrangler-service/src/main/java/io/cdap/wrangler/service/directive/DirectivesHandler.java
```diff
@@ -565,11 +552,11 @@ public void execute(HttpServiceRequest request, HttpServiceResponder responder,
         if (directiveRequest == null) {
           throw new BadRequestException("Request body is empty.");
         }
-        directiveRequest.getRecipe().setPragma(addLoadablePragmaDirectives(namespace, directiveRequest));
+        List<String> directives = new LinkedList<>(directiveRequest.getRecipe().getDirectives());
```
tivv: Any specific reason to use a linked list? It creates a lot of objects and generally slower than array list. I think even if you do a single `.add(0)` array list would still be faster as it adds just a single copy. Another options is to use `ArrayDeque` that allows to add to the head without extra copy (assuming enough capacity)

wrangler-service/src/main/java/io/cdap/wrangler/service/directive/RemoteExecutionTask.java
```diff
@@ -19,43 +19,109 @@
...
+      if (directives.isEmpty()) {
+        runnableTaskContext.writeResult(directiveRequest.getData());
+        return;
+      }
+
       ObjectSerDe<List<Row>> objectSerDe = new ObjectSerDe<>();
-      List<Row> inputRows = objectSerDe.toObject(directiveRequest.getData());
-      List<Row> rows = commonDirectiveExecutor
-        .executeDirectives(namespace, directiveRequest.getDirectives(), inputRows);
+      List<Row> rows = objectSerDe.toObject(directiveRequest.getData());
```
tivv, line 103: I am wondering if we should just use https://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/SerializationUtils.html

wrangler-api/src/main/java/io/cdap/wrangler/api/DirectiveParseException.java
chtyim: Yes. It was pretty badly defined. Probably need another round of cleanup

wrangler-api/src/main/java/io/cdap/wrangler/api/RecipePipeline.java
chtyim: This is actually internally to Wrangler, as custom directives shouldn't be using recipe pipeline at all.

wrangler-core/src/main/java/io/cdap/wrangler/executor/RecipePipelineExecutor.java
chtyim: There are a lot of places in Wrangler that is like this. Probably do another round of code cleanup.

wrangler-core/src/main/java/io/cdap/wrangler/executor/RecipePipelineExecutor.java
chtyim: There are a lot of places like this in Wrangler. For this particular case, I think it wants the message of the exception to not carrying the exception class name (if we do `new RecipeException(e)`, the default logic in `Throwable` is to use the `e.toString()` as the message, which the default implementation would insert the class name there). And the reason for that is probably because the message is going to be shown in the UI.

wrangler-core/src/main/java/io/cdap/wrangler/registry/DirectiveInfo.java
chtyim: It is still thread safe without using sychronized, just that it might generate the usage more than once, which is ok.

wrangler-core/src/main/java/io/cdap/wrangler/registry/SystemDirectiveRegistry.java
chtyim: sounds good

wrangler-core/src/main/java/io/cdap/wrangler/parser/GrammarWalker.java
```diff
@@ -0,0 +1,89 @@
...
+      }
+      String command = ((DirectiveName) tokenGroup.get(0)).value();
+      String root = command;
+      if (context.hasAlias(root)) {
+        root = context.getAlias(command);
+      }
+
+      // Checks if the directive has been excluded from being used.
+      if (!root.equals(command) && context.isExcluded(command)) {
+        throw new DirectiveParseException(
```
greeshmaswaminathan, line 84: Nit: could combine both checks with || . If we need specific message for alias, use root in the message.

wrangler-service/src/main/java/io/cdap/wrangler/service/directive/AbstractDirectiveHandler.java
chtyim: It is carrying from the old logic that was missed when porting from the old DirectiveHandler to the new WorkspaceHandler.

It is needed for loading UDD.

wrangler-core/src/main/java/io/cdap/wrangler/parser/GrammarWalker.java
```diff
@@ -0,0 +1,89 @@
...
+    if (!status.isSuccess()) {
+      Iterator<SyntaxError> errors = status.getErrors();
+      String prefix = "Encountered syntax error, please ensure the directive is valid:\n";
+      throw new DirectiveParseException(prefix + errors.next().getMessage(), errors);
+    }
+
+    Iterator<TokenGroup> tokenGroups = status.getSymbols().iterator();
+    while (tokenGroups.hasNext()) {
+      TokenGroup tokenGroup = tokenGroups.next();
+      if (tokenGroup == null) {
```
greeshmaswaminathan, line 73: Is this a valid case ?

wrangler-core/src/main/java/io/cdap/wrangler/parser/GrammarWalker.java
chtyim: added

wrangler-service/src/main/java/io/cdap/wrangler/service/directive/DirectivesHandler.java
chtyim: It was mainly for the add(0). BTW, the list is quite short (list of directives), so using anything won't make much difference

wrangler-service/src/main/java/io/cdap/wrangler/service/directive/RemoteExecutionTask.java
chtyim: Firstly, I was not intended to change existing logic. Secondly, based on actual result, it seems not too slow. Lastly, I am a bit reluctant to use apache common library.

wrangler-core/src/main/java/io/cdap/wrangler/parser/ConfigDirectiveContext.java
chtyim: Based on IDE, yes it is unused.

wrangler-core/src/main/java/io/cdap/wrangler/parser/GrammarWalker.java
chtyim: It was from existing logic, so I supposed it can be?

wrangler-core/src/main/java/io/cdap/wrangler/parser/GrammarWalker.java
chtyim: I think combining the two `if` actually make the code less readable.

wrangler-service/src/main/java/io/cdap/wrangler/service/directive/WorkspaceHandler.java
```diff
@@ -415,4 +409,79 @@ private String generateWorkspaceName(WorkspaceId id, @Nullable String path) {
...
+        userDirectives.add(command);
+      }
+    });
+    if (!userDirectives.isEmpty()) {
+      directives.add(0, "#pragma load-directives " + String.join(",", userDirectives) + ";");
+    }
+
+    RemoteDirectiveRequest directiveRequest = new RemoteDirectiveRequest(recipe, systemDirectives,
+                                                                         namespace, detail.getSampleAsBytes());
+    RunnableTaskRequest runnableTaskRequest = RunnableTaskRequest.getBuilder(RemoteExecutionTask.class.getName())
```
greeshmaswaminathan, line 498: Will it help to send the list of commands in RemoteDirectiveRequest? Since we are already compiling and walking the recipe, looks like we have all the information already required for the directives here. This will eliminate the need for using GrammarWalker again in remote execution.

wrangler-service/src/main/java/io/cdap/wrangler/service/directive/AbstractDirectiveHandler.java
tivv: Yes, I understand the intent. Could you point the method this new directive is going to be used? It's not used in this method as far as I can see.

greeshmaswaminathan: In the v2 API this will be used in WorkspaceHandler#specification()

tivv: Got you. Are there a reason it's added here? Such side effects are very easy to miss while refactoring. Also a very similar code is now present in two places. Can it be a separate method explicitly called from WorkspaceHandler#specification?

wrangler-service/src/main/java/io/cdap/wrangler/service/directive/WorkspaceHandler.java
chtyim: I tried to carry the parsed result over to the remote task, but turns out it is quite complex to serialize the `Token` interface, which has implementations for various type. The tree walking time is actually fast comparing to other operations (10s of milliseconds only).

chtyim:
This PR needs a new API introduced in https://github.com/cdapio/cdap/pull/13604

wrangler-service/src/main/java/io/cdap/wrangler/service/directive/AbstractDirectiveHandler.java
greeshmaswaminathan: We need parsing of recipe to find the user defined directive names that has to be appended to #pragma load-directives. This is automatically added only when UDD is being used. Separate method for this will involve parsing it again. Best I can think of to remove it as a side effect is to modify the return type for executeDirectives() to return both List<Row> and pragma string/udd names and let the caller add it to directives. @chtyim , thoughts?

wrangler-service/src/main/java/io/cdap/wrangler/service/directive/WorkspaceHandler.java
greeshmaswaminathan: Sounds good.

wrangler-service/src/main/java/io/cdap/wrangler/service/directive/DirectivesHandler.java
```diff
@@ -1068,14 +1071,30 @@ public void getConfig(HttpServiceRequest request, HttpServiceResponder responder
...
+   * Executes directives by extracting them from request.
+   *
+   * @param id data to be used for executing directives.
+   * @param directives the list of directives to execute
+   * @param sample sampling function.
+   * @param grammarVisitor visitor to call while parsing directives
+   * @return records generated from the directives.
+   */
+  private <E extends Exception> List<Row> executeDirectives(NamespacedId id, List<String> directives,
+      Function<List<Row>, List<Row>> sample,
```
chtyim: Align with

```
private <E extends Exception> List<Row> executeDirectives(NamespacedId id, List<String> directives,
                                                          Function<List<Row>, List<Row>> sample,
                                                          Visitor<E> grammarVisitor) {
```

wrangler-service/src/main/java/io/cdap/wrangler/service/directive/WorkspaceHandler.java
```diff
@@ -416,16 +421,20 @@ private String generateWorkspaceName(WorkspaceId id, @Nullable String path) {
    * @param directives the list of directives to apply. The list provided must be a mutable list for the addition of
    *                   {@code #pragma} directives for loading UDDs.
    * @param detail the workspace to operate on
+   * @param grammarVisitor visitor to call while parsing directives
    * @return the resulting rows after applying the directives
    */
-  private List<Row> executeDirectives(String namespace, List<String> directives,
-                                      WorkspaceDetail detail) throws Exception {
+  private <E extends Exception> List<Row> executeDirectives(
+      String namespace,
```
chtyim: Make the alignment as 
```
private <E extends Exception> List<Row> executeDirectives(String namespace,
                                                          List<String> directives,
```

wrangler-service/src/main/java/io/cdap/wrangler/service/directive/DirectivesHandler.java
```diff
@@ -1068,14 +1071,30 @@ public void getConfig(HttpServiceRequest request, HttpServiceResponder responder
...
+      Visitor<E> grammarVisitor) {
     return TransactionRunners.run(getContext(), ctx -> {
       WorkspaceDataset ws = WorkspaceDataset.get(ctx);
 
       Workspace workspace = ws.getWorkspace(id);
       // Extract rows from the workspace.
       List<Row> rows = fromWorkspace(workspace);
-      return executeDirectives(id.getNamespace().getName(), directives, sample.apply(rows));
+      return executeDirectives(id.getNamespace().getName(), directives, sample.apply(rows),
+          grammarVisitor);
```
chtyim: Align with the previous line as

```
return executeDirectives(id.getNamespace().getName(), directives, sample.apply(rows),
                         grammarVisitor);
```

wrangler-service/src/main/java/io/cdap/wrangler/service/directive/DirectivesHandler.java
```diff
@@ -1068,14 +1071,30 @@ public void getConfig(HttpServiceRequest request, HttpServiceResponder responder
    * @return records generated from the directives.
    */
   private List<Row> executeDirectives(NamespacedId id, List<String> directives,
-                                      Function<List<Row>, List<Row>> sample) {
+                                                            Function<List<Row>, List<Row>> sample) {
```
chtyim: Alignment

wrangler-service/src/main/java/io/cdap/wrangler/service/directive/WorkspaceHandler.java
```diff
@@ -338,7 +340,10 @@ public void specification(HttpServiceRequest request, HttpServiceResponder respo
       WorkspaceId wsId = new WorkspaceId(ns, workspaceId);
       WorkspaceDetail detail = wsStore.getWorkspaceDetail(wsId);
       List<String> directives = new ArrayList<>(detail.getWorkspace().getDirectives());
-      List<Row> result = executeDirectives(ns.getName(), directives, detail);
+      UserDirectivesCollector userDirectivesCollector = new UserDirectivesCollector();
+      List<Row> result = executeDirectives(ns.getName(), directives, detail,
+          userDirectivesCollector);
```
chtyim: Alignment

wrangler-service/src/main/java/io/cdap/wrangler/service/directive/WorkspaceHandler.java
```diff
@@ -435,13 +444,17 @@ private String generateWorkspaceName(WorkspaceId id, @Nullable String path) {
...
-                                   WorkspaceDetail detail) throws DirectiveLoadException, DirectiveParseException {
+  private <E extends Exception> List<Row> executeLocally(String namespace, List<String> directives,
+                                   WorkspaceDetail detail, GrammarWalker.Visitor<E> grammarVisitor)
+      throws DirectiveLoadException, DirectiveParseException, E {
+
     // load the udd
     composite.reload(namespace);
-    return executeDirectives(namespace, directives, new ArrayList<>(detail.getSample()));
+    return executeDirectives(namespace, directives, new ArrayList<>(detail.getSample()),
+        grammarVisitor);
```
chtyim: Alignment

wrangler-service/src/main/java/io/cdap/wrangler/service/directive/DirectivesHandler.java
tivv: Looks like my IDE lost code style settings. Reapplied and reformatted. Done

wrangler-service/src/main/java/io/cdap/wrangler/service/directive/WorkspaceHandler.java
tivv: Done

wrangler-service/src/main/java/io/cdap/wrangler/service/directive/DirectivesHandler.java
tivv: Done

wrangler-service/src/main/java/io/cdap/wrangler/service/directive/DirectivesHandler.java
tivv: Done

wrangler-service/src/main/java/io/cdap/wrangler/service/directive/WorkspaceHandler.java
tivv: Done

wrangler-service/src/main/java/io/cdap/wrangler/service/directive/WorkspaceHandler.java
tivv: Done

tivv (approved):
I am good for the changes, but it would be great for someone else to approve as well as some of the changes are done by me

greeshmaswaminathan (approved):

greeshmaswaminathan (approved):

PR #537, opened 2021-08-05
Added fix to avoid trying to get a column that is over the number of columns

sechegaray:


tivv (approved):
Thanks! Please make a cherry-pick into 6.5

PR #542, opened 2021-08-12
Let the db connection use old way to generate schema and spec

yaojiefeng:
This change is needed because the upgraded db workspaces will have different behavior depending on whether the database type supports schema:
1. For mysql, it will work and source schema is generated since it does not have schema, so the given path with just a table in that will generate the schema
2. For SQLServer, it will not generate the source schema, and getSpec call might fail because the path we give is a table name, and db connector will consider it as schema. 

To avoid such confusion, for db types, we use old ways to generate the spec.

### DIFF
--- a/wrangler-service/src/main/java/io/cdap/wrangler/service/directive/WorkspaceUpgrader.java
+++ b/wrangler-service/src/main/java/io/cdap/wrangler/service/directive/WorkspaceUpgrader.java
@@ -18,16 +18,22 @@
 package io.cdap.wrangler.service.directive;
 
 import io.cdap.cdap.api.NamespaceSummary;
+import io.cdap.cdap.api.data.schema.Schema;
 import io.cdap.cdap.api.service.SystemServiceContext;
 import io.cdap.cdap.etl.proto.ArtifactSelectorConfig;
 import io.cdap.cdap.etl.proto.connection.ConnectorDetail;
+import io.cdap.cdap.etl.proto.connection.PluginDetail;
 import io.cdap.cdap.etl.proto.connection.SpecGenerationRequest;
 import io.cdap.cdap.spi.data.transaction.TransactionRunners;
 import io.cdap.wrangler.PropertyIds;
 import io.cdap.wrangler.api.Row;
+import io.cdap.wrangler.dataset.connections.ConnectionStore;
 import io.cdap.wrangler.dataset.workspace.Workspace;
 import io.cdap.wrangler.dataset.workspace.WorkspaceDataset;
+import io.cdap.wrangler.proto.Namespace;
+import io.cdap.wrangler.proto.NamespacedId;
 import io.cdap.wrangler.proto.NotFoundException;
+import io.cdap.wrangler.proto.connection.Connection;
 import io.cdap.wrangler.proto.connection.ConnectionType;
 import io.cdap.wrangler.proto.workspace.v2.Artifact;
 import io.cdap.wrangler.proto.workspace.v2.Plugin;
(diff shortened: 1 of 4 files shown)


wrangler-service/src/main/java/io/cdap/wrangler/service/directive/WorkspaceUpgrader.java
```diff
@@ -129,15 +140,46 @@ private void upgradeWorkspacesInConnections(NamespaceSummary namespace) {
...
+          } catch (RecordConvertorException e) {
+            LOG.warn("Unable to get the source schema for workspace {}, the generated spec will not contain schema.",
+                     workspace.getName());
+          }
+          Connection conn = TransactionRunners.run(context, ctx -> {
+            ConnectionStore connStore = ConnectionStore.get(ctx);
+            return connStore.get(new NamespacedId(new Namespace(namespace.getName(), namespace.getGeneration()),
+                                                  connection));
+          });
+         dbProperties = DatabaseHandler.getSpecification(conn, workspace.getName());
```
albertshau: indentation

wrangler-service/src/main/java/io/cdap/wrangler/service/directive/WorkspaceUpgrader.java
```diff
@@ -129,15 +140,46 @@ private void upgradeWorkspacesInConnections(NamespaceSummary namespace) {
...
+        //    schema and the original workspace does not have this information. To make sure, the path does not mess up,
+        //    the above path for db connection will always be "/".
+        // 3. Using the old properties will ensure the old behavior get reserved for the db connectors
+        Schema dbSchema = null;
+        Map<String, String> dbProperties = Collections.emptyMap();
+        if (connectionType.equals(ConnectionType.DATABASE)) {
+          SchemaConverter schemaConvertor = new SchemaConverter();
+          try {
+            dbSchema = sample.isEmpty() ? null :
+                         schemaConvertor.toSchema("record", AbstractDirectiveHandler.createUberRecord(sample));
```
albertshau, line 158: is this calculating the schema from the workspace sample? Isn't that data transformed by whatever directives exist for the workspace? If so, it won't be the db schema.

albertshau (commented):
one question

wrangler-service/src/main/java/io/cdap/wrangler/service/directive/WorkspaceUpgrader.java
yaojiefeng: The initial sample will be the original data read from the database table. The directives are applied each time on the /execute call on this original data. The workspace will never store the executed result to make sure the directive can be removed. 
The old UI has this logic to get the specification for the pipeline:
1. call /specification in DatabaseHandler to get the properties(without schema) back
2. call /schema with an empty directive to get the source schema
3. call /schema with the directive applied to get the wrangler schema

wrangler-service/src/main/java/io/cdap/wrangler/service/directive/WorkspaceUpgrader.java
yaojiefeng: Fixed

wrangler-service/src/main/java/io/cdap/wrangler/service/directive/WorkspaceUpgrader.java
albertshau: ok sounds good.

albertshau (approved):

PR #546, opened 2021-08-15
increase url connection read timeout

yaojiefeng:
Noticed the sample request get timed out due to debugging using break point with for just several seconds, noticed the default read time out of url is 15 seconds, increase it to 2min,


wrangler-service/src/main/java/io/cdap/wrangler/service/directive/ConnectionDiscoverer.java
```diff
@@ -64,6 +64,7 @@
   private static final long RETRY_MAX_DELAY_MILLIS = TimeUnit.SECONDS.toMillis(5);
   private static final double RETRY_DELAY_MULTIPLIER = 1.2d;
   private static final double RETRY_RANDOMIZE_FACTOR = 0.1d;
+  private static final int URL_READ_TIMEOUT_MILLIS = 120000;
```
albertshau, line 67: this is quite long, can you make it configurable instead of hardcoded?

albertshau (approved):

PR #568, opened 2022-07-01
Add 'Average' Arithmetic Operation

adinilfeld:
# Add 'Average' Arithmetic Operation

## Description
Added the `average` method to `ArithmeticOperations.java`. It can now be used for directives in Wrangler. In particular, PR #548 in cdap-ui ([here](https://github.com/cdapio/cdap-ui/pull/548#issue-1291823927)) relies on this one.

Also made some minor comment formatting fixes (scroll to the bottom of the file to see the `average` function).

## PR Type
- [ ] Bug Fix
- [ ] Feature
- [ ] Build Fix
- [ ] Testing
- [x] General Improvement
- [ ] Cherry Pick

## Links
Jira: [PLUGIN-641](https://cdap.atlassian.net/browse/PLUGIN-641)

### DIFF
--- a/wrangler-core/src/main/java/io/cdap/wrangler/utils/ArithmeticOperations.java
+++ b/wrangler-core/src/main/java/io/cdap/wrangler/utils/ArithmeticOperations.java
@@ -14,15 +14,18 @@
  *  the License.
  */
 package io.cdap.wrangler.utils;
+import io.cdap.wrangler.api.DirectiveExecutionException;
+
 import java.math.BigDecimal;
 import java.math.BigInteger;
+import java.math.RoundingMode;
 
 /**
  * Cross-column operations for arithmetic
  */
 public class ArithmeticOperations {
   /**
-   Arithmetic operation - Add two columns.
+   * Arithmetic operation - Add two columns.
    */
   public static Integer add(Integer x, Integer y) {
     if (x == null || y == null) {
@@ -32,7 +35,7 @@ public class ArithmeticOperations {
   }
 
   /**
-   Arithmetic operation - Add two columns.
+   * Arithmetic operation - Add two columns.
    */
   public static Double add(Double x, Double y) {
     if (x == null || y == null) {
@@ -42,7 +45,7 @@ public class ArithmeticOperations {
   }
 
   /**
-   Arithmetic operation - Add two columns.
+   * Arithmetic operation - Add two columns.
    */
   public static Float add(Float x, Float y) {
     if (x == null || y == null) {
@@ -52,7 +55,7 @@ public class ArithmeticOperations {
   }
 
   /**
-   Arithmetic operation - Add two columns.
+   * Arithmetic operation - Add two columns.
    */
   public static BigDecimal add(BigDecimal x, BigDecimal y) {
     if (x == null || y == null) {
@@ -62,7 +65,7 @@ public class ArithmeticOperations {
   }
 
   /**
-   Arithmetic operation - Subtract a column from another column.
+   * Arithmetic operation - Subtract a column from another column.
    */
   public static Integer minus(Integer x, Integer y) {
     if (x == null || y == null) {
@@ -72,7 +75,7 @@ public class ArithmeticOperations {
   }
 
   /**
-   Arithmetic operation - Subtract a column from another column.
+   * Arithmetic operation - Subtract a column from another column.
    */
   public static Double minus(Double x, Double y) {
     if (x == null || y == null) {
@@ -82,7 +85,7 @@ public class ArithmeticOperations {
   }
 
   /**
-   Arithmetic operation - Subtract a column from another column.
+   * Arithmetic operation - Subtract a column from another column.
    */
   public static Float minus(Float x, Float y) {
     if (x == null || y == null) {
@@ -92,7 +95,7 @@ public class ArithmeticOperations {
   }
 
   /**
-   Arithmetic operation - Subtract a column from another column.
+   * Arithmetic operation - Subtract a column from another column.
    */
   public static BigDecimal minus(BigDecimal x, BigDecimal y) {
     if (x == null || y == null) {
@@ -112,7 +115,7 @@ public class ArithmeticOperations {
   }
 
   /**
-   Arithmetic operation - Multiply a column with another column.
+   * Arithmetic operation - Multiply a column with another column.
    */
   public static Double multiply(Double x, Double y) {
     if (x == null || y == null) {
@@ -122,7 +125,7 @@ public class ArithmeticOperations {
   }
 
   /**
-   Arithmetic operation - Multiply a column with another column.
+   * Arithmetic operation - Multiply a column with another column.
    */
   public static Float multiply(Float x, Float y) {
     if (x == null || y == null) {
@@ -132,7 +135,7 @@ public class ArithmeticOperations {
   }
 
   /**
-   Arithmetic operation - Multiply a column with another column.
+   * Arithmetic operation - Multiply a column with another column.
    */
   public static BigDecimal multiply(BigDecimal x, BigDecimal y) {
     if (x == null || y == null) {
@@ -142,7 +145,7 @@ public class ArithmeticOperations {
   }
 
   /**
-   Arithmetic operation - Divide a column by another and return the quotient.
+   * Arithmetic operation - Divide a column by another and return the quotient.
    */
   public static Integer divideq(Integer x, Integer y) {
     if (x == null || y == null || y == 0) {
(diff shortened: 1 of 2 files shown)


wrangler-core/src/main/java/io/cdap/wrangler/utils/ArithmeticOperations.java
```diff
@@ -388,12 +390,41 @@ public static Float min(Float f1, Float f2) {
...
+    for (Number num : nums) {
+      // Check for null input
+      if (num == null) {
+        continue;
+      } else {
+        allNull = false;
+      }
+      // Check for invalid input
+      if (num instanceof BigDecimal || num instanceof Byte) {
+        throw new DirectiveExecutionException("Input cannot be of type 'Decimal' or 'Byte'.");
```
fernst: Is there any reason why we decided to skip Decimal support from this function?

I understand there may be precision loss when casting to Double, but it seems arbitrary to not support the average function on Decimal columns

fernst (approved):
Asked a question about BigDecimal support. Code looks good to me.

fernst:
Please squash all 21 commits into 1 before merging

fernst (changes requested):
Please squash changes before merging

wrangler-core/src/main/java/io/cdap/wrangler/utils/ArithmeticOperations.java
ellliao: We originally skipped Decimal support because it appeared not to be supported for other arithmetic operations. It later became clear that there was indeed support for Decimal columns in the backend, it just was not reflected in the UI.

I have just pushed a commit adding Decimal support to the average operation; thank you for the feedback!

wrangler-core/src/main/java/io/cdap/wrangler/utils/ArithmeticOperations.java
```diff
@@ -388,12 +391,55 @@ public static Float min(Float f1, Float f2) {
...
+        continue;
+      } else {
+        allNull = false;
+      }
+      // Check for invalid input
+      if (num instanceof Byte) {
+        throw new DirectiveExecutionException("Input cannot be of type 'Byte'.");
+      }
+      // Compute average
+      BigDecimal numer;
```
fernst: Doing BigDecimal arithmetics is a very resource intensive operation, Can you try to run a few benchmarks with many (100k) input values and see how BigDecimal vs double arithmetics work?

fernst (changes requested):

wrangler-core/src/main/java/io/cdap/wrangler/utils/ArithmeticOperations.java
ellliao: We have just changed the function to perform double operations instead of BigDecimal ones unless there are BigDecimal inputs. As the existence of BigDecimal inputs makes BigDecimal operations unavoidable, should we still try running benchmarks?

fernst: I think having an idea of the performance would be useful, so yes.

ellliao: Ok, thanks for the help!

wrangler-core/src/main/java/io/cdap/wrangler/utils/ArithmeticOperations.java
```diff
@@ -388,12 +391,64 @@ public static Float min(Float f1, Float f2) {
...
+        continue;
+      } else {
+        allNull = false;
+      }
+      // Check for invalid input
+      if (num instanceof Byte) {
+        throw new DirectiveExecutionException("Input cannot be of type 'Byte'.");
+      }
+      // Compute average (no BigDecimal case)
+      if (noBigDecimal && num instanceof BigDecimal) {
```
fernst: This logic is a bit confusing.

As far as I see it, either we are doing BigDecimal arithmetics or we are not. As soon as you find one BigDecimal you switch into "BigDecimal mode" and then all calculations are performed over BigDecimals.

I would reorder this as:

1. have a variable that tracks if you have found a BigDecimal (similar to noBigDecimal)
2. Check if the current value is an instance of BigDecimal. If true, go into BigDecimal arithmetics mode (more precise). if false, use double arithmetics (faster).

I think it's also important to point out that the output schema is going to be different in each case. As soon as you enter "BigDecimal" mode, the output should be a decimal. If not, it should be double.

wrangler-core/src/main/java/io/cdap/wrangler/utils/ArithmeticOperations.java
```diff
@@ -388,12 +391,64 @@ public static Float min(Float f1, Float f2) {
...
+        if (t > 1) {
+          avg = BigDecimal.valueOf((double) avg);
+        } else {
+          avg = BigDecimal.ZERO;
+        }
+        noBigDecimal = false;
+      } else if (noBigDecimal) {
```
fernst: I think this statement makes more sense if it was the `else` of the next `if` block. This makes it obvious that these are the 2 paths the code takes (BigDecimal mode vs Double mode)

fernst (changes requested):

wrangler-core/src/main/java/io/cdap/wrangler/utils/ArithmeticOperations.java
ellliao: > As far as I see it, either we are doing BigDecimal arithmetics or we are not. As soon as you find one BigDecimal you switch into "BigDecimal mode" and then all calculations are performed over BigDecimals.

Yes, that's the idea

I'm not quite sure what you are saying -- are you recommending that we create two other functions (one for BigDecimal, one for double) to use depending on the situation?

wrangler-core/src/main/java/io/cdap/wrangler/utils/ArithmeticOperations.java
ellliao: Makes sense. Let me change that.

ellliao: Changed. Thanks!

wrangler-core/src/main/java/io/cdap/wrangler/utils/ArithmeticOperations.java
```diff
@@ -422,18 +422,16 @@ public static Number average(Number... nums) throws DirectiveExecutionException
...
         } else {
           avg = BigDecimal.ZERO;
         }
         noBigDecimal = false;
-      } else if (noBigDecimal) {
-        avg = (double) avg + (num.doubleValue() - (double) avg) / t;
       }
-      // Compute average (BigDecimal case)
+      // Compute average
       if (!noBigDecimal) {
```
fernst: nitpick: not...no bigDecimal? Can we change this variable name to `isBigDecimal` or `containsBigDecimal` or something like this?

ellliao: Sure, sounds good. I'll be changing the variable name to containsBigDecimal.

ellliao: Done. Thank you!

wrangler-core/src/main/java/io/cdap/wrangler/utils/ArithmeticOperations.java
ellliao: **Benchmark Summary:** Conducted benchmark tests using CDF batch pipeline with 3 CSVs containing 2 columns of randomly generated doubles from 0 to 1, with 100,000 entries in each.

**Results:**
double: 25, 23, 22 sec
decimal: 26, 24, 22 sec

**Notes:**
1. Not all decimal values were written to the output file, as decimal schema requires specification of scale and precision, causing errors in trying to write to output file as scale and precision varies (skipped errors for benchmark). This may have taken away a slight amount of time.
2. Decimal computation required column conversion to decimal type prior to average computation, which may have added a slight amount of time.

**Conclusion:** Times appear to be similar for decimal and double computations at this scale.

wrangler-core/src/main/java/io/cdap/wrangler/utils/ArithmeticOperations.java
adinilfeld: The logic should now be more clear after Ellen's most recent updates.

fernst (approved):
LGTM

sechegaray (approved):

PR #569, opened 2022-07-18
Expand Calculate Options

ellliao:
# Expand Calculate Options

## Description
Consolidated most functions (except `equal`) in `ArithmeticOperations.java` to take in any supported numeric type. Also expanded  some functions (`add`, `multiply`, `equal`, `max`, `min`) to take in >2 inputs/columns. PR [#565](https://github.com/cdapio/cdap-ui/pull/565) in [cdap-ui](https://github.com/cdapio/cdap-ui) depends on this one.

Also added unit tests for `ArithmeticOperations.java`.

## PR Type
- [ ] Bug Fix
- [x] Feature
- [ ] Build Fix
- [ ] Testing
- [ ] General Improvement
- [ ] Cherry Pick

## Links
Jira: none

### DIFF
--- a/wrangler-core/src/main/java/io/cdap/wrangler/utils/ArithmeticOperations.java
+++ b/wrangler-core/src/main/java/io/cdap/wrangler/utils/ArithmeticOperations.java
@@ -24,380 +24,483 @@ import java.math.RoundingMode;
  * Cross-column operations for arithmetic
  */
 public class ArithmeticOperations {
-  /**
-   * Arithmetic operation - Add two columns.
-   */
-  public static Integer add(Integer x, Integer y) {
-    if (x == null || y == null) {
-      return null;
+  // The Number types upon which arithmetic operations can be performed, in order of increasing generality
+  private enum NumberType {
+    SHORT(0),
+    INTEGER(1),
+    LONG(2),
+    FLOAT(3),
+    DOUBLE(4),
+    BIGDECIMAL(5);
+
+    private final int priority;
+
+    NumberType(int priority) {
+      this.priority = priority;
     }
-    return x + y;
+
+    public int getPriority() {
+      return priority;
+    }
   }
 
   /**
-   * Arithmetic operation - Add two columns.
+   * Return the "largest" (e.g. most general) type of the given numbers.
+   * Used for operations applied to columns of multiple types, in order to determine the type of the output column.
+   * The order of increasing generality is Short < Integer < Long < Float < Double < BigDecimal.
+   * Returns null if any input is null.
+   * Throws an error if any input of type Byte is given (this is not currently supported for arithmetic ops).
    */
-  public static Double add(Double x, Double y) {
-    if (x == null || y == null) {
-      return null;
+  private static NumberType getOutputType(Number... nums) throws DirectiveExecutionException {
+    NumberType outputType = NumberType.SHORT;
+    for (Number num : nums) {
+      if (num instanceof Byte) {
+        throw new DirectiveExecutionException("Input cannot be of type 'Byte'.");
+      } else if (num == null) {
+        return null;
+      } else if (num instanceof Integer && NumberType.INTEGER.getPriority() > outputType.getPriority()) {
+        outputType = NumberType.INTEGER;
+      } else if (num instanceof Long && NumberType.LONG.getPriority() > outputType.getPriority()) {
+        outputType = NumberType.LONG;
+      } else if (num instanceof Float && NumberType.FLOAT.getPriority() > outputType.getPriority()) {
+        outputType = NumberType.FLOAT;
+      } else if (num instanceof Double && NumberType.DOUBLE.getPriority() > outputType.getPriority()) {
+        outputType = NumberType.DOUBLE;
+      } else if (num instanceof BigDecimal) {
+        return NumberType.BIGDECIMAL;
+      }
     }
-    return x + y;
+    return outputType;
   }
 
   /**
-   * Arithmetic operation - Add two columns.
+   * Utility function to convert a number type to BigDecimal
    */
-  public static Float add(Float x, Float y) {
-    if (x == null || y == null) {
-      return null;
-    }
-    return x + y;
+  private static BigDecimal numberToBigDecimal(Number num) {
+    return num instanceof BigDecimal ? (BigDecimal) num : new BigDecimal(num.toString());
   }
 
   /**
-   * Arithmetic operation - Add two columns.
+   * Arithmetic operation - Find the sum of any number of columns, of any valid types.
+   * Output type is most general of input types.
+   * Returns null if any input value is null.
    */
-  public static BigDecimal add(BigDecimal x, BigDecimal y) {
-    if (x == null || y == null) {
+  public static Number add(Number... nums) throws DirectiveExecutionException {
+    NumberType outputType = getOutputType(nums);
+    if (outputType == null) {
       return null;
     }
-    return x.add(y).stripTrailingZeros();
+    switch (outputType) {
+      case SHORT:
+        short shortSum = 0;
+        for (Number num : nums) {
+          shortSum += num.shortValue();
+        }
+        return shortSum;
+      case INTEGER:
+        int intSum = 0;
+        for (Number num : nums) {
+          intSum += num.intValue();
+        }
+        return intSum;
+      case LONG:
+        long longSum = 0;
+        for (Number num : nums) {
+          longSum += num.longValue();
+        }
+        return longSum;
+      case FLOAT:
+        float floatSum = 0;
+        for (Number num : nums) {
+          floatSum += num.floatValue();
+        }
+        return floatSum;
+      case DOUBLE:
+        double doubleSum = 0;
+        for (Number num : nums) {
+          doubleSum += num.doubleValue();
+        }
+        return doubleSum;
+      default:
+        BigDecimal bdSum = BigDecimal.valueOf(0);
+        for (Number num : nums) {
+          bdSum = bdSum.add(numberToBigDecimal(num));
+        }
+        return bdSum.stripTrailingZeros();
+    }
   }
 
   /**
-   * Arithmetic operation - Subtract a column from another column.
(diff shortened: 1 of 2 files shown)


wrangler-core/src/main/java/io/cdap/directives/column/Copy.java
```diff
@@ -74,33 +74,34 @@ public void destroy() {
...
-          throw new DirectiveExecutionException(
-            NAME, String.format("Destination column '%s' already exists in the row. Use 'force' " +
-                                  "option to overwrite the column.", destination.value()));
-        }
-        row.setValue(didx, row.getValue(sidx));
-      }
-    }
-    return rows;
+    throw new DirectiveExecutionException("error!!!");
+//    for (Row row : rows) {
```
sechegaray: Don't leave code commented out please.

adinilfeld: Whoops, I didn't mean to commit that! Fixed now.

wrangler-core/src/main/java/io/cdap/wrangler/utils/ArithmeticOperations.java
```diff
@@ -24,380 +24,462 @@
...
+      case DOUBLE:
+        double doubleSum = 0;
+        for (Number num : nums) {
+          doubleSum += num.doubleValue();
+        }
+        return doubleSum;
+      default:
+        BigDecimal bdSum = BigDecimal.valueOf(0);
+        for (Number num : nums) {
+          bdSum = bdSum.add(new BigDecimal(num.toString()));
```
fernst: You should handle the case where `num` is already a BigDecimal. There is no need to convert to String if `num` is already a BigDecimal.

Also, something to notice with bigDecimals is that the `toString()` for BigDecimals can generate values in scientific notation, as seen in https://stackoverflow.com/questions/15834879/format-bigdecimal-without-scientific-notation-with-full-precision 

For BigDecimals, I think it would be better to use `BigDecimal.toPlainString()` if we need the string value for the number.

wrangler-core/src/main/java/io/cdap/wrangler/utils/ArithmeticOperations.java
```diff
@@ -24,380 +24,462 @@
...
+      case INTEGER:
+        return x.intValue() - y.intValue();
+      case LONG:
+        return x.longValue() - y.longValue();
+      case FLOAT:
+        return x.floatValue() - y.floatValue();
+      case DOUBLE:
+        return x.doubleValue() - y.doubleValue();
+      default:
+        BigDecimal bdX = new BigDecimal(x.toString());
```
fernst: Same comment as before. No need to convert to BigDecimal if the number is already a BigDecimal.

wrangler-core/src/main/java/io/cdap/wrangler/utils/ArithmeticOperations.java
```diff
@@ -24,380 +24,462 @@
...
+      case DOUBLE:
+        double doubleProd = 1;
+        for (Number num : nums) {
+          doubleProd *= num.doubleValue();
+        }
+        return doubleProd;
+      default:
+        BigDecimal bdProd = BigDecimal.valueOf(1);
+        for (Number num : nums) {
+          bdProd = bdProd.multiply(new BigDecimal(num.toString()));
```
fernst: Same comment as before. No need to convert to BigDecimal if the number is already a BigDecimal.

fernst (changes requested):

wrangler-core/src/main/java/io/cdap/wrangler/utils/ArithmeticOperations.java
adinilfeld: Great point about the conversions, thanks for pointing that out! Fixed that in the most recent commit.

For your second comment, are you referring to when we display the BigDecimals in the Wrangler UI? We noticed the same issue, but since the current release behavior includes the scientific notation, we decided to leave it how it is.

fernst: For display, scientific notation is probably fine. I would like to avoid any precision loss from casting a big decimal to string and then reading again. This can make us lose some of the less significant digits due to the print format

adinilfeld: Ah, I see. We shouldn't have to worry about that after the most recent change.

wrangler-core/src/main/java/io/cdap/wrangler/utils/ArithmeticOperations.java
```diff
@@ -24,380 +24,462 @@
...
-  /**
-   * Arithmetic operation - Divide a column by another and return the quotient.
-   */
-  public static Float divideq(Float x, Float y) {
-    if (x == null || y == null || y == 0) {
-      return null;
+  private static NumberType getOutputType(Number... nums) throws DirectiveExecutionException {
+    NumberType outputType = NumberType.SHORT;
+    for (Number num : nums) {
+      int outputTypePriority = java.util.Arrays.asList(NumberType.values()).indexOf(outputType);
```
fernst: Enumerations can have internal "values" attached to them, see https://github.com/cdapio/cdap/blob/163ade54065ecac3a4fd80fe3d76a5aee38474f9/cdap-common/src/main/java/io/cdap/cdap/common/conf/StringUtils.java#L110 as an example.

I think it would be better to have the priority as a value for the enumeration instead of doing an array seek for each input value.

This is an NxM operation, where N is the number of input values and M the length of the array of number type values (so 6)

wrangler-core/src/main/java/io/cdap/wrangler/utils/ArithmeticOperations.java
```diff
@@ -24,380 +24,462 @@
...
+      case DOUBLE:
+        double doubleSum = 0;
+        for (Number num : nums) {
+          doubleSum += num.doubleValue();
+        }
+        return doubleSum;
+      default:
+        BigDecimal bdSum = BigDecimal.valueOf(0);
+        for (Number num : nums) {
+          bdSum = bdSum.add(num instanceof BigDecimal ? (BigDecimal) num : new BigDecimal(num.toString()));
```
fernst: this function is repeated a lot. Make this a `numberToBigDecimal(Number number) function that performs this internally. that way the code is cleaner and more maintainable.

fernst (changes requested):

wrangler-core/src/main/java/io/cdap/wrangler/utils/ArithmeticOperations.java
adinilfeld: Oo, that's a cool feature I hadn't seen before. Fixed!

wrangler-core/src/main/java/io/cdap/wrangler/utils/ArithmeticOperations.java
adinilfeld: Good idea -- fixed!

wrangler-core/src/main/java/io/cdap/wrangler/utils/ArithmeticOperations.java
```diff
@@ -24,380 +24,479 @@
...
-    if (x == null || y == null || y == 0) {
-      return null;
+  private static NumberType getOutputType(Number... nums) throws DirectiveExecutionException {
+    NumberType outputType = NumberType.SHORT;
+    for (Number num : nums) {
+      if (num instanceof Byte) {
+        throw new DirectiveExecutionException("Input cannot be of type 'Byte'.");
+      } else if (num == null) {
+        return null;
+      } else if (num instanceof Integer && outputType.priority == 0) {
```
fernst: I think this would be cleaner to read if the comparison was made using the NumberType priority.

You can implement a `getPriotity()` method in the Enum and use that to extract the priority value. Then, you can simply compare the current outputType priority with the priority for each of the expected output types.

For example: 

```
if (num instanceof Integer && NumberType.INTEGER.getPriority() > outputType.getPriority()) {
...
} else if (num instanceof Long && NumberType.LONG.getPriority() > outputType.getPriority()) 
```

This way, it's more understandable that you are trying to compare the type of NUM with the priority for the ourput types in the NumberType field.

fernst (changes requested):
Just one final request, other than that code looks good

wrangler-core/src/main/java/io/cdap/wrangler/utils/ArithmeticOperations.java
```diff
@@ -24,380 +24,483 @@
...
+      } else if (num instanceof Integer && NumberType.INTEGER.getPriority() > outputType.getPriority()) {
+        outputType = NumberType.INTEGER;
+      } else if (num instanceof Long && NumberType.LONG.getPriority() > outputType.getPriority()) {
+        outputType = NumberType.LONG;
+      } else if (num instanceof Float && NumberType.FLOAT.getPriority() > outputType.getPriority()) {
+        outputType = NumberType.FLOAT;
+      } else if (num instanceof Double && NumberType.DOUBLE.getPriority() > outputType.getPriority()) {
+        outputType = NumberType.DOUBLE;
+      } else if (num instanceof BigDecimal) {
+        outputType = NumberType.BIGDECIMAL;
```
fernst: Another potential optimization here is that as soon as the output type is BigDecimal, return. As soon as one of the values is BigDecimal, the output will always be BigDecimal, so no need to keep verifying.

fernst (approved):

sechegaray (approved):

wrangler-core/src/main/java/io/cdap/wrangler/utils/ArithmeticOperations.java
ellliao: We ended up rolling this change back in PR #572, as returning after finding a BigDecimal sometimes resulted in later null values going uncaught, causing a null pointer exception.

PR #584, opened 2022-11-11
[CDAP-20082] Moving the bug fix under a feature flag

iam-divya:


wrangler-transform/src/main/java/io/cdap/wrangler/Wrangler.java
```diff
@@ -385,9 +387,11 @@ public void transform(StructuredRecord input, Emitter<StructuredRecord> emitter)
           emitter.emitError(new InvalidEntry<>(error.getCode(), error.getMessage(), input));
           errorMessages.add(error.getMessage());
         }
-        if (onErrorStrategy.equalsIgnoreCase("fail-pipeline")) {
-          throw new Exception(
-            String.format("Errors in Wrangler Transformation - %s", errorMessages));
+        if (WRANGLER_FAIL_PIPELINE_FOR_ERROR.isEnabled(getContext())) {
+          if (onErrorStrategy.equalsIgnoreCase("fail-pipeline")) {
```
vsethi09: Conditions can be combined instead of nesting:

```
if WRANGLER_FAIL_PIPELINE_FOR_ERROR.isEnabled(getContext())
   && onErrorStrategy.equalsIgnoreCase("fail-pipeline") {
    throw new Exception(...)
}
```

vsethi09: nit: define a const for `"fail-pipeline"`

vsethi09 (approved):

wrangler-transform/src/main/java/io/cdap/wrangler/Wrangler.java
```diff
@@ -385,9 +387,11 @@ public void transform(StructuredRecord input, Emitter<StructuredRecord> emitter)
           emitter.emitError(new InvalidEntry<>(error.getCode(), error.getMessage(), input));
           errorMessages.add(error.getMessage());
         }
-        if (onErrorStrategy.equalsIgnoreCase("fail-pipeline")) {
-          throw new Exception(
-            String.format("Errors in Wrangler Transformation - %s", errorMessages));
+        if (WRANGLER_FAIL_PIPELINE_FOR_ERROR.isEnabled(getContext())) {
```
saimukkamala: does this call always read from file or is it cached? if it does read from file, would it effect the performance? either way, IMO its a good idea to have a class level variable (boolean)

wrangler-transform/src/main/java/io/cdap/wrangler/Wrangler.java
saimukkamala: I believe we can merge this and above `if` conditions to have only one `if` condition

iam-divya: Makes sense. How are feature flags treated once it is removed? I assumed that if condition block will be removed hence added it in nested way.

iam-divya: Updated.

wrangler-transform/src/main/java/io/cdap/wrangler/Wrangler.java
iam-divya: @sechegaray Do you know if the values for this flag will be read repeatedly from the cdap-features/site.xml or would read from memory? This question is for both app and pipeline execution of wrangler code. Every record would go through this hence it is important to understand this flow.

wrangler-transform/src/main/java/io/cdap/wrangler/Wrangler.java
iam-divya: Updated

iam-divya: Updated

wrangler-transform/src/main/java/io/cdap/wrangler/Wrangler.java
saimukkamala: seems like `DefaultFeatureFlagsProvider` reads from file only once and caches the results

sechegaray: Hello @iam-divya ,  as @saimukkamala  mentioned DefaultFeatureFlagsProvider uses cConfiguration which loads properties only once.

sechegaray (approved):

saimukkamala (approved):

PR #590, opened 2022-11-16
Add API to create a recipe

vanathi-g:
Changes:
- add handler and storage classes for recipe management
- add api endpoint to create a new recipe
- also checks whether a recipe with same name already exists
- add unit tests for the above two cases

### DIFF
--- a/wrangler-proto/src/main/java/io/cdap/wrangler/proto/recipe/v2/Recipe.java
+++ b/wrangler-proto/src/main/java/io/cdap/wrangler/proto/recipe/v2/Recipe.java
@@ -25,15 +25,15 @@ import java.util.Objects;
  * Metadata information about a Recipe
  */
 public class Recipe {
+  private final RecipeId recipeId;
   private final String recipeName;
-  private final String recipeId;
   private final String description;
   private final List<String> directives;
   private final long createdTimeMillis;
   private final long updatedTimeMillis;
   private final int recipeStepsCount;
 
-  private Recipe(String recipeName, String recipeId, String description, List<String> directives,
+  private Recipe(RecipeId recipeId, String recipeName, String description, List<String> directives,
                  long createdTimeMillis, long updatedTimeMillis, int recipeStepsCount) {
     this.recipeName = recipeName;
     this.recipeId = recipeId;
@@ -48,7 +48,7 @@ public class Recipe {
     return recipeName;
   }
 
-  public String getRecipeId() {
+  public RecipeId getRecipeId() {
     return recipeId;
   }
 
@@ -91,7 +91,7 @@ public class Recipe {
     return Objects.hash(recipeName, recipeId, directives);
   }
 
-  public static Builder builder(String recipeId) {
+  public static Builder builder(RecipeId recipeId) {
     return new Builder(recipeId);
   }
 
@@ -109,7 +109,7 @@ public class Recipe {
    * Creates a Recipe meta object
    */
   public static class Builder {
-    private final String recipeId;
+    private final RecipeId recipeId;
     private String recipeName;
     private String description;
     private List<String> directives;
@@ -117,7 +117,7 @@ public class Recipe {
     private long updatedTimeMillis;
     private int recipeStepsCount;
 
-    Builder(String recipeId) {
+    Builder(RecipeId recipeId) {
       this.recipeId = recipeId;
       this.directives = new ArrayList<>();
     }
@@ -154,7 +154,7 @@ public class Recipe {
     }
 
     public Recipe build() {
-      return new Recipe(recipeName, recipeId, description, directives,
+      return new Recipe(recipeId, recipeName, description, directives,
                         createdTimeMillis, updatedTimeMillis, recipeStepsCount);
     }
   }
--- a/wrangler-proto/src/main/java/io/cdap/wrangler/proto/recipe/v2/Recipe.java
+++ b/wrangler-proto/src/main/java/io/cdap/wrangler/proto/recipe/v2/Recipe.java
@@ -83,9 +83,7 @@ public class Recipe {
     }
 
     Recipe other = (Recipe) o;
-    return Objects.equals(recipeId, other.recipeId) &&
-      Objects.equals(recipeName, other.recipeName) &&
-      Objects.equals(directives, other.directives);
+    return Objects.equals(recipeId, other.recipeId);
   }
 
   @Override
--- a/wrangler-proto/src/main/java/io/cdap/wrangler/proto/recipe/v2/Recipe.java
+++ b/wrangler-proto/src/main/java/io/cdap/wrangler/proto/recipe/v2/Recipe.java
@@ -1,0 +1,163 @@
+/*
+ * Copyright © 2022 Cask Data, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package io.cdap.wrangler.proto.recipe.v2;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import java.util.Objects;
+
+/**
+ * Metadata information about a Recipe
+ */
+public class Recipe {
+  private final String recipeName;
+  private final String recipeId;
+  private final String description;
+  private final List<String> directives;
+  private final long createdTimeMillis;
+  private final long updatedTimeMillis;
+  private final int recipeStepsCount;
+
+  private Recipe(String recipeName, String recipeId, String description, List<String> directives,
+                 long createdTimeMillis, long updatedTimeMillis, int recipeStepsCount) {
+    this.recipeName = recipeName;
+    this.recipeId = recipeId;
+    this.description = description;
+    this.directives = Collections.unmodifiableList(directives);
+    this.createdTimeMillis = createdTimeMillis;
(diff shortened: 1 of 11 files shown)


wrangler-proto/src/main/java/io/cdap/wrangler/proto/recipe/v2/Recipe.java
```diff
@@ -0,0 +1,163 @@
...
+    if (this == o) {
+      return true;
+    }
+
+    if (o == null || getClass() != o.getClass()) {
+      return false;
+    }
+
+    Recipe other = (Recipe) o;
+    return Objects.equals(recipeId, other.recipeId) &&
```
iam-divya: Do we need namespace here? Should we limit the equality check to PK columns?

wrangler-proto/src/main/java/io/cdap/wrangler/proto/recipe/v2/Recipe.java
```diff
@@ -0,0 +1,163 @@
...
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import java.util.Objects;
+
+/**
+ * Metadata information about a Recipe
+ */
+public class Recipe {
+  private final String recipeName;
```
iam-divya, line 29: namespace is missing here.

wrangler-proto/src/main/java/io/cdap/wrangler/proto/recipe/v2/RecipeCreationRequest.java
```diff
@@ -0,0 +1,46 @@
...
+
+/**
+ * Creation request for a Recipe
+ */
+public class RecipeCreationRequest {
+  private final String recipeName;
+  private final String description;
+  private final List<String> directives;
+
+  public RecipeCreationRequest(String recipeName, String description, List<String> directives) {
```
iam-divya, line 29: Namespace should also be part of this.

wrangler-storage/src/test/java/io/cdap/wrangler/store/recipe/RecipeStoreTest.java
```diff
@@ -0,0 +1,84 @@
...
+    store.saveRecipe(recipeId, recipeRow);
+
+    RecipeId newRecipeId = new RecipeId(summary);
+    Recipe newRecipe = Recipe.builder(newRecipeId.getRecipeId()).setRecipeName("duplicate-name").build();
+    RecipeRow newRecipeRow = RecipeRow.builder(newRecipe).build();
+    try {
+      store.saveRecipe(newRecipeId, newRecipeRow);
+      Assert.fail();
+    } catch (RecipeAlreadyExistsException e) {
+      // expected
```
iam-divya, line 81: Good to assert true here and assert false after this catch block.

wrangler-service/src/main/java/io/cdap/wrangler/service/directive/RecipeHandler.java
```diff
@@ -0,0 +1,84 @@
...
+import java.nio.charset.StandardCharsets;
+import java.util.List;
+import javax.ws.rs.POST;
+import javax.ws.rs.Path;
+import javax.ws.rs.PathParam;
+
+/**
+ * v2 endpoints for recipe
+ */
+public class RecipeHandler extends AbstractWranglerHandler {
```
iam-divya, line 44: Do we have an option to also add tests at end point  level?

wrangler-proto/src/main/java/io/cdap/wrangler/proto/recipe/v2/RecipeCreationRequest.java
vanathi-g: Not necessary as namespace is passed in URL as a path parameter

wrangler-service/src/main/java/io/cdap/wrangler/service/directive/RecipeHandler.java
vanathi-g: Right now we can't add as there are some dependency issues in the package, we are looking into alternatives

wrangler-storage/src/test/java/io/cdap/wrangler/store/recipe/RecipeStoreTest.java
vanathi-g: Will resolve in next PR

wrangler-proto/src/main/java/io/cdap/wrangler/proto/recipe/v2/Recipe.java
```diff
@@ -0,0 +1,161 @@
...
+
+package io.cdap.wrangler.proto.recipe.v2;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import java.util.Objects;
+
+/**
+ * Metadata information about a Recipe
```
iam-divya, line 25: Nit: good to explain about this class and its usage, since similar class RecipeRow also exist.

wrangler-storage/src/test/java/io/cdap/wrangler/store/recipe/RecipeStoreTest.java
iam-divya: Based on the discussion we had, shall we change this to, 
@Test(expected = RecipeAlreadyExistsException.class) 
It will be more specific. This test case will go through even if there are other exceptions other than RecipeAlreadyExistsException

iam-divya (approved):
LGTM

wrangler-storage/src/test/java/io/cdap/wrangler/store/recipe/RecipeStoreTest.java
vanathi-g: Sure, will do in next PR

wrangler-storage/src/main/java/io/cdap/wrangler/dataset/recipe/RecipeRow.java
```diff
@@ -0,0 +1,83 @@
...
+import io.cdap.cdap.api.NamespaceSummary;
+import io.cdap.wrangler.proto.recipe.v2.Recipe;
+import java.util.Objects;
+
+/**
+ * Stores information about Recipe, including information that should not be exposed to users.
+ * {@link Recipe} contains fields that are exposed to users.
+ */
+public class RecipeRow {
+  private final NamespaceSummary namespaceSummary;
```
saimukkamala: Every `Recipe` belongs to a `namespace` and it is essential information for a recipe, so `namespace` should be part of `Recipe` not in `RecipeRow`.

wrangler-proto/src/main/java/io/cdap/wrangler/proto/recipe/v2/Recipe.java
```diff
@@ -0,0 +1,161 @@
...
+ */
+public class Recipe {
+  private final String recipeName;
+  private final String recipeId;
```
saimukkamala: why don't we use `RecipeId` class instead of using recipeId as a `String`? This will also take care of `namespace`.

wrangler-proto/src/main/java/io/cdap/wrangler/proto/recipe/v2/RecipeId.java
```diff
@@ -0,0 +1,63 @@
...
+import io.cdap.cdap.api.NamespaceSummary;
+
+import java.util.Objects;
+import java.util.UUID;
+
+/**
+ * Recipe id
+ */
+public class RecipeId {
+  private final NamespaceSummary namespace;
```
saimukkamala, line 28: why do we need to store the entire `namespace` summary? `NamespaceSummary` has a description field, which might be large. Also, cdap does store `namespace` summary. Storing this information would be duplication of data. I believe storing namespace as a String would be sufficient.

wrangler-storage/src/main/java/io/cdap/wrangler/store/recipe/RecipeStore.java
```diff
@@ -0,0 +1,168 @@
...
+    return GSON.fromJson(iterator.next().getString(RECIPE_INFO_COL), RecipeRow.class);
+  }
+
+  private Collection<Field<?>> getRecipeKeys(RecipeId recipeId) {
+    List<Field<?>> keys = new ArrayList<>(getNamespaceKeys(recipeId.getNamespace()));
+    keys.add(Fields.stringField(RECIPE_ID_FIELD, recipeId.getRecipeId()));
+    return keys;
+  }
+
+  private Collection<Field<?>> getNamespaceKeys(NamespaceSummary namespace) {
```
saimukkamala: I see the same function in `WorkspaceStore.java`, can we reuse the same function or may be move the function to a common class, so both Stores can use the same function?

wrangler-proto/src/main/java/io/cdap/wrangler/proto/recipe/v2/RecipeId.java
```diff
@@ -0,0 +1,63 @@
...
+
+import io.cdap.cdap.api.NamespaceSummary;
+
+import java.util.Objects;
+import java.util.UUID;
+
+/**
+ * Recipe id
+ */
+public class RecipeId {
```
saimukkamala, line 27: Im trying to understand why this class exists. can you explain the purpose?

wrangler-proto/src/main/java/io/cdap/wrangler/proto/recipe/v2/RecipeId.java
```diff
@@ -0,0 +1,63 @@
...
+  private final NamespaceSummary namespace;
+  private final String recipeId;
+
+  public RecipeId(NamespaceSummary namespace) {
+    this.namespace = new NamespaceSummary(namespace.getName(), null, namespace.getGeneration());
```
saimukkamala, line 32: can you just add a comment - why we are making the namespacesummary description is set null.

saimukkamala (approved):

wrangler-proto/src/main/java/io/cdap/wrangler/proto/recipe/v2/RecipeId.java
vanathi-g: Will add in next PR

PR #592, opened 2022-11-24
Add API to get a Recipe

vanathi-g:
- add api endpoint to get a recipe (given the recipeId)
- add unit tests for getting existing as well as non-existent recipe

### DIFF
--- a/wrangler-proto/src/main/java/io/cdap/wrangler/proto/recipe/v2/RecipeId.java
+++ b/wrangler-proto/src/main/java/io/cdap/wrangler/proto/recipe/v2/RecipeId.java
@@ -29,8 +29,13 @@ public class RecipeId {
   private final String recipeId;
 
   public RecipeId(NamespaceSummary namespace) {
+    this(namespace, UUID.randomUUID().toString());
+  }
+
+  public RecipeId(NamespaceSummary namespace, String recipeId) {
+    // Set namespace description to null - to avoid duplicate storage of large descriptions
     this.namespace = new NamespaceSummary(namespace.getName(), null, namespace.getGeneration());
-    this.recipeId = UUID.randomUUID().toString();
+    this.recipeId = recipeId;
   }
 
   public NamespaceSummary getNamespace() {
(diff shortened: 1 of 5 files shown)


wrangler-storage/src/main/java/io/cdap/wrangler/store/recipe/RecipeStore.java
```diff
@@ -119,6 +120,14 @@ private void saveRecipe(RecipeId recipeId, RecipeRow recipeRow, boolean failIfNo
     });
   }
 
+  public Recipe getRecipe(RecipeId recipeId) throws RecipeNotFoundException {
```
iam-divya, line 128: Do we have retries setup in case of failures? It is important to set that for getRecipe as it could be used in pipelines also. TransactionRunner object that is created for recipe store can have retry configured if it is not done already.

wrangler-proto/src/main/java/io/cdap/wrangler/proto/recipe/v2/RecipeId.java
```diff
@@ -29,8 +29,13 @@ public class RecipeId {
   private final String recipeId;
 
   public RecipeId(NamespaceSummary namespace) {
+    this(namespace, UUID.randomUUID().toString());
+  }
+
```
saimukkamala, line 34: can you add a comment - this method should not be used while creating a new recipe

wrangler-storage/src/main/java/io/cdap/wrangler/store/recipe/RecipeStore.java
vanathi-g: The injected transaction runner takes care of retrying sql queries, so we need not do this

saimukkamala (approved):

iam-divya (approved):
LGTM

wrangler-proto/src/main/java/io/cdap/wrangler/proto/recipe/v2/RecipeId.java
vanathi-g: Will add in the next PR

PR #593, opened 2022-11-28
Add API to list Recipes

vanathi-g:
- Add api endpoint to fetch a paginated list of recipes
- Pagination query parameters:
    - `sortBy` field recipe name (ascending, default) or updated time (descending)
    - `pageSize` (default = 10)
    - `pageToken` (default = empty fetches first page)
- `RecipeListResponse` returns `nextPageToken`
- Add unit tests that test the following:
    - default values for pagination (and sort by name)
    - sort by updated
    - pagination

### DIFF
--- a/wrangler-proto/src/main/java/io/cdap/wrangler/proto/recipe/v2/RecipeId.java
+++ b/wrangler-proto/src/main/java/io/cdap/wrangler/proto/recipe/v2/RecipeId.java
@@ -28,13 +28,8 @@ public class RecipeId {
   private final NamespaceSummary namespace;
   private final String recipeId;
 
-  public RecipeId(NamespaceSummary namespace) {
-    this(namespace, UUID.randomUUID().toString());
-  }
-
-  public RecipeId(NamespaceSummary namespace, String recipeId) {
-    // Set namespace description to null - to avoid duplicate storage of large descriptions
-    this.namespace = new NamespaceSummary(namespace.getName(), null, namespace.getGeneration());
+  private RecipeId(NamespaceSummary namespace, String recipeId) {
+    this.namespace = namespace;
     this.recipeId = recipeId;
   }
 
@@ -64,5 +59,35 @@ public class RecipeId {
   @Override
   public int hashCode() {
     return Objects.hash(namespace, recipeId);
+  }
+
+  public static Builder builder(NamespaceSummary namespace) {
+    return new Builder(namespace);
+  }
+
+  public static Builder builder(RecipeId existing) {
+    return new Builder(existing.getNamespace()).setRecipeId(existing.getRecipeId());
+  }
+
+  /**
+   * Creates a Recipe id object
+   */
+  public static class Builder {
+    private final NamespaceSummary namespace;
+    private String recipeId;
+
+    Builder(NamespaceSummary namespace) {
+      this.namespace = new NamespaceSummary(namespace.getName(), null, namespace.getGeneration());
+      this.recipeId = UUID.randomUUID().toString();
+    }
+
+    public Builder setRecipeId(String recipeId) {
+      this.recipeId = recipeId;
+      return this;
+    }
+
+    public RecipeId build() {
+      return new RecipeId(namespace, recipeId);
+    }
   }
 }
--- a/wrangler-proto/src/main/java/io/cdap/wrangler/proto/recipe/v2/RecipeListResponse.java
+++ b/wrangler-proto/src/main/java/io/cdap/wrangler/proto/recipe/v2/RecipeListResponse.java
@@ -1,0 +1,37 @@
+/*
+ * Copyright © 2022 Cask Data, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package io.cdap.wrangler.proto.recipe.v2;
+
+import io.cdap.wrangler.proto.ServiceResponse;
+
+import java.util.List;
+
+/**
+ * Represents a paginated list of recipes as a response.
+ */
+public class RecipeListResponse extends ServiceResponse<Recipe> {
+  String nextPageToken;
+
+  public RecipeListResponse(List<Recipe> recipes, String nextPageToken) {
+    super(recipes);
+    this.nextPageToken = nextPageToken;
+  }
+
+  public String getNextPageToken() {
+    return nextPageToken;
+  }
+}
--- a/wrangler-service/src/main/java/io/cdap/wrangler/service/directive/RecipeHandler.java
+++ b/wrangler-service/src/main/java/io/cdap/wrangler/service/directive/RecipeHandler.java
@@ -25,6 +25,7 @@ import io.cdap.cdap.api.service.http.HttpServiceRequest;
 import io.cdap.cdap.api.service.http.HttpServiceResponder;
 import io.cdap.cdap.api.service.http.SystemHttpServiceContext;
 import io.cdap.cdap.internal.io.SchemaTypeAdapter;
+import io.cdap.wrangler.dataset.recipe.RecipePageRequest;
 import io.cdap.wrangler.dataset.recipe.RecipeRow;
 import io.cdap.wrangler.proto.recipe.v2.Recipe;
 import io.cdap.wrangler.proto.recipe.v2.RecipeCreationRequest;
@@ -39,6 +40,7 @@ import javax.ws.rs.GET;
 import javax.ws.rs.POST;
 import javax.ws.rs.Path;
 import javax.ws.rs.PathParam;
+import javax.ws.rs.QueryParam;
 
 /**
  * v2 endpoints for recipe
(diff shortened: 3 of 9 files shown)


wrangler-proto/src/main/java/io/cdap/wrangler/proto/recipe/v2/RecipeId.java
```diff
@@ -32,6 +32,7 @@ public RecipeId(NamespaceSummary namespace) {
     this(namespace, UUID.randomUUID().toString());
   }
 
+  // Do not use this method while creating a new recipe
   public RecipeId(NamespaceSummary namespace, String recipeId) {
```
iam-divya: Do you want to make this private and expose getInstance api which invokes this function? 
Enforcement via comments may not work in all the cases.

wrangler-proto/src/main/java/io/cdap/wrangler/proto/recipe/v2/RecipeListResponse.java
```diff
@@ -0,0 +1,37 @@
...
+package io.cdap.wrangler.proto.recipe.v2;
+
+import io.cdap.wrangler.proto.ServiceResponse;
+
+import java.util.List;
+
+/**
+ * Represents a paginated list of recipes as a response.
+ */
+public class RecipeListResponse extends ServiceResponse<Recipe> {
```
iam-divya, line 26: Can you define the interfaces for prev page logic, consumers can use it for integration until the implementation is getting ready.

wrangler-storage/src/main/java/io/cdap/wrangler/store/recipe/RecipeStore.java
```diff
@@ -133,6 +139,47 @@ public Recipe getRecipe(RecipeId recipeId) throws RecipeNotFoundException {
...
   }
 
+  public RecipeListResponse listRecipes(NamespaceSummary namespace, Integer pageSize, String pageToken, String sortBy) {
+    return TransactionRunners.run(transactionRunner, context -> {
+      List<Recipe> recipes = new ArrayList<>();
+      StructuredTable table = context.getTable(TABLE_ID);
+
+      // By default, sort by recipe name
+      Field<?> sortByField = Fields.stringField(RECIPE_NAME_FIELD, pageToken);
+      SortOrder sortOrder = SortOrder.ASC;
```
iam-divya: Good to have this configurable. You can still continue to have default values based on sort field.

wrangler-proto/src/main/java/io/cdap/wrangler/proto/recipe/v2/RecipeId.java
vanathi-g: Changed it to use a Builder

wrangler-storage/src/main/java/io/cdap/wrangler/store/recipe/RecipeStore.java
vanathi-g: Done

wrangler-storage/src/main/java/io/cdap/wrangler/store/recipe/RecipeStore.java
```diff
@@ -133,6 +136,66 @@ public Recipe getRecipe(RecipeId recipeId) throws RecipeNotFoundException {
...
+      Collection<Field<?>> end = getNamespaceKeys(NAMESPACE_FIELD, GENERATION_COL, namespace);
+
+      if (!pageToken.equals("")) {
+        // By default, sort by recipe name
+        Field<?> sortByField = Fields.stringField(RECIPE_NAME_FIELD, pageToken);
+        if (sortBy.equals(UPDATE_TIME_COL)) {
+          try {
+            sortByField = Fields.longField(UPDATE_TIME_COL, Long.parseLong(pageToken));
+          } catch (NumberFormatException e) {
+            throw new IllegalArgumentException("pageToken value is of invalid data type");
```
saimukkamala: This error seems vague. can we add more details so debugging will be simpler? - column name and expected data type.

iam-divya (approved):
LGTM

wrangler-storage/src/main/java/io/cdap/wrangler/dataset/recipe/RecipePageRequest.java
```diff
@@ -0,0 +1,160 @@
...
+      } else {
+        end.add(sortByField);
+      }
+    }
+
+    return Range.create(begin, INCLUSIVE, end, INCLUSIVE);
+  }
+
+  @Override
+  public <T> String getNextPageToken(T obj) {
```
saimukkamala: I don't see any need to use generics here. Any reason why we are using generics here?

wrangler-storage/src/main/java/io/cdap/wrangler/dataset/recipe/RecipePageRequest.java
```diff
@@ -0,0 +1,160 @@
...
+    String nextPageToken = "";
+    switch (this.getSortBy()) {
+      case RECIPE_NAME_FIELD:
+        nextPageToken = recipe.getRecipeName();
+        break;
+      case UPDATE_TIME_COL:
+        nextPageToken = String.valueOf(recipe.getUpdatedTimeMillis());
+        break;
+    }
+    return nextPageToken;
```
saimukkamala, line 114: In which cases can we expect the `nextPageToken` to be an empty string?

wrangler-storage/src/main/java/io/cdap/wrangler/dataset/utils/PageRequest.java
```diff
@@ -0,0 +1,77 @@
...
+    return sortOrder;
+  }
+
+  /**
+   * Constructs the range used to query the storage table.
+   * @return {@link Range} corresponding to page query parameters.
+   */
+  public abstract Range getScanRange();
+
+  public abstract <T> String getNextPageToken(T object);
```
saimukkamala: Same comment as before. why do we need generics here?

wrangler-storage/src/main/java/io/cdap/wrangler/dataset/utils/PageRequest.java
```diff
@@ -0,0 +1,77 @@
...
+
+/**
+ * An interface to implement a page request. Pagination utility class that contains relevant fields and methods
+ * that describe a page.
+ */
+public abstract class PageRequest {
+  /**
+   * Maximum number of results to return in a page.
+   */
+  int pageSize;
```
saimukkamala: `private` modifiers for all the member variables

wrangler-storage/src/main/java/io/cdap/wrangler/dataset/utils/PageRequest.java
```diff
@@ -0,0 +1,77 @@
...
+
+  /**
+   * Token to identify the page to retrieve. Either received from a previous page or left empty to fetch the first page.
+   */
+  String pageToken;
```
saimukkamala: from the logic in other classes I see `pageToken` can bell `null`. So, we should annotate variables with `@Nullable`. Similarly, any function which can possibly return a `null` value should be annotated as well. eg: `public String getPageToken()`

wrangler-service/src/main/java/io/cdap/wrangler/service/directive/RecipeHandler.java
```diff
@@ -96,7 +114,41 @@ public void getRecipe(HttpServiceRequest request, HttpServiceResponder responder
...
+  @TransactionPolicy(value = TransactionControl.EXPLICIT)
+  @Path("v2/contexts/{context}/recipes")
+  public void listRecipes(HttpServiceRequest request, HttpServiceResponder responder,
+                          @PathParam("context") String namespace,
+                          @QueryParam("pageSize") Integer pageSize,
+                          @QueryParam("pageToken") String pageToken,
+                          @QueryParam("sortBy")String sortBy,
+                          @QueryParam("sortOrder") String sortOrder) {
+    if (sortBy != null && !(sortByMap.containsKey(sortBy))) {
+      throw new IllegalArgumentException("specified sortBy field is not supported");
```
saimukkamala: Can we also add the sortBy field name in the exception, so debugging would be easy.

wrangler-service/src/main/java/io/cdap/wrangler/service/directive/RecipeHandler.java
```diff
@@ -96,7 +114,41 @@ public void getRecipe(HttpServiceRequest request, HttpServiceResponder responder
...
+    }
+
+    if (sortOrder != null && !(sortOrder.equals(SORT_ORDER_ASC) || sortOrder.equals(SORT_ORDER_DESC))) {
+      throw new IllegalArgumentException("specified sortOrder is not supported");
```
saimukkamala: same here, add the sortOrder to the exception.

wrangler-storage/src/main/java/io/cdap/wrangler/dataset/utils/PageRequest.java
vanathi-g: Here it is necessary as the nextPageToken is determined using the resource returned in the respective page. The resource may vary depending on the API that uses this.

wrangler-storage/src/main/java/io/cdap/wrangler/dataset/recipe/RecipePageRequest.java
vanathi-g: For the last page of results, it will be set as an empty string in the response.

wrangler-storage/src/main/java/io/cdap/wrangler/dataset/recipe/RecipePageRequest.java
vanathi-g: Changed it to enforce Recipe object

wrangler-storage/src/main/java/io/cdap/wrangler/dataset/recipe/RecipePageRequest.java
```diff
@@ -0,0 +1,194 @@
...
+   * Creates a {@link RecipePageRequest} object
+   */
+  public static class Builder {
+    public static final String SORT_BY_NAME = "name";
+    public static final String SORT_BY_UPDATE_TIME = "updated";
+    public static final String SORT_ORDER_ASC = "asc";
+    public static final String SORT_ORDER_DESC = "desc";
+    private static final int PAGE_SIZE_DEFAULT = 10;
+
+    private final Map<String, String> sortByMap = new HashMap<>();
```
saimukkamala: we should make this mapping static

saimukkamala (approved):

PR #595, opened 2022-12-02
Add delete API endpoint

vanathi-g:
- Add api endpoint to delete a recipe (given the recipeId)
- Add unit tests for deleting existing as well as non-existent recipe


saimukkamala (approved):

wrangler-storage/src/test/java/io/cdap/wrangler/store/recipe/RecipeStoreTest.java
```diff
@@ -176,4 +176,22 @@ public void testListRecipesPagination() {
     Assert.assertEquals(values.get(0), recipeRow3.getRecipe());
     Assert.assertEquals(page2.getNextPageToken(), "");
   }
+
+  @Test(expected = RecipeNotFoundException.class)
```
iam-divya, line 180: Can you add one more test case of adding a recipe with the same name and namespace as deleted recipe?

iam-divya (commented):
Feel free to add it in another PR also. Rest looks good.

PR #598, opened 2022-12-07
Adding new api: Edit recipe

saimukkamala:
- Added edit recipe api
- Added validation for recipe object for requests
- refactored saveRecipe function in RecipeStore

### DIFF
--- a/wrangler-service/src/main/java/io/cdap/wrangler/service/directive/RecipeHandler.java
+++ b/wrangler-service/src/main/java/io/cdap/wrangler/service/directive/RecipeHandler.java
@@ -27,6 +27,7 @@ import io.cdap.cdap.api.service.http.SystemHttpServiceContext;
 import io.cdap.cdap.internal.io.SchemaTypeAdapter;
 import io.cdap.wrangler.dataset.recipe.RecipePageRequest;
 import io.cdap.wrangler.dataset.recipe.RecipeRow;
+import io.cdap.wrangler.proto.BadRequestException;
 import io.cdap.wrangler.proto.recipe.v2.Recipe;
 import io.cdap.wrangler.proto.recipe.v2.RecipeCreationRequest;
 import io.cdap.wrangler.proto.recipe.v2.RecipeId;
@@ -40,6 +41,7 @@ import java.util.regex.Pattern;
 import javax.ws.rs.DELETE;
 import javax.ws.rs.GET;
 import javax.ws.rs.POST;
+import javax.ws.rs.PUT;
 import javax.ws.rs.Path;
 import javax.ws.rs.PathParam;
 import javax.ws.rs.QueryParam;
@@ -50,6 +52,7 @@ import javax.ws.rs.QueryParam;
 public class RecipeHandler extends AbstractWranglerHandler {
   private static final Gson GSON =
     new GsonBuilder().registerTypeAdapter(Schema.class, new SchemaTypeAdapter()).create();
+  private static final Pattern RECIPE_NAME_PATTERN = Pattern.compile("[a-zA-Z0-9 ]*");
 
   private RecipeStore recipeStore;
 
@@ -66,29 +69,8 @@ public class RecipeHandler extends AbstractWranglerHandler {
                            @PathParam("context") String namespace) {
     respond(responder, namespace, ns -> {
       RecipeId recipeId = RecipeId.builder(ns).build();
-      RecipeCreationRequest creationRequest = GSON.fromJson(
-        StandardCharsets.UTF_8.decode(request.getContent()).toString(), RecipeCreationRequest.class);
-
-      Pattern pattern = Pattern.compile("[a-zA-Z0-9 ]*");
-      if (!pattern.matcher(creationRequest.getRecipeName()).matches()) {
-        throw new IllegalArgumentException("recipe name should contain only alphanumeric characters or spaces");
-      }
-
-      List<String> directives = creationRequest.getDirectives();
-      long now = System.currentTimeMillis();
-
-      Recipe recipe = Recipe.builder(recipeId)
-        .setRecipeName(creationRequest.getRecipeName())
-        .setDescription(creationRequest.getDescription())
-        .setCreatedTimeMillis(now)
-        .setUpdatedTimeMillis(now)
-        .setDirectives(directives)
-        .setRecipeStepsCount(directives.size())
-        .build();
-
-      RecipeRow recipeRow = RecipeRow.builder(recipe).build();
-
-      recipeStore.saveRecipe(recipeId, recipeRow);
+      Recipe recipe = buildRecipeFromRequest(request, recipeId);
+      recipeStore.createRecipe(recipeId, RecipeRow.builder(recipe).build());
       responder.sendJson(recipe);
     });
   }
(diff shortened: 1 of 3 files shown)


wrangler-service/src/main/java/io/cdap/wrangler/service/directive/RecipeHandler.java
```diff
@@ -147,4 +147,52 @@ public void deleteRecipe(HttpServiceRequest request, HttpServiceResponder respon
...
+      RecipeId recipeId = RecipeId.builder(ns).setRecipeId(recipeIdString).build();
+      // get recipe from request
+      Recipe recipe = GSON.fromJson(
+        StandardCharsets.UTF_8.decode(request.getContent()).toString(), Recipe.class);
+      // validate recipe
+      validateRecipeRequest(recipe);
+      // create Recipe for update from request recipe
+      long now = System.currentTimeMillis();
+
+      Recipe newRecipe = Recipe.builder(recipeId)
```
iam-divya: UI allows editing recipe name and desc only today. Should we get rest of information from database?

wrangler-storage/src/main/java/io/cdap/wrangler/store/recipe/RecipeStore.java
```diff
@@ -89,37 +89,61 @@ public RecipeStore(TransactionRunner transactionRunner) {
    * @param recipeRow recipe to create/update
    */
   public void saveRecipe(RecipeId recipeId, RecipeRow recipeRow) {
```
iam-divya: CreateRecipe would be a better name since save can point to both create and update.

wrangler-storage/src/main/java/io/cdap/wrangler/store/recipe/RecipeStore.java
```diff
@@ -89,37 +89,61 @@ public RecipeStore(TransactionRunner transactionRunner) {
...
-        newRecipe = RecipeRow.builder(updatedRecipe).build();
-      }
+  // Create or update a recipe
+  // failIfNotFound should be false for create, and true for update
+  private void saveRecipe(RecipeId recipeId, RecipeRow recipeRow, boolean failIfNotFound, StructuredTable table)
+    throws IOException, RecipeAlreadyExistsException {
+    RecipeRow oldRecipe = getRecipeInternal(table, recipeId, failIfNotFound);
+    if (oldRecipe != null) {
+      Recipe updatedRecipe = Recipe.builder(recipeRow.getRecipe())
+        .setCreatedTimeMillis(oldRecipe.getRecipe().getCreatedTimeMillis()).build();
```
iam-divya, line 132: create timestamp is already available in oldRecipeWithName in updateRecipe, should we move this logic there? It is needed only for update. We may need similar value-fill for columns that are not passed from UI for update.

wrangler-service/src/main/java/io/cdap/wrangler/service/directive/RecipeHandler.java
vanathi-g: Should we add another builder method to copy an existing recipe and then set necessary values (created time and update time) alone?

saimukkamala: This api allows us to edit name, description and directives only. Rest of the fields are picked up from database.

wrangler-storage/src/main/java/io/cdap/wrangler/store/recipe/RecipeStore.java
saimukkamala: makes sense. Updated.

wrangler-storage/src/main/java/io/cdap/wrangler/store/recipe/RecipeStore.java
saimukkamala: `oldRecipeWithName` retrieves recipe with name provided in the request (to make sure the new recipe name to which we want to update is already not being used by different recipe). In case the name is updated, either `oldRecipeWithName` is `null` or might retrieve a different recipe with the same name. Ideally, we should retrieve existing recipe with its ID, which we are doing it here.

wrangler-service/src/main/java/io/cdap/wrangler/service/directive/RecipeHandler.java
saimukkamala: Also, in the edit request, we expect the entire (i.e. mandatory fields - which are name, desc, list of directives) Recipe object, rather than figuring out which fields were updated and which ones we need to pick from database. I believe this is a clean and easier way to update. wdyt?
Since UI always "gets" a recipe before "updating", this way of updating should not be a problem.

wrangler-service/src/main/java/io/cdap/wrangler/service/directive/RecipeHandler.java
```diff
@@ -147,4 +147,52 @@ public void deleteRecipe(HttpServiceRequest request, HttpServiceResponder respon
...
+  public void updateRecipe(HttpServiceRequest request, HttpServiceResponder responder,
+                           @PathParam("context") String namespace,
+                           @PathParam("recipe-id") String recipeIdString) {
+    respond(responder, namespace, ns-> {
+      RecipeId recipeId = RecipeId.builder(ns).setRecipeId(recipeIdString).build();
+      // get recipe from request
+      Recipe recipe = GSON.fromJson(
+        StandardCharsets.UTF_8.decode(request.getContent()).toString(), Recipe.class);
+      // validate recipe
+      validateRecipeRequest(recipe);
```
vanathi-g: What about validating the case where recipe id passed as query param is not the same id of recipe passed in body of request?

wrangler-service/src/main/java/io/cdap/wrangler/service/directive/RecipeHandler.java
iam-divya: I remember seeing inline editing of name/description in listing page where directives might not have been sent. We can confirm this with Radhika or Divami team.

saimukkamala: The builder method to create a new recipe from an existing recipe is already written. In this case, we use `recipeId` from query param rather than using the recipeId in the request body (which we ignore). So, we wouldn't want to completely clone the `Recipe` object passed from request body.

iam-divya (approved):
Please close the edit issue with Frontend team.

wrangler-service/src/main/java/io/cdap/wrangler/service/directive/RecipeHandler.java
saimukkamala: what we are doing right now is - we are ignoring the recipeId passed in the body of request and only using the recipe id passed as query param. Ideally, we should be creating a new class which only contains the fields that can be edited. Now thinking about it, I feel it's better we create a new object for Recipe edit request to avoid any confusion.

saimukkamala: Made a change - reusing `RecipeCreationRequest` as request body for Update recipe request.

vanathi-g (approved):

PR #600, opened 2022-12-16
adding new API: apply recipe on a workspace

saimukkamala:


wrangler-service/src/main/java/io/cdap/wrangler/service/directive/WorkspaceHandler.java
```diff
@@ -437,6 +424,57 @@ schema, new Plugin("Wrangler", "transform", properties,
...
+    respond(responder, namespace, ns -> {
+      validateNamespace(ns, "Executing directives in system namespace is currently not supported");
+
+      RecipeId recipeId = RecipeId.builder(ns).setRecipeId(recipeIdString).build();
+      Recipe recipe = recipeStore.getRecipeById(recipeId);
+
+      DirectiveExecutionResponse response = execute(ns, request, new WorkspaceId(ns, workspaceId),
+                                                    recipe.getDirectives());
+      responder.sendJson(response);
+    });
```
iam-divya, line 443: How do we send the errors like line number from applying recipe? Do you want to handle that in a separate task/PR?

iam-divya (approved):

wrangler-service/src/main/java/io/cdap/wrangler/service/directive/WorkspaceHandler.java
saimukkamala: Yes, we will be doing it in a separate PR.

PR #602, opened 2022-12-23
Adding metrics for recipe management

saimukkamala:


wrangler-service/src/main/java/io/cdap/wrangler/service/directive/RecipeHandler.java
```diff
@@ -56,6 +57,13 @@ public class RecipeHandler extends AbstractWranglerHandler {
 
   private RecipeStore recipeStore;
 
+  // Injected by CDAP
+  @SuppressWarnings("unused")
+  private Metrics metrics;
```
saimukkamala, line 62: FYI - this `Metrics` object is injected by CDAP with dimensions "namespace", "app", "service", "run", "instance"

iam-divya (approved):

PR #603, opened 2023-01-04
Better error propagation while executing directives

vanathi-g:
## Changes
- Add `rowIndex` and `directiveIndex` fields to `RecipeException` to track which row and directive caused the error
- Add a new response object returned by API call when RecipeException occurs
- Change handlers to handle different exceptions being re-thrown as `RecipeException`s
- Get and propagate row, directive indices from parser & recipe executor classes

### DIFF
--- a/wrangler-api/src/main/java/io/cdap/wrangler/api/RecipeException.java
+++ b/wrangler-api/src/main/java/io/cdap/wrangler/api/RecipeException.java
@@ -21,16 +21,31 @@ package io.cdap.wrangler.api;
  * communicating issues with execution of pipeline.
  */
 public class RecipeException extends Exception {
-  public RecipeException(Exception e) {
-    super(e);
+  public static final int UNKNOWN_INDEX = -1;
+
+  // Index of row in dataset and directive in recipe that caused the error
+  private final int rowIndex;
+  private final int directiveIndex;
+
+  public RecipeException(String message, Throwable throwable, int rowIndex, int directiveIndex) {
+    super(message, throwable);
+    this.rowIndex = rowIndex;
+    this.directiveIndex = directiveIndex;
   }
 
-  public RecipeException(String message) {
-    super(message);
+  public RecipeException(String message, Throwable throwable, int directiveIndex) {
+    this(message, throwable, UNKNOWN_INDEX, directiveIndex);
   }
 
   public RecipeException(String message, Throwable throwable) {
-    super(message, throwable);
+    this(message, throwable, UNKNOWN_INDEX, UNKNOWN_INDEX);
+  }
+
+  public int getRowIndex() {
+    return rowIndex;
+  }
+
+  public int getDirectiveIndex() {
+    return directiveIndex;
   }
 }
-
--- a/wrangler-core/src/main/java/io/cdap/wrangler/executor/RecipePipelineExecutor.java
+++ b/wrangler-core/src/main/java/io/cdap/wrangler/executor/RecipePipelineExecutor.java
@@ -20,9 +20,6 @@ import io.cdap.cdap.api.data.format.StructuredRecord;
 import io.cdap.cdap.api.data.schema.Schema;
 import io.cdap.wrangler.api.Directive;
 import io.cdap.wrangler.api.DirectiveExecutionException;
-import io.cdap.wrangler.api.DirectiveLoadException;
-import io.cdap.wrangler.api.DirectiveNotFoundException;
-import io.cdap.wrangler.api.DirectiveParseException;
 import io.cdap.wrangler.api.ErrorRecord;
 import io.cdap.wrangler.api.ErrorRowException;
 import io.cdap.wrangler.api.Executor;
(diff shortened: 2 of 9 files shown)


wrangler-api/src/main/java/io/cdap/wrangler/api/RecipeException.java
```diff
@@ -21,16 +21,31 @@
...
+  private final int rowIndex;
+  private final int directiveIndex;
+
+  public RecipeException(String message, Throwable throwable, int rowIndex, int directiveIndex) {
+    super(message, throwable);
+    this.rowIndex = rowIndex;
+    this.directiveIndex = directiveIndex;
   }
 
-  public RecipeException(String message) {
```
iam-divya, line 28: Is this method not being used anywhere?

wrangler-core/src/main/java/io/cdap/wrangler/parser/GrammarBasedParser.java
```diff
@@ -95,8 +98,8 @@ public List<Directive> parse()
       return result;
     } catch (CompileException e) {
       throw new DirectiveParseException(e.getMessage(), e.iterator(), e);
```
iam-divya: Why is DirectiveParseException not re-thrown as recipe exception?

wrangler-api/src/main/java/io/cdap/wrangler/api/RecipeException.java
vanathi-g: No it's not used anywhere

wrangler-core/src/main/java/io/cdap/wrangler/parser/GrammarBasedParser.java
vanathi-g: DirectiveParseExceptions are thrown from several different places (for ex, from initialise methods of directives, grammar parse methods etc). In most cases, the index of the directive is not easily available. So, it's not re-thrown here to keep it consistent with the other uses

wrangler-core/src/main/java/io/cdap/wrangler/parser/GrammarBasedParser.java
```diff
@@ -95,8 +98,8 @@ public List<Directive> parse()
...
-    } catch (DirectiveLoadException | DirectiveNotFoundException | DirectiveParseException e) {
-      throw e;
+    } catch (DirectiveLoadException | DirectiveNotFoundException e) {
+      throw new RecipeException(e.getMessage(), e, directiveIndex.get());
     } catch (Exception e) {
       throw new DirectiveParseException(e.getMessage(), e);
```
iam-divya: In that case, do you want to add directive number to DirectiveParseException and pass it from here? It will make this function uniformly capture directive index for all kind of errors.

iam-divya:
Unexpected exception, expected<io.cdap.wrangler.api.DirectiveNotFoundException> but was<io.cdap.wrangler.api.RecipeException>
 
Seems to be a genuine test failure, can you please check?

saimukkamala (approved):

wrangler-core/src/main/java/io/cdap/wrangler/executor/RecipePipelineExecutor.java
```diff
@@ -159,7 +160,7 @@ private List<Directive> getDirectives() throws RecipeException {
     if (directives == null) {
       try {
         this.directives = recipeParser.parse();
-      } catch (DirectiveParseException | DirectiveNotFoundException | DirectiveLoadException e) {
+      } catch (DirectiveParseException e) {
         throw new RecipeException(e.getMessage(), e);
```
iam-divya: Do you want to add directiveIndex here?

wrangler-core/src/main/java/io/cdap/wrangler/parser/GrammarBasedParser.java
```diff
@@ -93,12 +94,10 @@ public List<Directive> parse()
...
 
       return result;
-    } catch (CompileException e) {
-      throw new DirectiveParseException(e.getMessage(), e.iterator(), e);
     } catch (DirectiveLoadException | DirectiveNotFoundException | DirectiveParseException e) {
-      throw e;
+      throw new RecipeException(e.getMessage(), e, directiveIndex.get());
     } catch (Exception e) {
-      throw new DirectiveParseException(e.getMessage(), e);
+      throw new RecipeException(e.getMessage(), e);
```
iam-divya, line 100: Directive index needs to be added here.

iam-divya (approved):
Left a minor comment.

wrangler-core/src/main/java/io/cdap/wrangler/parser/GrammarBasedParser.java
vanathi-g: Not necessary because the other exceptions are thrown from functions where the directive index is not available. Will check again and update if needed later

PR #608, opened 2023-02-01
Removing API metrics which are already captured by CDAP

saimukkamala:
- Removing API metrics since CDAP captures all HTTP api calls by default.
- Only metric we are pushing right now is Number of directives saved per recipe.

### DIFF
--- a/wrangler-service/src/main/java/io/cdap/wrangler/service/directive/WorkspaceHandler.java
+++ b/wrangler-service/src/main/java/io/cdap/wrangler/service/directive/WorkspaceHandler.java
@@ -117,9 +117,6 @@ public class WorkspaceHandler extends AbstractDirectiveHandler {
   @SuppressWarnings("unused")
   private Metrics metrics;
 
-  private static final String APPLY_RECIPE_SUCCESS_METRIC = "recipe.apply.success";
-  private static final String APPLY_RECIPE_FAILURE_METRIC = "recipe.apply.failure";
-
   @Override
   public void initialize(SystemHttpServiceContext context) throws Exception {
     super.initialize(context);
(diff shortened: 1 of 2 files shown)


wrangler-service/src/main/java/io/cdap/wrangler/service/directive/WorkspaceHandler.java
```diff
@@ -117,9 +117,6 @@ public class WorkspaceHandler extends AbstractDirectiveHandler {
   @SuppressWarnings("unused")
   private Metrics metrics;
 
-  private static final String APPLY_RECIPE_SUCCESS_METRIC = "recipe.apply.success";
-  private static final String APPLY_RECIPE_FAILURE_METRIC = "recipe.apply.failure";
```
sau42shri, line 121: Which file/function are these metrics being captured right now?

iam-divya (approved):
LGTM

wrangler-service/src/main/java/io/cdap/wrangler/service/directive/WorkspaceHandler.java
saimukkamala: CDAP by default captures metrics for all HTTP API calls. `io.cdap.cdap.common.metrics.MetricsReporterHook` in `cdap` repo is responsible for capturing these metrics.

sau42shri: But would it be able to capture specific metric like count of directives in a recipe. I understand it could capture "ApplyRecipe" as that is a http call specific metric.

sau42shri:
LGTM

sau42shri (commented):
LGTM

PR #609, opened 2023-02-06
[PLUGIN-1528] Add support to specify scale and rounding-mode for decimal datatype in set-type directive

vanathi-g:
Currently, there is no way to set the precision and scale of converted decimal values when using the set-type directive. 
## Changes
- Add `scale` and `rounding-mode` as optional arguments to the set-type directive
- `rounding-mode` maps to the different rounding modes available in Java and is used when the decimal value's scale is not equal to the set scale.
- By default, no value for scale is assumed (current behaviour) and `round-mode` = HALF_EVEN is assumed when scale is set.
- Modified unit test that checks conversion to decimal type and directive parsing. Added necessary new tests.

Note: Support to specify expected precision will be added later.


wrangler-core/src/main/java/io/cdap/directives/column/SetType.java
```diff
@@ -23,43 +23,62 @@
...
+    builder.define("round", TokenType.BOOLEAN, Optional.TRUE);
     return builder.build();
   }
 
   @Override
   public void initialize(Arguments args) {
     col = ((ColumnName) args.value("column")).value();
     type = ((Identifier) args.value("type")).value();
+    if (type.equalsIgnoreCase("decimal")) {
+      scale = args.contains("scale") ? ((Numeric) args.value("scale")).value().intValue() : DEFAULT_SCALE;
```
sau42shri: Although this is ok for smaller set of arguments. In this case, you are searching the arguments twice for "scale", since both contains and value will search for the input. Instead, you can simply take the value in a param and use that for null (similar to TryGetValue you have in dictionaries). You can choose to ignore this comment, but would be good to put this into practice for any such search pattern.

iam-divya: Arguments is implemented as a map here. 'Contains' map to Map's containsKey(https://github.com/data-integrations/wrangler/blob/8be2f88adc4fb69fb8c103301228d32f35bdea96/wrangler-core/src/main/java/io/cdap/wrangler/parser/MapArguments.java#L153) and value maps to map's get function which does not invoke containsKey again.

wrangler-core/src/main/java/io/cdap/wrangler/utils/ColumnConverter.java
```diff
@@ -210,25 +213,30 @@ private static Object convertType(String col, String toType, Object object) thro
...
+          decimal = BigDecimal.valueOf((Double) object);
         } else if (object instanceof byte[]) {
-          return Bytes.toBigDecimal((byte[]) object);
-        } else {
+          decimal = Bytes.toBigDecimal((byte[]) object);
+        }
+        if (decimal == null) {
           break;
+        } else if (roundingMode.equals(RoundingMode.UNNECESSARY) && (decimal.scale() != scale)) {
+          throw new DirectiveExecutionException("Input value's scale is not equal to set output schema scale");
```
iam-divya: Why is this line needed? Flow from UI(not command line), these new args wont be set and do we throw an exception in that case?

wrangler-core/src/main/java/io/cdap/wrangler/utils/ColumnConverter.java
```diff
@@ -210,25 +213,30 @@ private static Object convertType(String col, String toType, Object object) thro
       }
 
       case "DECIMAL": {
+        BigDecimal decimal = null;
```
iam-divya: nit: Do you want to call the variable something else? like convertedObject or something else instead of 'decimal'?

wrangler-core/src/test/java/io/cdap/directives/column/SetTypeTest.java
```diff
@@ -154,9 +155,9 @@ public void testToDecimal() throws Exception {
         .add("bytes_col", new byte[] {0, 0, 0, 0, 39, 16}).add("decimal_col", new BigDecimal("10000"))
     );
     String[] directives = new String[] {
-      "set-type str_col decimal", "set-type int_col Decimal", "set-type double_col DECIMAL",
-      "set-type short_col DECIMAL", "set-type long_col Decimal", "set-type float_col DECIMAL",
-      "set-type bytes_col decimal", "set-type decimal_col decimal"
+      "set-type str_col decimal 0", "set-type int_col Decimal 0", "set-type double_col DECIMAL 1",
```
iam-divya: We should also be supporting just, 'set-type str_col decimal' without any scale or rounding args

vanathi-g: yes, that is supported

wrangler-core/src/main/java/io/cdap/directives/column/SetType.java
```diff
@@ -22,44 +22,81 @@
...
+
 
   @Override
   public UsageDefinition define() {
     UsageDefinition.Builder builder = UsageDefinition.builder(NAME);
     builder.define("column", TokenType.COLUMN_NAME);
     builder.define("type", TokenType.IDENTIFIER);
+    builder.define("scale", TokenType.NUMERIC, Optional.TRUE);
+    builder.define("round", TokenType.BOOLEAN, Optional.TRUE);
+    builder.define("rounding-mode", TokenType.TEXT, Optional.TRUE);
```
tivv, line 67: Why do we need both arguments for rounding? It looks more complex than having 1. 
Also, we should allow to set precision as well, should not we?

wrangler-core/src/main/java/io/cdap/directives/column/SetType.java
```diff
@@ -22,44 +22,81 @@
...
 @Name(SetType.NAME)
 @Categories(categories = {"column"})
-@Description("Converting data type of a column.")
+@Description("Converting data type of a column. Optional arguments scale, " +
+  "round and rounding-mode are used only when type is decimal.")
 public final class SetType implements Directive, Lineage {
   public static final String NAME = "set-type";
+
+  // Default scale that is assumed in CDAP output schema
+  public static final int DEFAULT_SCALE = 3;
```
tivv: By default we should do exactly as we did before. We can't change the existing logic

wrangler-core/src/main/java/io/cdap/directives/column/SetType.java
vanathi-g: 1. Two arguments for rounding because if the user isn't familiar with Java rounding modes, they can just enable rounding and we can assume HALF_UP. 
2. We can't set a 'max' precision on a Java BigDecimal object. Since Wrangler infers datatypes for the output schema from the actual row values, we can't propagate the set precision to the output schema.

wrangler-core/src/main/java/io/cdap/directives/column/SetType.java
vanathi-g: This is the default scale used when rounding is enabled but scale is not specified. The existing logic is not changed.

iam-divya:
LGTM. Please wait for Vitalii's LGTM before merging.

iam-divya (approved):

wrangler-core/src/main/java/io/cdap/directives/column/SetType.java
tivv: I don't think introducing two parameter makes it simpler. Any assumptions is source of confusion. And it's not java-specific, you can find more or less same set everywhere, starting from https://en.wikipedia.org/wiki/Rounding

wrangler-core/src/main/java/io/cdap/directives/column/SetType.java
tivv: I believe now it just convert to BigDecimal. E.g. currently if you have `1.23456` value it will remain `1.23456` value and would work if you set schema properly. With this change it's going to fail as default scale is 3 and  RoundingMode.UNNECESSARY.

tivv (changes requested):
Please see comments inline.

wrangler-core/src/main/java/io/cdap/directives/column/SetType.java
vanathi-g: Done, removed the 'round' argument

wrangler-core/src/main/java/io/cdap/directives/column/SetType.java
vanathi-g: Changed the code to preserve current logic.

wrangler-core/src/main/java/io/cdap/directives/column/SetType.java
```diff
@@ -22,44 +22,69 @@
...
     col = ((ColumnName) args.value("column")).value();
     type = ((Identifier) args.value("type")).value();
+    if (type.equalsIgnoreCase("decimal")) {
+      scale = args.contains("scale") ? ((Numeric) args.value("scale")).value().intValue() : null;
+      if (scale == null && args.contains("rounding-mode")) {
+        throw new DirectiveParseException("'rounding-mode' can only be specified when a 'scale' is set");
+      }
+      try {
+        roundingMode = args.contains("rounding-mode") ?
+          RoundingMode.valueOf(((Text) args.value("rounding-mode")).value()) : RoundingMode.UNNECESSARY;
```
tivv: This is actually where we can have a default value of, say, https://docs.oracle.com/javase/7/docs/api/java/math/RoundingMode.html#HALF_EVEN. This would make it same as floating point rounding and users would be able to use "settype decimal, 2" without rounding mode with reasonal default and without errors

wrangler-core/src/main/java/io/cdap/directives/datamodel/DataModelMapColumn.java
```diff
@@ -151,7 +152,7 @@ public void initialize(Arguments args) throws DirectiveParseException {
   @Override
   public List<Row> execute(List<Row> rows, ExecutorContext context) throws DirectiveExecutionException {
     for (Row row : rows) {
-      ColumnConverter.convertType(NAME, row, column, targetFieldTypeName);
+      ColumnConverter.convertType(NAME, row, column, targetFieldTypeName, null, RoundingMode.UNNECESSARY);
```
tivv, line 155: Do you really need to change it? If so, add a test. I am not sure what `args.contains("scale")` would return and if you would get an NPE.
BTW: It would be nice to handle null scale as well same as no scale

wrangler-core/src/main/java/io/cdap/wrangler/utils/ColumnConverter.java
```diff
@@ -284,4 +292,16 @@ private static Object convertType(String col, String toType, Object object) thro
...
+
+  private static BigDecimal setDecimalScale(BigDecimal decimal, Integer scale, RoundingMode roundingMode)
+    throws DirectiveExecutionException {
+    if (scale == null) {
+      return decimal;
+    }
+    try {
+      return decimal.setScale(scale, roundingMode);
+    } catch (ArithmeticException e) {
+      throw new DirectiveExecutionException(e);
```
tivv: Please add a proper error message similar to line 291

tivv (commented):
Please also update all documentation (`.md`) files where the directive is refered

wrangler-core/src/main/java/io/cdap/directives/datamodel/DataModelMapColumn.java
vanathi-g: Setting it as null is alright because convertType is a utility method where setting scale as null is handled anyways. In any case, args.contains("scale") returns false when scale is null (ex, when scale isn't specified in the directive). Null scale is the same as no scale.

wrangler-core/src/main/java/io/cdap/directives/column/SetType.java
vanathi-g: Why HALF_EVEN? Initially I assumed the default rounding mode as HALF_UP but the idea is to keep the current behaviour (no rounding if the scale is different) the same right?

wrangler-core/src/main/java/io/cdap/wrangler/utils/ColumnConverter.java
vanathi-g: Done

wrangler-core/src/main/java/io/cdap/directives/column/SetType.java
tivv: According to the documentation HALF_EVEN is the one used in IEEE-754. Also it has better math properties (statistically neutral). 

Yes, we should keep the current behavior unchanged - when scale is not set, we should do as before. Now if scale is set, we are free to provide a good default for the rounding mode as this is new behavior

tivv: But for the default rounding mode it would be great to have references on what other systems do

vanathi-g: Makes sense. Done, changed the default mode to HALF_EVEN.

wrangler-core/src/main/java/io/cdap/directives/column/SetType.java
```diff
@@ -22,44 +22,70 @@
...
+@Description("Converting data type of a column. Optional arguments scale and rounding-mode " +
+  "are used only when type is decimal.")
 public final class SetType implements Directive, Lineage {
   public static final String NAME = "set-type";
+
   private String col;
   private String type;
+  private Integer scale;
+  private RoundingMode roundingMode;
+
```
tivv, line 60: nit: double empty line

wrangler-core/src/main/java/io/cdap/directives/column/SetType.java
```diff
@@ -22,44 +22,70 @@
...
+      scale = args.contains("scale") ? ((Numeric) args.value("scale")).value().intValue() : null;
+      if (scale == null && args.contains("rounding-mode")) {
+        throw new DirectiveParseException("'rounding-mode' can only be specified when a 'scale' is set");
+      }
+      try {
+        roundingMode = args.contains("rounding-mode") ?
+          RoundingMode.valueOf(((Text) args.value("rounding-mode")).value()) :
+          (scale == null ? RoundingMode.UNNECESSARY : RoundingMode.HALF_EVEN);
+      } catch (IllegalArgumentException e) {
+        throw new DirectiveParseException("Specified rounding-mode is not a valid Java rounding mode");
```
tivv: Add the value you are parsing to the message and add `e` as cause

wrangler-core/src/main/java/io/cdap/wrangler/parser/MigrateToV2.java
```diff
@@ -126,11 +126,13 @@ public String migrate() throws DirectiveParseException {
         }
         break;
 
-        //set-type <column> <type>
+        //set-type <column> <type> [<scale> <round> <rounding-mode>]
```
tivv: Adjust docs to the latest revision

wrangler-core/src/main/java/io/cdap/wrangler/utils/ColumnConverter.java
```diff
@@ -284,4 +292,17 @@ private static Object convertType(String col, String toType, Object object) thro
...
+  private static BigDecimal setDecimalScale(BigDecimal decimal, Integer scale, RoundingMode roundingMode)
+    throws DirectiveExecutionException {
+    if (scale == null) {
+      return decimal;
+    }
+    try {
+      return decimal.setScale(scale, roundingMode);
+    } catch (ArithmeticException e) {
+      throw new DirectiveExecutionException(
+        String.format("'rounding-mode' cannot be '%s' for the specified scaling operation", roundingMode));
```
tivv: Add `e` as a cause.  Also I don't think the message is telling the story. The most often case would be using RoundingMode.UNNESSESARY and the rounding is needed. It's better to say something about the value and scale tried, e.g. that scale can't be set to X for value Y with rounding mode Z

tivv (commented):
Last polishing, generally looks good. 
Please also update all documentation files (*.md) that refer to the directive (fast search shows 5 files with the directive).

tivv:
And add the JIRA reference to the title

wrangler-core/src/main/java/io/cdap/directives/column/SetType.java
vanathi-g: Done.

wrangler-core/src/main/java/io/cdap/wrangler/parser/MigrateToV2.java
vanathi-g: Done, changed the necessary readme files too.

wrangler-core/src/main/java/io/cdap/wrangler/utils/ColumnConverter.java
vanathi-g: Done, added a better error message

tivv (approved):

PR #613, opened 2023-03-07
Implementation of BQ filter pushdown for Wrangler precondition

DJSagarAhire:
This PR adds support for relational execution of filter expressions in Wrangler's precondition. This doesn't add any user-facing change yet, but once this flow is invoked, the expression is ready to be pushed down to either Spark or BQ (or any other engine as required). The user-facing changes will follow in another PR focusing on Wrangler.java.

The condition specified is assumed to be written in ANSI SQL and no validation or conversion will be performed.

### DIFF
--- a/wrangler-transform/src/main/java/io/cdap/wrangler/Precondition.java
+++ b/wrangler-transform/src/main/java/io/cdap/wrangler/Precondition.java
@@ -16,9 +16,17 @@
 
 package io.cdap.wrangler;
 
+import io.cdap.cdap.etl.api.relational.Expression;
+import io.cdap.cdap.etl.api.relational.ExpressionFactory;
+import io.cdap.cdap.etl.api.relational.InvalidRelation;
+import io.cdap.cdap.etl.api.relational.LinearRelationalTransform;
+import io.cdap.cdap.etl.api.relational.Relation;
+import io.cdap.cdap.etl.api.relational.RelationalTranformContext;
+import io.cdap.cdap.etl.api.relational.StringExpressionFactoryType;
 import io.cdap.wrangler.api.Row;
 import org.apache.commons.jexl3.scripting.JexlScriptEngine;
 
+import java.util.Optional;
 import javax.script.Bindings;
 import javax.script.CompiledScript;
 import javax.script.ScriptContext;
@@ -29,14 +37,25 @@ import javax.script.SimpleScriptContext;
 /**
  * A precondition expression that filters data into the directives.
  */
-public class Precondition {
+public class Precondition implements LinearRelationalTransform {
   private final String condition;
   private final CompiledScript script;
   // SimpleScriptContext is pretty expensive to construct due to all PrintWriter creation, so let's cache it
   private final ThreadLocal<ScriptContext> contextCache = ThreadLocal.withInitial(this::createContext);
+  boolean isConditionSQL;
 
   public Precondition(String condition) throws PreconditionException {
+    this(condition, false);
+  }
+
+  public Precondition(String condition, boolean isConditionSQL) throws PreconditionException {
     this.condition = condition;
+    this.isConditionSQL = isConditionSQL;
+
+    if (isConditionSQL) {
+      script = null;
+      return;
+    }
     JexlScriptEngine engine = new JexlScriptEngine();
     try {
       script = engine.compile(condition);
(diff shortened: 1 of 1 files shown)


wrangler-transform/src/main/java/io/cdap/wrangler/Precondition.java
```diff
@@ -88,4 +98,32 @@ public boolean apply(Row row) throws PreconditionException {
...
+    return relation.filter(compiledExpression);
+  }
+
+  // TODO: Find out from where we can get schema information. Perhaps RelationalTranformContext
+  private Schema getSchema() {
+    return null;
+  }
+
+  // TODO: Find how this info can be obtained. Perhaps RelationalTranformContext
+  boolean isPushdownEnabled() {
```
fernst: Pushdown will always be enabled. The Spark engine is also a pushdown engine, so the answer will always be yes.

The dialect needs to be selected by the SQL Engine implementation.

wrangler-transform/src/main/java/io/cdap/wrangler/Precondition.java
```diff
@@ -88,4 +98,32 @@ public boolean apply(Row row) throws PreconditionException {
...
+  @Override
+  public Relation transform(RelationalTranformContext relationalTranformContext, Relation relation) {
+    if (!isConditionSQL) {
+      return new InvalidRelation("Condition is not specified as SQL");
+    }
+
+    Schema inputSchema = getSchema();
+
+    ExpressionConversionContext context = new ExpressionConversionContext();
+    SQLDialect dest = isPushdownEnabled() ? SQLDialect.BIGQUERYSQL : SQLDialect.SPARKSQL;
```
fernst: Why should the destination be known here? The engine will decide what SQLDialect it wants the expression in.

wrangler-transform/src/main/java/io/cdap/wrangler/mock/SQLDialect.java
```diff
@@ -0,0 +1,26 @@
...
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package io.cdap.wrangler.mock;
+
+/**
+ * Mock implementation for SQLDialect.
+ */
+public enum SQLDialect {
```
fernst: This should come from the platform.

fernst (changes requested):
Why is the destination dialect relevant to Wrangler? All wrangler needs to know is that the expression is valid. The extraction of this expression will be handled by the pushdown plugin.

wrangler-transform/src/main/java/io/cdap/wrangler/mock/SQLDialect.java
DJSagarAhire: It will. I've just created a mock implementation as the platform PR isn't merged yet.

wrangler-transform/src/main/java/io/cdap/wrangler/Precondition.java
```diff
@@ -88,4 +98,32 @@ public boolean apply(Row row) throws PreconditionException {
...
+
+  @Override
+  public Relation transform(RelationalTranformContext relationalTranformContext, Relation relation) {
+    if (!isConditionSQL) {
+      return new InvalidRelation("Condition is not specified as SQL");
+    }
+
+    Schema inputSchema = getSchema();
+
+    ExpressionConversionContext context = new ExpressionConversionContext();
```
tivv: It's not clear where this would come from. It should come from the platform / app somehow, so it should be part of the relationalTranformContext. The main ask for this PR is to fully lay out the API / platform part, not the wrangler part

wrangler-transform/src/main/java/io/cdap/wrangler/Precondition.java
DJSagarAhire: Resolving this for now as this is not relevant for ANSI SQL conditions.

wrangler-transform/src/main/java/io/cdap/wrangler/mock/SQLDialect.java
DJSagarAhire: Resolving this for now as this is not relevant for ANSI SQL conditions.

wrangler-transform/src/main/java/io/cdap/wrangler/Precondition.java
DJSagarAhire: Resolving this for now as this is not relevant for ANSI SQL conditions.

wrangler-transform/src/main/java/io/cdap/wrangler/Precondition.java
DJSagarAhire: Changed accordingly.

wrangler-transform/src/main/java/io/cdap/wrangler/Precondition.java
```diff
@@ -29,23 +37,34 @@
...
-    try {
-      script = engine.compile(condition);
-    } catch (ScriptException e) {
-      if (e.getCause() != null) {
-        throw new PreconditionException(e.getCause().getMessage());
-      } else {
-        throw new PreconditionException(e.getMessage());
+    this.isConditionSQL = isConditionSQL;
+
+    if (!isConditionSQL) {
```
fernst: nit: write this as a simple return.

```
if (isConditionSQL( {
  script = null;
  return;
}

...
```

fernst (changes requested):
The UI piece is still missing, please implement the UI changes as well.

Also, can you add a few tests if possible?

fernst (approved):
LGTM, add tests please.

PR #614, opened 2023-03-14
Add metrics for directives

vanathi-g:
The following new metrics are collected per pipeline run to help improve Wrangler directives:
- Number of times a directive is used
- Number of times a JEXL function category is used

## Changes
- Add utility classes and a new interface to simplify emitting metrics from Wrangler directives
- Add functions to emit metrics from the prepareRun method in Wrangler plugin class

### DIFF
--- a/wrangler-api/pom.xml
+++ b/wrangler-api/pom.xml
@@ -24,7 +24,7 @@
 
   <artifactId>wrangler-api</artifactId>
   <name>Wrangler API</name>
-  <description>Wrangler API Libaray.</description>
+  <description>Wrangler API Library.</description>
 
   <dependencies>
     <dependency>
@@ -36,6 +36,12 @@
     <dependency>
       <groupId>io.cdap.cdap</groupId>
       <artifactId>cdap-etl-api</artifactId>
+      <version>${cdap.version}</version>
+      <scope>provided</scope>
+    </dependency>
+    <dependency>
+      <groupId>io.cdap.cdap</groupId>
+      <artifactId>cdap-watchdog-api</artifactId>
       <version>${cdap.version}</version>
       <scope>provided</scope>
     </dependency>
--- a/wrangler-api/src/main/java/io/cdap/wrangler/api/EntityMetrics.java
+++ b/wrangler-api/src/main/java/io/cdap/wrangler/api/EntityMetrics.java
@@ -1,0 +1,31 @@
+/*
+ * Copyright © 2023 Cask Data, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package io.cdap.wrangler.api;
+
+import java.util.List;
+
+/**
+ * Metrics emitted by a Wrangler entity
+ */
+public interface EntityMetrics {
+  /**
+   * This method can be used to return a list of metrics to be emitted by a Wrangler entity (ex: {@link Directive}).
+   * Note that this method doesn't emit metrics, it only returns metadata to be used in metrics emission logic elsewhere
+   * @return list of {@link EntityMetricDef}s to be emitted
+   */
+  List<EntityMetricDef> getMetrics();
+}
--- a/wrangler-core/src/main/java/io/cdap/directives/row/Fail.java
+++ b/wrangler-core/src/main/java/io/cdap/directives/row/Fail.java
@@ -16,6 +16,7 @@
 
 package io.cdap.directives.row;
 
+import com.google.common.collect.ImmutableList;
 import io.cdap.cdap.api.annotation.Description;
 import io.cdap.cdap.api.annotation.Name;
 import io.cdap.cdap.api.annotation.Plugin;
@@ -23,6 +24,7 @@ import io.cdap.wrangler.api.Arguments;
 import io.cdap.wrangler.api.Directive;
 import io.cdap.wrangler.api.DirectiveExecutionException;
 import io.cdap.wrangler.api.DirectiveParseException;
+import io.cdap.wrangler.api.EntityMetricDef;
 import io.cdap.wrangler.api.ExecutorContext;
 import io.cdap.wrangler.api.Row;
 import io.cdap.wrangler.api.annotations.Categories;
@@ -35,6 +37,7 @@ import io.cdap.wrangler.expression.EL;
 import io.cdap.wrangler.expression.ELContext;
 import io.cdap.wrangler.expression.ELException;
 import io.cdap.wrangler.expression.ELResult;
+import io.cdap.wrangler.metrics.DirectiveJEXLCategoryMetric;
 
 import java.util.List;
 
@@ -45,7 +48,7 @@ import java.util.List;
 @Name(Fail.NAME)
 @Categories(categories = { "row", "data-quality"})
 @Description("Fails when the condition is evaluated to true.")
-public class Fail implements Directive, Lineage {
+public class Fail implements Directive, Lineage, DirectiveJEXLCategoryMetric {
   public static final String NAME = "fail";
   private String condition;
   private EL el;
@@ -105,5 +108,11 @@ public class Fail implements Directive, Lineage {
                                  .readable(String.format("Pipeline set to fail based on condition '%s'", condition));
     el.variables().forEach(col -> builder.relation(col, col));
     return builder.build();
+  }
+
+  @Override
+  public List<EntityMetricDef> getMetrics() {
+    EntityMetricDef jexlCategoryMetric = getJEXLCategoryMetric(el.getScriptParsedText());
+    return (jexlCategoryMetric == null) ? ImmutableList.of() : ImmutableList.of(jexlCategoryMetric);
   }
 }
--- a/wrangler-core/src/main/java/io/cdap/directives/row/RecordConditionFilter.java
+++ b/wrangler-core/src/main/java/io/cdap/directives/row/RecordConditionFilter.java
@@ -16,6 +16,7 @@
 
 package io.cdap.directives.row;
 
+import com.google.common.collect.ImmutableList;
 import io.cdap.cdap.api.annotation.Description;
 import io.cdap.cdap.api.annotation.Name;
 import io.cdap.cdap.api.annotation.Plugin;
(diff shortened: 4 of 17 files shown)


wrangler-core/src/main/java/io/cdap/directives/transformation/ColumnExpression.java
```diff
@@ -123,6 +130,17 @@ public Mutation lineage() {
...
   }
-}
-
 
+  @Override
+  public List<DirectiveMetric> getMetrics() {
+    String jexlFunction = el.script.getParsedText().split(":")[0].trim();
+    if (elDefaultFunctions.containsKey(jexlFunction)) {
+      DirectiveMetric directiveMetric = new DirectiveMetric("wrangler.jexl-category.count", ImmutableMap.of(
+        "srv", "dataprep",
```
iam-divya: Do you want to define constants for srv, aet etc.,

wrangler-transform/src/main/java/io/cdap/wrangler/Wrangler.java
```diff
@@ -482,6 +490,28 @@ private RecipeParser getRecipeParser(StageContext context)
...
+          Metrics child = metrics.child(directiveMetric.getTags());
+          // make this common to both count and gauge
+          child.count(directiveMetric.getMetricName(), 1);
+        }
+      }
+      emitDirectiveUsedMetric(directive.define().getDirectiveName(), metrics);
+    }
+  }
+
+  private void emitDirectiveUsedMetric(String directiveName, Metrics metrics) {
```
iam-divya: Lets add some javadoc on how we are mapping the wrangler metrics into common entities of CDAP.

wrangler-api/src/main/java/io/cdap/wrangler/api/DirectiveWithMetrics.java
```diff
@@ -0,0 +1,26 @@
...
+ */
+
+package io.cdap.wrangler.api;
+
+import java.util.List;
+
+/**
+ * A {@link Directive} that can emit metrics
+ */
+public interface DirectiveWithMetrics extends Directive {
```
iam-divya: Should this be implementing Directive? Can this not be another interface that Directive impl can implement when custom metrics are needed to be defined? 
interface DirectiveMetricDef: getServiceName, getAppEntityType, getAppEntityTypeValue
class DefaultDirectiveMetricDef impl DirectiveMetricDef with a constructor that takes in directive name with hardcoded service name and aet? 
ColumnExpression can implement DirectiveMetricDef with custom values. 
What do you think about this approach? 
(Redesign this as list of metric where needed)

wrangler-core/src/main/java/io/cdap/wrangler/expression/EL.java
```diff
@@ -55,7 +55,7 @@ public final class EL {
   private static volatile boolean used;
 
   private final Set<String> variables;
-  private final JexlScript script;
+  public final JexlScript script;
```
saimukkamala: Instead making this variable public, we should return the parsed text (which is a String) of the JexlScript in a public method

wrangler-transform/src/main/java/io/cdap/wrangler/Wrangler.java
```diff
@@ -482,6 +496,45 @@ private RecipeParser getRecipeParser(StageContext context)
...
+      }
+
+      if (directive instanceof EntityWithMetrics) {
+        wranglerMetrics.addAll(((EntityWithMetrics) directive).getMetrics());
+      }
+
+      for (WranglerMetric wranglerMetric : wranglerMetrics) {
+        Metrics child = metrics.child(wranglerMetric.getTags());
+        if (wranglerMetric.getType().equals(MetricType.COUNTER)) {
+          child.count(wranglerMetric.getName(), (int) wranglerMetric.getValue());
```
saimukkamala: we should use `countLong` instead of `count` rather than casting long to int

wrangler-api/src/main/java/io/cdap/wrangler/api/WranglerMetric.java
```diff
@@ -0,0 +1,54 @@
...
+package io.cdap.wrangler.api;
+
+import io.cdap.cdap.api.metrics.MetricType;
+
+import java.util.Map;
+
+/**
+ * Represents generic metadata information for any metric that is emitted in Wrangler
+ */
+public class WranglerMetric {
```
saimukkamala: nit: `EntityMetric` would make more sense since metrics are emitted by an entity which implements `EntityWithMetrics` interface

wrangler-core/src/main/java/io/cdap/directives/row/Fail.java
```diff
@@ -106,4 +109,13 @@ public Mutation lineage() {
     el.variables().forEach(col -> builder.relation(col, col));
     return builder.build();
   }
+
+  @Override
+  public List<WranglerMetric> getMetrics() {
+    WranglerMetric jexlCategoryMetric = getJEXLCategoryMetric(el.script.getParsedText());
+    if (jexlCategoryMetric != null) {
+      return ImmutableList.of(jexlCategoryMetric);
```
saimukkamala: nit:
` if (jexlCategoryMetric != null) {
      return ImmutableList.of(jexlCategoryMetric);
    }
    return ImmutableList.of();` 
    replace with single statement:
`return (jexlCategoryMetric == null) ? ImmutableList.of() : ImmutableList.of(jexlCategoryMetric);`

wrangler-transform/src/main/java/io/cdap/wrangler/Wrangler.java
```diff
@@ -482,6 +496,45 @@ private RecipeParser getRecipeParser(StageContext context)
...
+        wranglerMetrics.add(getDirectiveUsageMetric(directive.define().getDirectiveName()));
+      }
+
+      if (directive instanceof EntityWithMetrics) {
+        wranglerMetrics.addAll(((EntityWithMetrics) directive).getMetrics());
+      }
+
+      for (WranglerMetric wranglerMetric : wranglerMetrics) {
+        Metrics child = metrics.child(wranglerMetric.getTags());
+        if (wranglerMetric.getType().equals(MetricType.COUNTER)) {
```
saimukkamala: we should use switch case for readability

wrangler-api/pom.xml
```diff
@@ -39,6 +39,12 @@
       <version>${cdap.version}</version>
       <scope>provided</scope>
     </dependency>
+    <dependency>
+      <groupId>io.cdap.cdap</groupId>
+      <artifactId>cdap-watchdog-api</artifactId>
+      <version>6.9.0-SNAPSHOT</version>
```
iam-divya: We should use ${cdap.version} here

wrangler-api/src/main/java/io/cdap/wrangler/api/WranglerMetric.java
iam-divya: +1, I would call this EntityMetricDef as it is confusing between EntityMetric vs EntityWithMetric. Also EntityWithMetric can be EntityMetric.

wrangler-core/src/main/java/io/cdap/directives/row/RecordConditionFilter.java
```diff
@@ -119,4 +122,13 @@ public Mutation lineage() {
     el.variables().forEach(column -> builder.relation(column, column));
     return builder.build();
   }
+
+  @Override
+  public List<WranglerMetric> getMetrics() {
```
iam-divya: Is the same implementation repeating in multiple places? Do you want to make this as default implementation?

wrangler-core/src/main/java/io/cdap/wrangler/metrics/DirectiveWithJEXLCategoryMetric.java
```diff
@@ -0,0 +1,59 @@
...
+import static io.cdap.cdap.api.metrics.MetricType.COUNTER;
+import static io.cdap.cdap.common.conf.Constants.Metrics.Tag.APP_ENTITY_TYPE;
+import static io.cdap.cdap.common.conf.Constants.Metrics.Tag.APP_ENTITY_TYPE_NAME;
+import static io.cdap.cdap.common.conf.Constants.Metrics.Tag.SERVICE;
+
+/**
+ * Represents a directive that emits the 'JEXL category' metric. This metric counts the number of times
+ * a JEXL function category in {@link io.cdap.wrangler.expression.EL.DefaultFunctions} is used.
+ */
+public interface DirectiveWithJEXLCategoryMetric extends EntityWithMetrics {
```
iam-divya: nit: DirectiveWithJEXLCategoryMetric can be DirectiveJEXLCategoryMetric or JEXLCategoryDirectiveMetric

wrangler-core/src/main/java/io/cdap/directives/row/RecordConditionFilter.java
saimukkamala: afaik I believe we can't make this default since the information required for building metrics comes from implementation of the Directive.

vanathi-g: Yes, the required JEXL script is in the directive implementation

wrangler-core/src/main/java/io/cdap/wrangler/expression/EL.java
vanathi-g: Done

wrangler-api/src/main/java/io/cdap/wrangler/api/WranglerMetric.java
vanathi-g: I have renamed the classes to `WranglerEntityMetric` and `WranglerEntityWithMetrics` where the latter interface represents an entity in wrangler that is capable of emitting `WranglerEntityMetric`s

wrangler-core/src/main/java/io/cdap/directives/row/Fail.java
vanathi-g: Done

wrangler-api/pom.xml
vanathi-g: Done

wrangler-transform/src/main/java/io/cdap/wrangler/Wrangler.java
vanathi-g: Done

wrangler-transform/src/main/java/io/cdap/wrangler/Wrangler.java
vanathi-g: Done

wrangler-api/src/main/java/io/cdap/wrangler/api/WranglerMetric.java
vanathi-g: Renamed on further discussion to EntityMetricDef and EntityMetrics

saimukkamala (approved):

wrangler-api/src/main/java/io/cdap/wrangler/api/EntityMetrics.java
```diff
@@ -0,0 +1,31 @@
...
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package io.cdap.wrangler.api;
+
+import java.util.List;
+
+/**
+ * A Wrangler entity that can emit metrics
```
iam-divya: nit: Metrics emitted by the Wrangler Entity

iam-divya (approved):

PR #615, opened 2023-03-15
CDAP-20454: UI changes in Wrangler to toggle between SQL and JEXL.

akritig187:
**UI Changes in Wrangler to add a toggle between SQL and JEXL expressions** 

Screenshots:
image

image

image

image

### DIFF
--- a/wrangler-transform/src/main/java/io/cdap/wrangler/Wrangler.java
+++ b/wrangler-transform/src/main/java/io/cdap/wrangler/Wrangler.java
@@ -16,6 +16,7 @@
 
 package io.cdap.wrangler;
 
+import com.google.common.base.Strings;
 import com.google.common.collect.ImmutableMap;
 import io.cdap.cdap.api.annotation.Description;
 import io.cdap.cdap.api.annotation.Macro;
@@ -97,6 +98,10 @@ public class Wrangler extends Transform<StructuredRecord, StructuredRecord> {
   private static final String ON_ERROR_PROCEED = "send-to-error-port";
   private static final String ERROR_STRATEGY_DEFAULT = "wrangler.error.strategy.default";
 
+  // Precondition languages
+  private static final String PRECONDITION_LANGUAGE_JEXL = "jexl";
+  private static final String PRECONDITION_LANGUAGE_SQL = "sql";
+
   // Plugin configuration.
   private final Config config;
 
@@ -144,16 +149,26 @@ public class Wrangler extends Transform<StructuredRecord, StructuredRecord> {
 
     try {
       Schema iSchema = configurer.getStageConfigurer().getInputSchema();
-      if (!config.containsMacro(Config.NAME_FIELD) && !(config.field.equals("*") || config.field.equals("#"))) {
+      if (!config.containsMacro(Config.NAME_FIELD) && !(config.getField().equals("*")
+        || config.getField().equals("#"))) {
         validateInputSchema(iSchema, collector);
       }
 
-      String directives = config.directives;
-      if (config.udds != null && !config.udds.trim().isEmpty()) {
+      String directives = config.getDirectives();
+      if (config.getUDDs() != null && !config.getUDDs().trim().isEmpty()) {
         if (config.containsMacro("directives")) {
-          directives = String.format("#pragma load-directives %s;", config.udds);
+          directives = String.format("#pragma load-directives %s;", config.getUDDs());
         } else {
-          directives = String.format("#pragma load-directives %s;%s", config.udds, config.directives);
+          directives = String.format("#pragma load-directives %s;%s", config.getUDDs(), config.getDirectives());
+        }
+      }
+
+      if (!config.containsMacro(Config.NAME_PRECONDITION_LANGUAGE)) {
+        if (PRECONDITION_LANGUAGE_SQL.equalsIgnoreCase(config.getPreconditionLanguage())) {
+          validatePrecondition(config.getPreconditionSQL(), true, collector);
+          validateSQLModeDirectives(collector);
+        } else {
+          validatePrecondition(config.getPreconditionJEXL(), false, collector);
         }
       }
 
@@ -201,7 +216,6 @@ public class Wrangler extends Transform<StructuredRecord, StructuredRecord> {
             }
           }
         }
-
       } catch (CompileException e) {
         collector.addFailure("Compilation error occurred : " + e.getMessage(), null);
       } catch (DirectiveParseException e) {
@@ -218,13 +232,27 @@ public class Wrangler extends Transform<StructuredRecord, StructuredRecord> {
           .withConfigProperty(Config.NAME_SCHEMA).withStacktrace(e.getStackTrace());
       }
 
-      // Check if pre-condition is not null or empty and if so compile expression.
-      if (!config.containsMacro(Config.NAME_PRECONDITION)) {
-        if (config.precondition != null && !config.precondition.trim().isEmpty()) {
+      // Check if jexl pre-condition is not null or empty and if so compile expression.
+      if (!config.containsMacro(Config.NAME_PRECONDITION) && !config.containsMacro(Config.NAME_PRECONDITION_LANGUAGE)) {
+        if (PRECONDITION_LANGUAGE_JEXL.equalsIgnoreCase(config.getPreconditionLanguage())
+          && checkPreconditionNotEmpty(false)) {
           try {
-            new Precondition(config.precondition);
+            new Precondition(config.getPreconditionJEXL());
           } catch (PreconditionException e) {
             collector.addFailure(e.getMessage(), null).withConfigProperty(Config.NAME_PRECONDITION);
+          }
+        }
+      }
+
+      // Check if sql pre-condition is not null or empty and if so compile expression.
+      if (!config.containsMacro(Config.NAME_PRECONDITION_SQL)
+        && !config.containsMacro(Config.NAME_PRECONDITION_LANGUAGE)) {
+        if (PRECONDITION_LANGUAGE_SQL.equalsIgnoreCase(config.getPreconditionLanguage())
+          && checkPreconditionNotEmpty(true)) {
+          try {
+            new Precondition(config.getPreconditionSQL(), true);
+          } catch (PreconditionException e) {
+            collector.addFailure(e.getMessage(), null).withConfigProperty(Config.NAME_PRECONDITION_SQL);
           }
         }
       }
@@ -309,12 +337,27 @@ public class Wrangler extends Transform<StructuredRecord, StructuredRecord> {
       );
     }
 
-    // Check if pre-condition is not null or empty and if so compile expression.
-    if (config.precondition != null && !config.precondition.trim().isEmpty()) {
-      try {
-        condition = new Precondition(config.precondition);
-      } catch (PreconditionException e) {
-        throw new IllegalArgumentException(e.getMessage(), e);
+    // Check if jexl pre-condition is not null or empty and if so compile expression.
+    if (!config.containsMacro(Config.NAME_PRECONDITION_LANGUAGE)) {
+      if (PRECONDITION_LANGUAGE_JEXL.equalsIgnoreCase(config.getPreconditionLanguage())
+        && checkPreconditionNotEmpty(false)) {
+        try {
+          condition = new Precondition(config.getPreconditionJEXL());
+        } catch (PreconditionException e) {
+          throw new IllegalArgumentException(e.getMessage(), e);
+        }
+      }
+    }
+
+    // Check if sql pre-condition is not null or empty and if so compile expression.
+    if (!config.containsMacro(Config.NAME_PRECONDITION_LANGUAGE)) {
+      if (PRECONDITION_LANGUAGE_SQL.equalsIgnoreCase(config.getPreconditionLanguage())
+        && checkPreconditionNotEmpty(true)) {
+        try {
+          condition = new Precondition(config.getPreconditionSQL(), true);
+        } catch (PreconditionException e) {
+          throw new IllegalArgumentException(e.getMessage(), e);
+        }
       }
     }
 
@@ -357,12 +400,12 @@ public class Wrangler extends Transform<StructuredRecord, StructuredRecord> {
     try {
       // Creates a row as starting point for input to the pipeline.
       Row row = new Row();
-      if ("*".equalsIgnoreCase(config.field)) {
+      if ("*".equalsIgnoreCase(config.getField())) {
         row = StructuredToRowTransformer.transform(input);
-      } else if ("#".equalsIgnoreCase(config.field)) {
+      } else if ("#".equalsIgnoreCase(config.getField())) {
         row.add(input.getSchema().getRecordName(), input);
       } else {
-        row.add(config.field, StructuredToRowTransformer.getValue(input, config.field));
+        row.add(config.getField(), StructuredToRowTransformer.getValue(input, config.getField()));
       }
 
       // If pre-condition is set, then evaluate the precondition
@@ -451,12 +494,51 @@ public class Wrangler extends Transform<StructuredRecord, StructuredRecord> {
   private void validateInputSchema(@Nullable Schema inputSchema, FailureCollector collector) {
     if (inputSchema != null) {
       // Check the existence of field in input schema
-      Schema.Field inputSchemaField = inputSchema.getField(config.field);
+      Schema.Field inputSchemaField = inputSchema.getField(config.getField());
       if (inputSchemaField == null) {
-        collector.addFailure(String.format("Field '%s' must be present in input schema.", config.field), null)
+        collector.addFailure(String.format("Field '%s' must be present in input schema.", config.getField()), null)
           .withConfigProperty(Config.NAME_FIELD);
       }
     }
+  }
+
+  private void validatePrecondition(String precondition, Boolean isConditionSQL, FailureCollector collector) {
+    String field = Config.NAME_PRECONDITION;
+    String language = "Precondition (JEXL)";
+
+    if (isConditionSQL == true) {
+      field = Config.NAME_PRECONDITION_SQL;
+      language = "Precondition (SQL)";
+    }
+    if (Strings.isNullOrEmpty(precondition)) {
+      collector.addFailure(String.format("%s must be present.", language),
+                           null)
+        .withConfigProperty(field);
+    }
+  }
+
+  private void validateSQLModeDirectives(FailureCollector collector) {
+    if (!Strings.isNullOrEmpty(config.getDirectives())) {
+      collector.addFailure("Directives are not supported for precondition of type SQL", null)
+        .withConfigProperty(Config.NAME_DIRECTIVES);
+    }
+
+    if (!Strings.isNullOrEmpty(config.getUDDs())) {
+      collector.addFailure("UDDs are not supported for precondition of type SQL", null)
+        .withConfigProperty(Config.NAME_UDD);
+    }
+  }
+
+  private boolean checkPreconditionNotEmpty(Boolean isConditionSQL) {
+    if (!isConditionSQL && !Strings.isNullOrEmpty(config.getPreconditionJEXL())
+      && !config.getPreconditionJEXL().trim().isEmpty()) {
+      return true;
+    }
+    if (isConditionSQL && !Strings.isNullOrEmpty(config.getPreconditionSQL())
+      && !config.getPreconditionSQL().trim().isEmpty()) {
+      return true;
+    }
+    return false;
   }
 
   /**
@@ -474,9 +556,9 @@ public class Wrangler extends Transform<StructuredRecord, StructuredRecord> {
     registry = new CompositeDirectiveRegistry(SystemDirectiveRegistry.INSTANCE, new UserDirectiveRegistry(context));
     registry.reload(context.getNamespace());
 
-    String directives = config.directives;
-    if (config.udds != null && !config.udds.trim().isEmpty()) {
-      directives = String.format("#pragma load-directives %s;%s", config.udds, config.directives);
+    String directives = config.getDirectives();
+    if (config.getUDDs() != null && !config.getUDDs().trim().isEmpty()) {
+      directives = String.format("#pragma load-directives %s;%s", config.getUDDs(), config.getDirectives());
     }
 
     return new GrammarBasedParser(context.getNamespace(), new MigrateToV2(directives).migrate(), registry);
@@ -487,16 +569,32 @@ public class Wrangler extends Transform<StructuredRecord, StructuredRecord> {
    */
   public static class Config extends PluginConfig {
     static final String NAME_PRECONDITION = "precondition";
+    static final String NAME_PRECONDITION_SQL = "preconditionSQL";
+    static final String NAME_PRECONDITION_LANGUAGE = "preconditionLanguage";
     static final String NAME_FIELD = "field";
     static final String NAME_DIRECTIVES = "directives";
     static final String NAME_UDD = "udd";
     static final String NAME_SCHEMA = "schema";
     static final String NAME_ON_ERROR = "on-error";
 
-    @Name(NAME_PRECONDITION)
-    @Description("Precondition expression specifying filtering before applying directives (true to filter)")
+    @Name(NAME_PRECONDITION_LANGUAGE)
+    @Description("Toggle to configure precondition language between JEXL and SQL")
     @Macro
+    private String preconditionLanguage;
+
+    @Name(NAME_PRECONDITION)
+    @Description("JEXL Precondition expression specifying filtering before applying directives (true to filter)")
+    @Macro
+    @Nullable
     private String precondition;
+
+
+    @Name(NAME_PRECONDITION_SQL)
+    @Description("SQL Precondition expression specifying filtering before applying directives (true to filter)")
+    @Macro
+    @Nullable
+    private String preconditionSQL;
+
 
     @Name(NAME_DIRECTIVES)
     @Description("Recipe for wrangling the input records")
@@ -525,11 +623,13 @@ public class Wrangler extends Transform<StructuredRecord, StructuredRecord> {
     @Nullable
     private final String onError;
 
-    public Config(String precondition, String directives, String udds,
+    public Config(String preconditionLanguage, String precondition, String directives, String udds,
                   String field, String schema, String onError) {
+      this.preconditionLanguage = preconditionLanguage;
       this.precondition = precondition;
       this.directives = directives;
       this.udds = udds;
+      this.preconditionSQL = precondition;
       this.field = field;
       this.schema = schema;
       this.onError = onError;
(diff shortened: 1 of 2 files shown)


wrangler-transform/widgets/Wrangler-transform.json
```diff
@@ -77,6 +104,40 @@
   ],
   "emit-alerts": true,
   "emit-errors": true,
+  "filters": [
+    {
+      "name": "PreconditionValueJEXL",
+      "condition": {
```
DJSagarAhire, line 110: Can we also clear the existing contents of the SQL precondition when JEXL is selected? As far as I can see the selection change just hides the field but does not clear it. Correct me if I'm wrong.

wrangler-transform/widgets/Wrangler-transform.json
```diff
@@ -77,6 +104,40 @@
...
+        },
+        {
+          "type": "properties",
+          "name": "precondition"
+        }
+      ]
+    },
+    {
+      "name": "preconditionValueSQL",
+      "condition": {
```
DJSagarAhire, line 122: Similar to above, can we also clear the JEXL precondition field and directives fields when the SQL precondition toggle is selected?

DJSagarAhire (commented):
One comment, LGTM otherwise

wrangler-transform/widgets/Wrangler-transform.json
```diff
@@ -18,13 +18,40 @@
...
           "widget-type": "textbox",
           "label": "Precondition",
           "name": "precondition",
           "widget-attributes" : {
             "default" : "false"
           }
+        },
+        {
+          "widget-type": "textbox",
+          "label": "Precondition",
```
DJSagarAhire: We should make it clear that this textbox will accept a SQL precondition in the label. Similar for JEXL.

wrangler-transform/src/main/java/io/cdap/wrangler/Wrangler.java
```diff
@@ -498,6 +500,16 @@ public static class Config extends PluginConfig {
...
 
+    @Name(NAME_PRECONDITION_SQL)
+    @Description("Precondition expression specifying filtering before applying directives (true to filter)")
+    @Macro
+    private String preconditionSql;
+
+    @Name(NAME_PRECONDITION_TOGGLE)
+    @Description("Precondition toggle to toggle between JEXL and SQL")
+    @Macro
+    private String preconditionToggle;
```
fernst: This field should be called something like "Precondition Format" or "Precondition Language".

wrangler-transform/src/main/java/io/cdap/wrangler/Wrangler.java
```diff
@@ -498,6 +500,16 @@ public static class Config extends PluginConfig {
     @Macro
     private String precondition;
 
+    @Name(NAME_PRECONDITION_SQL)
+    @Description("Precondition expression specifying filtering before applying directives (true to filter)")
+    @Macro
+    private String preconditionSql;
+
+    @Name(NAME_PRECONDITION_TOGGLE)
+    @Description("Precondition toggle to toggle between JEXL and SQL")
```
fernst: "Toggle to configure precondition language between JEXL and SQL"

wrangler-transform/widgets/Wrangler-transform.json
fernst: I would simply name this "Precondition (SQL)".

wrangler-transform/src/main/java/io/cdap/wrangler/Wrangler.java
```diff
@@ -498,6 +500,16 @@ public static class Config extends PluginConfig {
...
+    @Macro
+    private String preconditionToggle;
+
     @Name(NAME_DIRECTIVES)
```
fernst, line 599: We also need to update the validation rules to account for the missing Precondition field, and ensure all fields are set correctly.

fernst (changes requested):
Left a few comments and questions

wrangler-transform/widgets/Wrangler-transform.json
itsankit-google: I think this was done by UI because we wanted to retain user input on connections toggle: https://github.com/cdapio/cdap-ui/pull/467.

So it would be good to check with UI team if there is a way to clear existing contents.
As of now, I don't think it is possible.

DJSagarAhire: If it's not possible I guess there's not much we can do about it 🤷🏻

itsankit-google: Is there any specific reason to clear these fields? It makes sense to retain them according to me when we think in terms of user experience, if a user accidentally clicks on the `SQL` radio button after adding filling up the fields related to `JEXL` they would not want to fill them up again after switching back to `JEXL`.

DJSagarAhire: The input in the other field will never be used so it would be nice to clear it so that if it is accidentally referenced somewhere in code it immediately throws an exception. We will add validation based on the toggle so clearing the field isn't strictly required, though.

itsankit-google: I think we should get some product input here.

wrangler-transform/widgets/Wrangler-transform.json
```diff
@@ -77,6 +104,40 @@
   ],
   "emit-alerts": true,
   "emit-errors": true,
+  "filters": [
+    {
+      "name": "PreconditionValueJEXL",
+      "condition": {
+        "expression": "preconditionToggle == 'jexl'"
```
itsankit-google: Have we verified an upgraded pipeline in this case? How will it render the UI when the pipeline is upgraded which didn't have any value for toggle??

wrangler-transform/src/main/java/io/cdap/wrangler/Wrangler.java
akritig187: Since, both the textboxes will be non-nullable, it will give validation issue. 
image

itsankit-google: You can check the value of `precondition` based on the value of toggle instead of setting it as `required`.

akritig187: So, in that case can't we use just one single textbox, instead of two?

fernst: Simple answer is that a Jexl expression and a SQL Expression are incompatible, so it makes sense that these 2 settings are completely independent from each other.

wrangler-transform/src/main/java/io/cdap/wrangler/Wrangler.java
fernst: Can you rename the field to `preconditionFormat`? I don't think we use the word toggle anywhere else, and this is effectively a selector.

wrangler-transform/src/main/java/io/cdap/wrangler/Wrangler.java
```diff
@@ -498,6 +500,16 @@ public static class Config extends PluginConfig {
     @Macro
     private String precondition;
 
+    @Name(NAME_PRECONDITION_SQL)
+    @Description("Precondition expression specifying filtering before applying directives (true to filter)")
+    @Macro
+    private String preconditionSql;
```
fernst: Please update validation rules for these fields.

Precondition SQL should be not-empty when SQL mode is selected.

Precondition in Jexl should be not-empty as well when the jexl mode is selected.

fernst:
I'm wondering if it makes more sense to keep the directives visible and fail validation (show an error) if SQL mode is selected.

This way, it's obvious to the user that the directives won't execute.

Also, in SQL mode, the output schema is exactly the same as the input schema, and we should set this accordingly when validating.

@DJSagarAhire @akritig187 @itsankit-google WDYT?

fernst (changes requested):

wrangler-transform/widgets/Wrangler-transform.json
akritig187: Yeah I checked this and the toggle is taking the default `jexl` value and is working fine.

wrangler-transform/src/main/java/io/cdap/wrangler/Wrangler.java
```diff
@@ -219,12 +227,13 @@ public void configurePipeline(PipelineConfigurer configurer) {
       }
 
       // Check if pre-condition is not null or empty and if so compile expression.
-      if (!config.containsMacro(Config.NAME_PRECONDITION)) {
-        if (config.precondition != null && !config.precondition.trim().isEmpty()) {
+      if (!config.containsMacro(Config.NAME_PRECONDITION_JEXL)) {
+        if (config.preconditionLanguage == "jexl" && config.preconditionJexl != null
```
DJSagarAhire: Since this is a string it should use `.equals()` and not `==`. Also, can you please move this long boolean condition into a function so that it is more readable here? And also can you add a similar function for checking if the SQL input is valid?

wrangler-transform/src/main/java/io/cdap/wrangler/Wrangler.java
```diff
@@ -459,6 +469,26 @@ private void validateInputSchema(@Nullable Schema inputSchema, FailureCollector
     }
   }
 
+  private void validatePrecondition(String precondition, String field, FailureCollector collector) {
+    if (precondition == null) {
+      collector.addFailure(String.format("Precondition %s must be present.", field.substring(12)).toUpperCase(),
```
DJSagarAhire: This `substring` makes the code very brittle. I suggest simply using `%s must be present` and putting `field` directly instead.

wrangler-transform/src/main/java/io/cdap/wrangler/Wrangler.java
```diff
@@ -459,6 +469,26 @@ private void validateInputSchema(@Nullable Schema inputSchema, FailureCollector
...
+                           null)
+        .withConfigProperty(field);
+    }
+  }
+
+  private void validateDirectivesAndUDDs(FailureCollector collector) {
```
DJSagarAhire: UDDs are also technically directives so I think we can name this `validateDirectives()` to simplify it.

wrangler-transform/src/main/java/io/cdap/wrangler/Wrangler.java
```diff
@@ -486,17 +516,32 @@ private RecipeParser getRecipeParser(StageContext context)
...
     static final String NAME_SCHEMA = "schema";
     static final String NAME_ON_ERROR = "on-error";
 
-    @Name(NAME_PRECONDITION)
+    @Name(NAME_PRECONDITION_LANGUAGE)
+    @Description("Toggle to configure precondition language between JEXL and SQL")
+    private String preconditionLanguage;
+
+    @Name(NAME_PRECONDITION_JEXL)
+    @Description("Precondition expression specifying filtering before applying directives (true to filter)")
```
DJSagarAhire: This description should mention JEXL.

wrangler-transform/src/main/java/io/cdap/wrangler/Wrangler.java
```diff
@@ -486,17 +516,32 @@ private RecipeParser getRecipeParser(StageContext context)
...
+    @Macro
+    @Nullable
+    private String preconditionJexl;
+
+
+    @Name(NAME_PRECONDITION_SQL)
     @Description("Precondition expression specifying filtering before applying directives (true to filter)")
```
DJSagarAhire: This description should mention SQL.

DJSagarAhire (commented):
Since the validation logic has now become slightly involved, I think a couple of unit tests would help a lot here.

wrangler-transform/src/main/java/io/cdap/wrangler/Wrangler.java
```diff
@@ -486,17 +516,32 @@ private RecipeParser getRecipeParser(StageContext context)
...
-    @Name(NAME_PRECONDITION)
+    @Name(NAME_PRECONDITION_LANGUAGE)
+    @Description("Toggle to configure precondition language between JEXL and SQL")
+    private String preconditionLanguage;
+
+    @Name(NAME_PRECONDITION_JEXL)
+    @Description("Precondition expression specifying filtering before applying directives (true to filter)")
+    @Macro
+    @Nullable
+    private String preconditionJexl;
```
fernst: This is a backwards incompatible change. We should never change the name of existing properties.

This property must continue to be called `precondition`.

wrangler-transform/src/main/java/io/cdap/wrangler/Wrangler.java
```diff
@@ -157,6 +157,15 @@ public void configurePipeline(PipelineConfigurer configurer) {
         }
       }
 
+      String language = config.preconditionLanguage;
+      if (language.equals("jexl")) {
```
fernst: I don't think this is compatible with pipeline upgrades. 

We should allow the precondition language to be null, and if null, we allow jexl.

Many times, upgrades are done via scripts and not through the UI. We cannot assume this property will be set.

fernst (changes requested):

wrangler-transform/src/main/java/io/cdap/wrangler/Wrangler.java
itsankit-google: +1.

wrangler-transform/widgets/Wrangler-transform.json
```diff
@@ -77,6 +104,32 @@
   ],
   "emit-alerts": true,
   "emit-errors": true,
+  "filters": [
+    {
+      "name": "PreconditionValueJEXL",
+      "condition": {
+        "expression": "preconditionLanguage == 'jexl'"
```
itsankit-google: Re-iterating what @fernst mentioned earlier, when upgrades are done via scripts and not through the UI. We cannot assume this property will be set since it will not be present in older pipelines. When `null` we should assume it to be `jexl`.

wrangler-transform/src/main/java/io/cdap/wrangler/Wrangler.java
```diff
@@ -459,6 +486,49 @@ private void validateInputSchema(@Nullable Schema inputSchema, FailureCollector
...
+
+    if (config.udds != null) {
+      collector.addFailure("UDDs are not supported for precondition of type SQL", null)
+        .withConfigProperty(Config.NAME_UDD);
+    }
+  }
+
+  private boolean checkPreconditionNotEmpty(Boolean isConditionSQL) {
+    if (isConditionSQL != false) {
+      if (!config.preconditionLanguage.equals("sql") && config.precondition != null
```
itsankit-google: nit: generally we should keep constants first to avoid NPE i.e. `"sql".equals(config.preconditionLanguage))` and replace other two checks with `Strings.isNullOrEmpty(config.precondition)`.

wrangler-transform/src/main/java/io/cdap/wrangler/Wrangler.java
```diff
@@ -459,6 +486,49 @@ private void validateInputSchema(@Nullable Schema inputSchema, FailureCollector
...
+        && !config.precondition.trim().isEmpty()) {
+        return true;
+      }
+    } else {
+      if (config.preconditionLanguage.equals("sql") && config.preconditionSQL != null
+        && !config.preconditionSQL.trim().isEmpty()) {
```
itsankit-google: similar comment here.

wrangler-transform/src/main/java/io/cdap/wrangler/Wrangler.java
```diff
@@ -157,6 +157,14 @@ public void configurePipeline(PipelineConfigurer configurer) {
         }
       }
 
+      String language = config.preconditionLanguage;
+      if (language.equals("sql")) {
```
itsankit-google: similarly here: `"sql".equals(language)`

itsankit-google: Also if `language` is not used anywhere else replace it by `"sql".equals(config.preconditionLanguage)`.

itsankit-google (commented):
I feel it would be good to add a getter for `preconditionLanguage` instead of accessing it directly.

```
String getPreconditionLanguage() {
    if (Strings.isNullOrEmpty(preconditionLanguage)) {
    // due to backward compatibility...
        return "jexl";
    } 
    return preconditionLanguage.
}
```
And replace `config.preconditionLanguage` with `config.getPreconditionLanguage()`.
This way we ensure we never encounter NPE.

wrangler-transform/src/main/java/io/cdap/wrangler/Wrangler.java
```diff
@@ -61,6 +61,7 @@
 import io.cdap.wrangler.registry.SystemDirectiveRegistry;
 import io.cdap.wrangler.registry.UserDirectiveRegistry;
 import io.cdap.wrangler.utils.StructuredToRowTransformer;
+import jdk.internal.joptsimple.internal.Strings;
```
itsankit-google: use `com.google.common.base.Strings`

wrangler-transform/src/main/java/io/cdap/wrangler/Wrangler.java
```diff
@@ -459,6 +486,49 @@ private void validateInputSchema(@Nullable Schema inputSchema, FailureCollector
...
+    if (!Strings.isNullOrEmpty(config.udds)) {
+      collector.addFailure("UDDs are not supported for precondition of type SQL", null)
+        .withConfigProperty(Config.NAME_UDD);
+    }
+  }
+
+  private boolean checkPreconditionNotEmpty(Boolean isConditionSQL) {
+    if (isConditionSQL != false) {
+      if (!"sql".equals(config.getPreconditionLanguage()) && !Strings.isNullOrEmpty(config.precondition)
+        && !config.precondition.trim().isEmpty()) {
```
itsankit-google, line 519: can remove: `config.precondition.trim().isEmpty()`

wrangler-transform/src/main/java/io/cdap/wrangler/Wrangler.java
```diff
@@ -459,6 +486,49 @@ private void validateInputSchema(@Nullable Schema inputSchema, FailureCollector
...
+      }
+    } else {
+      if ("sql".equals(config.getPreconditionLanguage()) && !Strings.isNullOrEmpty(config.preconditionSQL)
+        && !config.preconditionSQL.trim().isEmpty()) {
+        return true;
+      }
+    }
+
+    return false;
+  }
```
itsankit-google: Generally we should avoid nested `if-else` and the method should only check what the name suggests:
```
private boolean checkPreconditionNotEmpty(Boolean isConditionSQL) {
  if (isConditionSQL && !Strings.isNullOrEmpty(config.precondition)
        && !config.precondition.trim().isEmpty()) {
      return true;
  }
  if (!isConditionSQL && !Strings.isNullOrEmpty(config.preconditionSQL)
        && !config.preconditionSQL.trim().isEmpty()) {
      return true;
  }
  return false;
}
```

wrangler-transform/src/main/java/io/cdap/wrangler/Wrangler.java
```diff
@@ -309,15 +327,24 @@ public void initialize(TransformContext context) throws Exception {
       );
     }
 
-    // Check if pre-condition is not null or empty and if so compile expression.
-    if (config.precondition != null && !config.precondition.trim().isEmpty()) {
+    // Check if jexl pre-condition is not null or empty and if so compile expression.
+    if (checkPreconditionNotEmpty(false)) {
```
itsankit-google:
```
if ("jexl".equals(config.getPreconditionLanguage()) && checkPreconditionNotEmpty(false)) {
.....
}
```

wrangler-transform/src/main/java/io/cdap/wrangler/Wrangler.java
```diff
@@ -309,15 +327,24 @@ public void initialize(TransformContext context) throws Exception {
...
       try {
         condition = new Precondition(config.precondition);
       } catch (PreconditionException e) {
         throw new IllegalArgumentException(e.getMessage(), e);
       }
     }
 
+    // Check if sql pre-condition is not null or empty and if so compile expression.
+    if (checkPreconditionNotEmpty(true)) {
```
itsankit-google:
```
if ("sql".equals(config.getPreconditionLanguage()) && checkPreconditionNotEmpty(true)) {
.....
}
```

wrangler-transform/src/main/java/io/cdap/wrangler/Wrangler.java
```diff
@@ -157,6 +158,13 @@ public void configurePipeline(PipelineConfigurer configurer) {
         }
       }
 
+      if ("sql".equals(config.getPreconditionLanguage())) {
+        validatePrecondition(config.preconditionSQL, true, collector);
```
fernst: use getters here instead of referencing the underlying variable directly.

wrangler-transform/src/main/java/io/cdap/wrangler/Wrangler.java
```diff
@@ -157,6 +158,13 @@ public void configurePipeline(PipelineConfigurer configurer) {
...
+        validateDirectives(collector);
+      } else {
+        validatePrecondition(config.precondition, false, collector);
```
fernst: use getters here instead of referencing the underlying variable directly.

wrangler-transform/src/main/java/io/cdap/wrangler/Wrangler.java
```diff
@@ -218,9 +225,9 @@ public void configurePipeline(PipelineConfigurer configurer) {
           .withConfigProperty(Config.NAME_SCHEMA).withStacktrace(e.getStackTrace());
       }
 
-      // Check if pre-condition is not null or empty and if so compile expression.
+      // Check if jexl pre-condition is not null or empty and if so compile expression.
       if (!config.containsMacro(Config.NAME_PRECONDITION)) {
-        if (config.precondition != null && !config.precondition.trim().isEmpty()) {
+        if ("jexl".equals(config.getPreconditionLanguage()) && checkPreconditionNotEmpty(false)) {
```
fernst: this can be be `equalsIgnoreCase` for added safety

wrangler-transform/src/main/java/io/cdap/wrangler/Wrangler.java
```diff
@@ -229,6 +236,17 @@ public void configurePipeline(PipelineConfigurer configurer) {
         }
       }
 
+      // Check if sql pre-condition is not null or empty and if so compile expression.
+      if (!config.containsMacro(Config.NAME_PRECONDITION_SQL)) {
+        if ("sql".equals(config.getPreconditionLanguage()) && checkPreconditionNotEmpty(true)) {
```
fernst: this can be be equalsIgnoreCase for added safety

wrangler-transform/src/main/java/io/cdap/wrangler/Wrangler.java
```diff
@@ -487,17 +553,32 @@ private RecipeParser getRecipeParser(StageContext context)
...
+    static final String NAME_PRECONDITION_LANGUAGE = "preconditionLanguage";
     static final String NAME_FIELD = "field";
     static final String NAME_DIRECTIVES = "directives";
     static final String NAME_UDD = "udd";
     static final String NAME_SCHEMA = "schema";
     static final String NAME_ON_ERROR = "on-error";
 
+    @Name(NAME_PRECONDITION_LANGUAGE)
+    @Description("Toggle to configure precondition language between JEXL and SQL")
+    private String preconditionLanguage;
```
fernst, line 583: This configuration option should support macros as well.

wrangler-transform/src/main/java/io/cdap/wrangler/Wrangler.java
```diff
@@ -459,6 +486,45 @@ private void validateInputSchema(@Nullable Schema inputSchema, FailureCollector
...
+      language = "Precondition (SQL)";
+    }
+    if (Strings.isNullOrEmpty(precondition)) {
+      collector.addFailure(String.format("%s must be present.", language),
+                           null)
+        .withConfigProperty(field);
+    }
+  }
+
+  private void validateDirectives(FailureCollector collector) {
```
fernst: Since this is only invoked when SQL is selected, maybe this method should be called something like `validateSQLModeDirectives`

fernst: Code looks great in this function.

fernst (changes requested):

wrangler-transform/src/main/java/io/cdap/wrangler/Wrangler.java
tivv: Let's also use at least constants for list of languages, if not an enum. It helps a lot to reduce number of spelling mistakes

wrangler-transform/src/main/java/io/cdap/wrangler/Wrangler.java
fernst: This hasn't been resolved.

wrangler-transform/src/main/java/io/cdap/wrangler/Wrangler.java
fernst: Good point too

wrangler-transform/src/main/java/io/cdap/wrangler/Wrangler.java
fernst: This has not been resolved.

wrangler-transform/src/main/java/io/cdap/wrangler/Wrangler.java
fernst: This has not been resolved.

wrangler-transform/src/main/java/io/cdap/wrangler/Wrangler.java
```diff
@@ -144,19 +149,27 @@ public void configurePipeline(PipelineConfigurer configurer) {
...
         if (config.containsMacro("directives")) {
-          directives = String.format("#pragma load-directives %s;", config.udds);
+          directives = String.format("#pragma load-directives %s;", config.getUDDs());
         } else {
-          directives = String.format("#pragma load-directives %s;%s", config.udds, config.directives);
+          directives = String.format("#pragma load-directives %s;%s", config.getUDDs(), config.getDirectives());
         }
       }
 
+      if (PRECONDITION_LANGUAGE_SQL.equalsIgnoreCase(config.getPreconditionLanguage())) {
```
itsankit-google: since the field can be macro, we should first check:
`if (!config.containsMacro(Config.NAME_PRECONDITION_LANGUAGE)`

fernst (approved):

itsankit-google (approved):
LGTM, please squash the commits before merge.

PR #616, opened 2023-03-23
[CDAP-20454] Add implementation of LinearRelationalTransform to Wrangler instead of Precondition

DJSagarAhire:
The `Precondition` class cannot implement BQ Pushdown by implementing the `LinearRelationalTransform` interface as it is not a pipeline stage. Thus moving the BQ Pushdown implementation to the `Wrangler` class instead.


fernst (approved):
LGTM

PR #624, opened 2023-04-03
[CDAP-20566] Add directive metrics 

vanathi-g:
The following new metrics are collected per pipeline run to help improve Wrangler directives:

- Number of times a directive is used
- Number of times a JEXL function category is used
- Number of times a hash algorithm is used

## Changes
- Add utility classes and a new interface to simplify emitting metrics from Wrangler directives
- Add functions to emit metrics from the prepareRun method in Wrangler plugin class

Note: same as PR #614 but with cdap-common dependency issue removed [CDAP-20502]

[CDAP-20502]: https://cdap.atlassian.net/browse/CDAP-20502?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/wrangler-api/src/main/java/io/cdap/wrangler/api/EntityCountMetricDef.java
+++ b/wrangler-api/src/main/java/io/cdap/wrangler/api/EntityCountMetricDef.java
@@ -1,0 +1,63 @@
+/*
+ * Copyright © 2023 Cask Data, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package io.cdap.wrangler.api;
+
+/**
+ * Represents generic metadata information for a count metric that is emitted in Wrangler. The entity's type and name
+ * will be automatically mapped to corresponding metric tags.
+ */
+public class EntityCountMetricDef {
+  /**
+   * Metric name
+   */
+  private final String name;
+  /**
+   * Value by which to increment the count
+   */
+  private final long count;
+  /**
+   * System app entity type
+   */
+  private final String appEntityType;
+  /**
+   * System app entity type name
+   */
+  private final String appEntityTypeName;
+
+  public EntityCountMetricDef(String name, String appEntityType, String appEntityTypeName, long count) {
+    this.name = name;
+    this.appEntityType = appEntityType;
+    this.appEntityTypeName = appEntityTypeName;
+    this.count = count;
+  }
+
+  public String getName() {
+    return name;
+  }
+
+  public String getAppEntityType() {
+    return appEntityType;
+  }
+
+  public String getAppEntityTypeName() {
+    return appEntityTypeName;
+  }
+
+  public long getCount() {
+    return count;
+  }
+}
--- a/wrangler-core/src/main/java/io/cdap/wrangler/metrics/Constants.java
+++ b/wrangler-core/src/main/java/io/cdap/wrangler/metrics/Constants.java
@@ -1,0 +1,34 @@
+/*
+ * Copyright © 2023 Cask Data, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package io.cdap.wrangler.metrics;
+
+/**
+ * Constants for emitting CDAP metrics from Wrangler
+ */
+public class Constants {
+  /**
+   * Metric tags (same as those defined in CDAP)
+   */
+  public static final class Tags {
+    public static final String APP_ENTITY_TYPE = "aet";
+    public static final String APP_ENTITY_TYPE_NAME = "tpe";
+  }
+
+  private Constants() {
+    throw new AssertionError("Cannot instantiate a static utility class.");
+  }
+}
--- a/wrangler-transform/src/main/java/io/cdap/wrangler/Wrangler.java
+++ b/wrangler-transform/src/main/java/io/cdap/wrangler/Wrangler.java
@@ -595,20 +595,23 @@ public class Wrangler extends Transform<StructuredRecord, StructuredRecord> impl
    */
   private void emitDirectiveMetrics(List<Directive> directives, Metrics metrics) throws DirectiveLoadException {
     for (Directive directive : directives) {
-      List<EntityCountMetric> metricDefs = new ArrayList<>();
-
-      // Emit directive usage metric only if directive is present in system directive registry
+      // skip emitting metrics if the directive is not system metric
       if (registry.get(Contexts.SYSTEM, directive.define().getDirectiveName()) != null) {
-        metricDefs.add(getDirectiveUsageMetric(directive.define().getDirectiveName()));
+        continue;
       }
+      List<EntityCountMetric> countMetrics = new ArrayList<>();
 
+      // add usage metric
+      countMetrics.add(getDirectiveUsageMetric(directive.define().getDirectiveName()));
+
+      // add custom directive metrics
       if (directive.getCountMetrics() != null) {
-        metricDefs.addAll(directive.getCountMetrics());
+        countMetrics.addAll(directive.getCountMetrics());
       }
 
-      for (EntityCountMetric metricDef : metricDefs) {
-        Metrics child = metrics.child(getEntityMetricTags(metricDef));
-        child.countLong(metricDef.getName(), metricDef.getCount());
+      for (EntityCountMetric countMetric : countMetrics) {
+        Metrics child = metrics.child(getEntityMetricTags(countMetric));
+        child.countLong(countMetric.getName(), countMetric.getCount());
       }
     }
   }
--- a/wrangler-transform/src/main/java/io/cdap/wrangler/Wrangler.java
+++ b/wrangler-transform/src/main/java/io/cdap/wrangler/Wrangler.java
@@ -49,7 +49,7 @@ import io.cdap.wrangler.api.Compiler;
 import io.cdap.wrangler.api.Directive;
 import io.cdap.wrangler.api.DirectiveLoadException;
 import io.cdap.wrangler.api.DirectiveParseException;
-import io.cdap.wrangler.api.EntityCountMetricDef;
+import io.cdap.wrangler.api.EntityCountMetric;
 import io.cdap.wrangler.api.ErrorRecord;
 import io.cdap.wrangler.api.ExecutorContext;
 import io.cdap.wrangler.api.RecipeParser;
@@ -595,7 +595,7 @@ public class Wrangler extends Transform<StructuredRecord, StructuredRecord> impl
    */
   private void emitDirectiveMetrics(List<Directive> directives, Metrics metrics) throws DirectiveLoadException {
     for (Directive directive : directives) {
-      List<EntityCountMetricDef> metricDefs = new ArrayList<>();
+      List<EntityCountMetric> metricDefs = new ArrayList<>();
 
       // Emit directive usage metric only if directive is present in system directive registry
       if (registry.get(Contexts.SYSTEM, directive.define().getDirectiveName()) != null) {
@@ -606,19 +606,19 @@ public class Wrangler extends Transform<StructuredRecord, StructuredRecord> impl
         metricDefs.addAll(directive.getCountMetrics());
       }
 
-      for (EntityCountMetricDef metricDef : metricDefs) {
+      for (EntityCountMetric metricDef : metricDefs) {
         Metrics child = metrics.child(getEntityMetricTags(metricDef));
         child.countLong(metricDef.getName(), metricDef.getCount());
       }
     }
   }
 
-  private EntityCountMetricDef getDirectiveUsageMetric(String directiveName) {
-    return new EntityCountMetricDef(
+  private EntityCountMetric getDirectiveUsageMetric(String directiveName) {
+    return new EntityCountMetric(
       DIRECTIVE_METRIC_NAME, DIRECTIVE_ENTITY_TYPE, directiveName, DIRECTIVE_METRIC_COUNT);
   }
 
-  private Map<String, String> getEntityMetricTags(EntityCountMetricDef metricDef) {
+  private Map<String, String> getEntityMetricTags(EntityCountMetric metricDef) {
     Map<String, String> tags = new HashMap<>();
     tags.put(APP_ENTITY_TYPE, metricDef.getAppEntityType());
     tags.put(APP_ENTITY_TYPE_NAME, metricDef.getAppEntityTypeName());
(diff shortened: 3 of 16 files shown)


wrangler-api/pom.xml
```diff
@@ -39,6 +39,11 @@
       <version>${cdap.version}</version>
       <scope>provided</scope>
     </dependency>
+    <dependency>
+      <groupId>io.cdap.cdap</groupId>
+      <artifactId>cdap-watchdog-api</artifactId>
```
albertshau: Which classes from this are being used? This should be a CDAP internal module. Plugins should normally only need cdap-api and cdap-etl-api

wrangler-core/src/main/java/io/cdap/wrangler/metrics/DirectiveJEXLCategoryMetric.java
```diff
@@ -0,0 +1,60 @@
...
+   * present in {@link io.cdap.wrangler.expression.EL.DefaultFunctions}
+   *
+   * @param jexlScript the JEXL script from which the JEXL function category will be parsed
+   * @return {@link EntityMetricDef} with the metric name and necessary tags representing a JEXL category metric
+   */
+  default EntityMetricDef getJEXLCategoryMetric(String jexlScript) {
+    String jexlCategory = jexlScript.split(":")[0].trim();
+    if (EL_DEFAULT_FUNCTIONS.containsKey(jexlCategory)) {
+      return new EntityMetricDef(
+        JEXL_CATEGORY_METRIC_NAME, ImmutableMap.of(SERVICE, APPLICATION_NAME,
```
albertshau: the service tag is for service program types. This would be the DataPrepService (https://github.com/data-integrations/wrangler/blob/develop/wrangler-service/src/main/java/io/cdap/wrangler/service/DataPrepService.java) that powers the real-time wrangler UI.

The service tag is automatically set by the platform when getting the Metrics interface within a service program, just like how the namespace and application tags are automatically set. In a pipeline, it will also automatically have the workflow tag set. The plugin should only be adding these entity type and name tags, it should not be doing anything with the program type (service, workflow, etc) tags.

wrangler-core/src/main/java/io/cdap/wrangler/metrics/Constants.java
```diff
@@ -0,0 +1,37 @@
...
+
+  /**
+   * Metric tags (same as those defined in CDAP)
+   */
+  public static final class Tags {
+    public static final String APP_ENTITY_TYPE = "aet";
+    public static final String APP_ENTITY_TYPE_NAME = "tpe";
+    public static final String SERVICE = "srv";
+
+  }
```
albertshau, line 29: nit: add a newline before constructor

wrangler-core/src/main/java/io/cdap/wrangler/metrics/Constants.java
```diff
@@ -0,0 +1,37 @@
...
+ * the License.
+ */
+
+package io.cdap.wrangler.metrics;
+
+/**
+ * Constants for emitting CDAP metrics from Wrangler
+ */
+public class Constants {
+  public static final String APPLICATION_NAME = "dataprep";
```
albertshau: this shouldn't be used, the app name is something only the platform knows, not the plugin.

wrangler-transform/src/main/java/io/cdap/wrangler/Wrangler.java
```diff
@@ -573,6 +586,49 @@ private Optional<ExpressionFactory<String>> getExpressionFactory(RelationalTranf
...
+  private void emitDirectiveMetrics(List<Directive> directives, Metrics metrics) throws DirectiveLoadException {
+    for (Directive directive : directives) {
+      List<EntityMetricDef> entityMetricDefs = new ArrayList<>();
+
+      // Emit directive usage metric only if directive is present in system directive registry
+      if (registry.get(Contexts.SYSTEM, directive.define().getDirectiveName()) != null) {
+        entityMetricDefs.add(getDirectiveUsageMetric(directive.define().getDirectiveName()));
+      }
+
+      if (directive instanceof EntityMetrics) {
```
albertshau: instanceof is a code smell and should be avoided when possible.

In this particular case, the Directive interface can be modified to have the method that is in EntityMetrics, with a default implementation that returns nothing.

wrangler-core/src/main/java/io/cdap/wrangler/metrics/DirectiveJEXLCategoryMetric.java
```diff
@@ -0,0 +1,60 @@
...
+import static io.cdap.wrangler.metrics.Constants.Tags.APP_ENTITY_TYPE;
+import static io.cdap.wrangler.metrics.Constants.Tags.APP_ENTITY_TYPE_NAME;
+import static io.cdap.wrangler.metrics.Constants.Tags.SERVICE;
+
+/**
+ * Represents a directive that emits the 'JEXL category' metric. This metric counts the number of times
+ * a JEXL function category in {@link io.cdap.wrangler.expression.EL.DefaultFunctions} is used.
+ */
+public interface DirectiveJEXLCategoryMetric extends EntityMetrics {
+  String JEXL_CATEGORY_METRIC_NAME = "wrangler.jexl-category.count";
```
albertshau: private final for all of these variables

wrangler-core/src/main/java/io/cdap/wrangler/metrics/DirectiveJEXLCategoryMetric.java
```diff
@@ -0,0 +1,60 @@
...
+
+  /**
+   * This method parses the JEXL function category from the given JEXL script and returns the metric if the category is
+   * present in {@link io.cdap.wrangler.expression.EL.DefaultFunctions}
+   *
+   * @param jexlScript the JEXL script from which the JEXL function category will be parsed
+   * @return {@link EntityMetricDef} with the metric name and necessary tags representing a JEXL category metric
+   */
+  default EntityMetricDef getJEXLCategoryMetric(String jexlScript) {
+    String jexlCategory = jexlScript.split(":")[0].trim();
```
albertshau: this is a pretty unreliable way to get this information. There isn't any way to get an accurate count while Jexl is doing the parsing?

wrangler-transform/src/main/java/io/cdap/wrangler/Wrangler.java
```diff
@@ -573,6 +586,49 @@ private Optional<ExpressionFactory<String>> getExpressionFactory(RelationalTranf
...
+        entityMetricDefs.addAll(((EntityMetrics) directive).getMetrics());
+      }
+
+      for (EntityMetricDef entityMetricDef : entityMetricDefs) {
+        Metrics child = metrics.child(entityMetricDef.getTags());
+        switch (entityMetricDef.getType()) {
+          case COUNTER:
+            child.countLong(entityMetricDef.getName(), entityMetricDef.getValue());
+            break;
+          case GAUGE:
```
albertshau: I can't really think of a situation where a directive would want to emit a gauge. Do we have a use case? If not it's better not to have code for it.

wrangler-transform/src/main/java/io/cdap/wrangler/Wrangler.java
```diff
@@ -573,6 +586,49 @@ private Optional<ExpressionFactory<String>> getExpressionFactory(RelationalTranf
...
+   *
+   * @param directives a list of Wrangler directives
+   * @param metrics CDAP {@link Metrics} object using which metrics can be emitted
+   * @throws DirectiveLoadException
+   */
+  private void emitDirectiveMetrics(List<Directive> directives, Metrics metrics) throws DirectiveLoadException {
+    for (Directive directive : directives) {
+      List<EntityMetricDef> entityMetricDefs = new ArrayList<>();
+
+      // Emit directive usage metric only if directive is present in system directive registry
```
albertshau: why only system ones?

wrangler-transform/src/main/java/io/cdap/wrangler/Wrangler.java
```diff
@@ -573,6 +586,49 @@ private Optional<ExpressionFactory<String>> getExpressionFactory(RelationalTranf
...
+            break;
+          // TODO: add logic for distribution metrics (currently not supported)
+        }
+      }
+    }
+  }
+
+  private EntityMetricDef getDirectiveUsageMetric(String directiveName) {
+    return new EntityMetricDef(
+      DIRECTIVE_METRIC_NAME, ImmutableMap.of(SERVICE, APPLICATION_NAME,
```
albertshau: the child should only have the entity type and entity type name. The `Metrics` instance will contain SERVICE if it's being used within a service program.

This logic should also be moved to some more common place which gets called by both the Transform (run within a pipeline) and by the Service. Then we'll be able to differentiate between directive usage at development time (service) vs execution time (pipeline).

wrangler-api/src/main/java/io/cdap/wrangler/api/EntityMetricDef.java
```diff
@@ -0,0 +1,54 @@
...
+import io.cdap.cdap.api.metrics.MetricType;
+
+import java.util.Map;
+
+/**
+ * Represents generic metadata information for a metric that is emitted in Wrangler
+ */
+public class EntityMetricDef {
+  private final String name;
+  private final Map<String, String> tags;
```
albertshau: It's better to have two String variables for entityType and entityName. A Map is less clear and more error prone, as it looks like it can contain anything whereas in reality it must contain both the entity type and name.

wrangler-transform/src/main/java/io/cdap/wrangler/Wrangler.java
albertshau: note that this requires modifying the aggregations in the CDAP level, unless we're only trying to capture the service metrics.

albertshau (commented):
Please also have @yaojiefeng take a look, he will be more familiar with metrics than me.

wrangler-transform/src/main/java/io/cdap/wrangler/Wrangler.java
vanathi-g: EntityMetrics is a common interface that can be implemented by any "entity" in wrangler that would like to emit metrics. We are treating a Directive as an entity here. Hence why the function shouldn't be moved specifically to the Directive interface.

wrangler-core/src/main/java/io/cdap/wrangler/metrics/Constants.java
vanathi-g: It's already defined in the wrangler-transform package that has the Wrangler plugin class as well though: https://github.com/data-integrations/wrangler/blob/develop/wrangler-transform/src/main/java/io/cdap/wrangler/Wrangler.java#L101

wrangler-transform/src/main/java/io/cdap/wrangler/Wrangler.java
vanathi-g: For now we want to emit usage metrics only for system directives and not for UDDs

wrangler-api/src/main/java/io/cdap/wrangler/api/EntityMetricDef.java
vanathi-g: EntityMetrics is a generic class so it's possible some other entity will add additional tags as well.

wrangler-api/pom.xml
vanathi-g: Only MetricType is being used

wrangler-transform/src/main/java/io/cdap/wrangler/Wrangler.java
vanathi-g: We're trying to capture only at execution time i.e. per pipeline run. In the case of modifying CDAP aggregations, is it alright to add the APP_ENTITY_TYPE and APP_ENTITY_TYPE_NAME tags to this aggregation? [workflow aggregation](https://github.com/cdapio/cdap/blob/develop/cdap-watchdog/src/main/java/io/cdap/cdap/metrics/store/DefaultMetricStore.java#L160)

wrangler-api/pom.xml
vanathi-g: Removed dependency on cdap-watchdog-api and assumed only count metrics are used for now

wrangler-core/src/main/java/io/cdap/wrangler/metrics/DirectiveJEXLCategoryMetric.java
vanathi-g: Understood, we'll need to change the existing workflow aggregation in CDAP to use these tags during pipeline runs.

wrangler-core/src/main/java/io/cdap/wrangler/metrics/DirectiveJEXLCategoryMetric.java
vanathi-g: Used ParserTokenManager to get the first token

wrangler-core/src/main/java/io/cdap/wrangler/metrics/Constants.java
albertshau: that's the application name for the service (also it looks unused in that class?). Application name for a pipeline is the pipeline name, which is not a constant.

wrangler-api/src/main/java/io/cdap/wrangler/api/EntityMetricDef.java
albertshau: Do we have actual use cases for more generic tags? I'm not sure how that could even work, as they have to match the aggregation specified in CDAP.

If entity type and name are always required (which seems to be the case), it is better to make them explicit variables and not keys in a Map.

Especially in API classes like this one, we need to be careful about what methods are defined, as we cannot remove a method or change the signature, as that would break backwards compatibility. So we should not be putting things in for possible future use cases that we don't know about. Adding methods is backwards compatible, so it's better to add things when we actually will use them.

wrangler-transform/src/main/java/io/cdap/wrangler/Wrangler.java
albertshau: why don't we want UDDs though? It seems like an arbitrary limit and it's actually more complex to filter them out than to include them.

It's also weird to include any EntityCountMetricDef that a UDD returns, but not count the actual UDD.

wrangler-api/src/main/java/io/cdap/wrangler/api/EntityCountMetricDef.java
```diff
@@ -0,0 +1,46 @@
...
+ */
+
+package io.cdap.wrangler.api;
+
+import java.util.Map;
+
+/**
+ * Represents generic metadata information for a count metric that is emitted in Wrangler
+ */
+public class EntityCountMetricDef {
```
albertshau: nit: let's just name it EntityCountMetric, the 'Def' doesn't really describe anything meaningful.

wrangler-api/src/main/java/io/cdap/wrangler/api/EntityMetricDef.java
vanathi-g: Lets say we later add some new tags for the service aggregation in CDAP. Now, we might want to capture some additional metrics here while using the Wrangler service. I think having a map provides more flexibility in terms of what tags can be used for a metric. 

Unless we want to restrict usage of other tags? From what I've seen, it's not possible to reassign a tag that is already set in the passed metrics object nor put a duplicate value for the tag either.

vanathi-g: Discussed with Sai and made the necessary changes. Added two fields and removed the map of tags. Adding the tags later just before emitting the metrics.

saimukkamala (approved):

albertshau (commented):
sorry, forgot to click submit review, have one minor comment and one about if we can also track UDDs

wrangler-api/src/main/java/io/cdap/wrangler/api/EntityCountMetricDef.java
saimukkamala: makes sense. Added the change

wrangler-transform/src/main/java/io/cdap/wrangler/Wrangler.java
saimukkamala: Currently we are focusing on usage metrics for system directives. We will add metrics for UDDs if required, in the next release.

Also, updated the logic to emit any metrics only if the directive is system directive.

albertshau (approved):

PR #634, opened 2023-04-19
[CDAP-20584] Reverse the condition for SQL Preconditions

DJSagarAhire:
This PR reverses the condition for the SQL precondition so that the SQL filter will now pass through the records for which the condition is true and filter out the others. The JEXL filters are unchanged.

### DIFF
--- a/wrangler-transform/src/main/java/io/cdap/wrangler/Wrangler.java
+++ b/wrangler-transform/src/main/java/io/cdap/wrangler/Wrangler.java
@@ -581,8 +581,7 @@ public class Wrangler extends Transform<StructuredRecord, StructuredRecord> impl
         return new InvalidRelation("Cannot find an Expression Factory");
       }
 
-      String negatedSQL = String.format("NOT (%s)", config.getPreconditionSQL());
-      Expression filterExpression = expressionFactory.get().compile(negatedSQL);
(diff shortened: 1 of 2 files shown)


sahusanket (approved):

wrangler-transform/src/main/java/io/cdap/wrangler/Wrangler.java
```diff
@@ -655,7 +654,7 @@ public static class Config extends PluginConfig {
 
 
     @Name(NAME_PRECONDITION_SQL)
-    @Description("SQL Precondition expression specifying filtering before applying directives (true to filter)")
+    @Description("SQL Precondition expression specifying filtering before applying directives (false to filter)")
```
tivv, line 663: Let's name it differently, file "Filter" and not "Precondition" everywhere. This would reduce the confusion.

fernst: I agree, let's rename the field to SQL Filter

DJSagarAhire: Will run this change by Deepinder tomorrow and get his input.

tivv (approved):

PR #640, opened 2023-05-22
Fix date type handling when parsing excel files

vanathi-g:
## Changes:
- Use `DataFormatter` to use set cell format when parsing date values

## Preview of changes:

Sample date formats (excel sheet):

Current behaviour:

Behaviour after fix:


wrangler-core/src/main/java/io/cdap/directives/parser/ParseExcel.java
```diff
@@ -165,8 +167,8 @@ public List<Row> execute(List<Row> records, final ExecutorContext context)
 
                   case NUMERIC:
                     if (HSSFDateUtil.isCellDateFormatted(cell)) {
-                      newRow.add(name, cell.getDateCellValue());
-                      value = cell.getDateCellValue().toString();
+                      newRow.add(name, formatter.formatCellValue(cell));
```
iam-divya: Is there any test case that need to be updated to capture this scenario?

wrangler-core/src/main/java/io/cdap/directives/parser/ParseExcel.java
```diff
@@ -165,8 +167,8 @@ public List<Row> execute(List<Row> records, final ExecutorContext context)
...
+                      value = formatter.formatCellValue(cell);
```
iam-divya, line 169: nit: adding this line after value assignment, newRow.add(name, value);

iam-divya (commented):
Left comments

wrangler-core/src/main/java/io/cdap/directives/parser/ParseExcel.java
vanathi-g: Done, added a test

wrangler-core/src/main/java/io/cdap/directives/parser/ParseExcel.java
vanathi-g: Cleaned up code

saimukkamala (approved):

PR #641, opened 2023-05-24
Relational directives

saimukkamala:


wrangler-transform/src/main/java/io/cdap/wrangler/Wrangler.java
```diff
@@ -569,23 +572,66 @@ private RecipeParser getRecipeParser(StageContext context)
...
+    }
+
+    for (Directive directive : directives) {
+      // Expression exp = expFactory.compile(sql);
+      if (!(directive instanceof RelationalDirective)) {
+        throw new RuntimeException("Directive is not relational Directive");
+      }
+      // currently supporting only drop column
+      // SQL will be returned as "DROP COLUMN col1, col2"
+      String sql = ((RelationalDirective) directive).getSQL();
```
iam-divya: Can we not invoke execute on directives and offload directive execution to RecipePipelineExecutor. We can introduce a new function for relationalDirective execute

wrangler-transform/src/main/java/io/cdap/wrangler/Wrangler.java
```diff
@@ -569,23 +572,66 @@ private RecipeParser getRecipeParser(StageContext context)
...
+      List<String> cols = getColumnsOfDropSQL(sql);
+      for (String col : cols) {
+        filteredRelation = filteredRelation.dropColumn(col);
```
iam-divya: We should move this logic into the directive. SparkSQLDataset should be returned from each Directive and passed onto the next one. SparkSQLEngine.transform should be invoked in each directive for executing. 
Building just one SQL for each Directive should be good enough. SparkSQLEngine will internally chain the SQL execute in one map operation is my understanding, please check that part. 
There is a comment on that already, https://github.com/cdapio/cdap/blob/1cc26e664a22977fb39f29ea03a46ee4ab531f92/cdap-app-templates/cdap-etl/hydrator-spark-core-base/src/main/java/io/cdap/cdap/etl/spark/batch/SparkSQLEngine.java#L193

wrangler-transform/widgets/Wrangler-transform.json
```diff
@@ -73,6 +73,24 @@
         }
       ]
     },
+    {
+      "label": "RelationalDirectives",
```
iam-divya, line 96: Do you want to keep these changes by feature flag? Otherwise we have to move this code to a branch and keep syncing the branch with Develop or other future changes.

wrangler-transform/src/main/java/io/cdap/wrangler/Wrangler.java
saimukkamala: Agreed. These changes are for poc. will move the logic to directives.

wrangler-transform/src/main/java/io/cdap/wrangler/Wrangler.java
saimukkamala: That's a good point. will look into offloading directive execution to RecipePipelineExecutor.

wrangler-transform/widgets/Wrangler-transform.json
saimukkamala: We will add feature flag. until then we won't merge to develop

PR #645, opened 2023-06-19
[CDAP-15361] Add Schema handling in Wrangler

vanathi-g:
## Background
Currently, Wrangler infers the CDAP schema of the data where necessary during design-time (UI display, creating plugin/applying directives). This leads to inference errors and information loss. Hence, a mechanism for directives to correctly update schema after transformation is implemented and input schema information is not lost.

## Changes
- Added `getOutputSchema` method to Executor interface (Directives can use this to return custom schema after transform)
- Input schema and output schema after applying a recipe are stored in `TransientStore` and accessed when necessary. 
- WorkspaceHandler's `specification` and AbstractDirectiveHandler's `generateExecutionResponse` method now use the output schema instead of inferring
- For backwards compatibility, if a directive does not override `getOutputSchema`, the schema is generated as a fallback mechanism

NOTE: the schema changes are only applicable in design time (when Wrangler runs as a service, not in the plugin)


wrangler-api/src/main/java/io/cdap/wrangler/api/Directive.java
```diff
@@ -126,4 +132,44 @@ default List<EntityCountMetric> getCountMetrics() {
...
+   */
+  @Override
+  default Schema getOutputSchema(Schema inputSchema, SchemaFieldGenerator generator, List<Row> output)
+    throws SchemaGenerationException {
+    Set<Schema.Field> outputFields = new HashSet<>(); // Use a set to avoid duplicate fields
+    Map<String, String> renamedFields = getRenamedFields(inputSchema);
+    Map<String, Schema> typeChangedFields = getTypeChangedFields(inputSchema);
+    for (Row row : output) {
+      for (Pair<String, Object> field : row.getFields()) {
+        String originalName = renamedFields.getOrDefault(field.getFirst(), field.getFirst());
```
iam-divya: Nit: just a get would do in this case since default value is also same.

wrangler-api/src/main/java/io/cdap/wrangler/api/schema/SchemaFieldGenerator.java
```diff
@@ -0,0 +1,26 @@
...
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package io.cdap.wrangler.api.schema;
+
+import io.cdap.cdap.api.data.schema.Schema;
+
+/**
+ * JAVADOC
```
iam-divya: Good to have the doc added

wrangler-core/src/main/java/io/cdap/wrangler/executor/RecipePipelineExecutor.java
```diff
@@ -112,6 +114,12 @@ public List<Row> execute(List<Row> rows) throws RecipeException {
           context.getTransientStore().reset(TransientVariableScope.LOCAL);
         }
 
+        // Get input schema from transientStore
+        // TODO check if possible to access workspace id somehow for prefix - worst case change set variable dir
```
iam-divya: nit: remove or update comment

wrangler-api/src/main/java/io/cdap/wrangler/api/Directive.java
```diff
@@ -126,4 +132,44 @@ default List<EntityCountMetric> getCountMetrics() {
...
+   *   <li>Get map of renamed fields and map of type changed fields</li>
+   *   <li>Iterate over list of fields in each output row</li>
+   *   <li>If output field is renamed, get original name.</li>
+   *   <li>Check if it is in map of type changed fields or if it exists in inputSchema, then use that schema</li>
+   *   <li>If it is a new field, generate the schema using provided {@link SchemaFieldGenerator}</li>
+   *   <li>Finally, add the created field to set of output fields</li>
+   * </ol>
+   */
+  @Override
+  default Schema getOutputSchema(Schema inputSchema, SchemaFieldGenerator generator, List<Row> output)
```
tivv: I would separate responsibilities. The directive should be able to provide schema if it knows how to do it, otherwise return null. And caller should handle null and do detection

wrangler-api/src/main/java/io/cdap/wrangler/api/schema/SchemaFieldGenerator.java
tivv: Please fix

tivv (commented):
Please refer to JIRA int the PR title

wrangler-core/src/main/java/io/cdap/wrangler/executor/RecipePipelineExecutor.java
```diff
@@ -112,6 +114,12 @@ public List<Row> execute(List<Row> rows) throws RecipeException {
...
+        Schema schema = context.getEnvironment() != null &&
+          context.getEnvironment().equals(ExecutorContext.Environment.SERVICE) ?
+          context.getTransientStore().get("ws_schema") : null;
```
iam-divya: nit: should we call it as output schema for readability here and other places including updated schema(to become updated output schema)?

iam-divya (dismissed):

wrangler-core/src/main/java/io/cdap/directives/column/Rename.java
```diff
@@ -75,6 +78,11 @@ public List<Row> execute(List<Row> rows, ExecutorContext context) throws Directi
     return rows;
   }
 
+  @Override
+  public Map<String, String> getRenamedFields(Schema inputSchema) {
```
iam-divya: I see that you have made changes in rename and settype directives, is there any other directive which would need change to accommodate? 

How are we handling parsing directives?

wrangler-api/src/main/java/io/cdap/wrangler/api/Directive.java
vanathi-g: No, we need it as the get would return null if the field name is not present in the map.

wrangler-core/src/main/java/io/cdap/directives/column/Rename.java
vanathi-g: I will revert the changes to those directives, added them here just for testing purposes. Next week I'm planning to raise a PR with changes to all the renaming directives, then the week after that changes to data type changing directives. I already have a list that need to be changed, which includes some of the parsing directives that require specific changes.

wrangler-api/src/main/java/io/cdap/wrangler/api/Directive.java
vanathi-g: Done, moved the schema generation handling logic to RecipePipelineExecutor

wrangler-core/src/main/java/io/cdap/wrangler/executor/RecipePipelineExecutor.java
vanathi-g: Done

wrangler-api/src/main/java/io/cdap/wrangler/api/Directive.java
```diff
@@ -126,4 +132,20 @@ default List<EntityCountMetric> getCountMetrics() {
...
+   * transformation on the data.
+   * @param inputSchema schema of the input data
+   * @param generator {@link SchemaFieldGenerator} which can generate a {@link Schema.Field}
+   *                                              given a field name and value (Java object)
+   * @param output output data generated by directive after its execution
+   * @return {@link Schema} of the output data
+   * @implNote By default, the implementation returns a null (no-op)
+   * */
+  @Override
+  default Schema getOutputSchema(Schema inputSchema, SchemaFieldGenerator generator, List<Row> output) {
```
tivv: Mark as Nullable. 
Also, why would a Directive need params `generator` and `output`?

tivv: Overall I think we can remove most of the generator changes as those are now internal to the Wrangler itself

vanathi-g: UDDs need a way to generate schema in the same way as Wrangler internally does. Ex: can use this interface in custom implementation of the `getOutputSchema` method to say "generate schema for column A but carry over schema for column B". The internal implementation of schema generation logic can be reused by UDD developers as well

vanathi-g: Output is needed to generate schema for new columns

wrangler-api/src/main/java/io/cdap/wrangler/api/Directive.java
iam-divya: Got it. Missed that.

wrangler-api/src/main/java/io/cdap/wrangler/api/Directive.java
vanathi-g: Changed the `SchemaUpdater` interface to have corresponding methods for all four operations (add, remove, rename and change type). Removed the generator interface and reverted changes to SchemaConverter

wrangler-core/src/main/java/io/cdap/wrangler/executor/RecipePipelineExecutor.java
```diff
@@ -148,9 +147,9 @@ public List<Row> execute(List<Row> rows) throws RecipeException {
             }
           }
           if (schema != null) {
-            Schema previousOutputSchema = context.getTransientStore().get(TRANSIENT_STORE_OUTPUT_SCHEMA_KEY);
-            context.getTransientStore().set(TransientVariableScope.GLOBAL, TRANSIENT_STORE_OUTPUT_SCHEMA_KEY,
-                                            getSchemaUnion(previousOutputSchema, schema));
+            schema = context.getTransientStore().get("ws_schema_output") == null ?
```
tivv, line 150: What is the reason to remove the constant?

wrangler-api/src/main/java/io/cdap/wrangler/api/schema/SchemaUpdater.java
```diff
@@ -0,0 +1,66 @@
...
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * This interface represents an entity that changes the schema of input data after applying a transformation.
+ */
+public interface SchemaUpdater {
+  Map<String, Schema> ADDED_FIELDS = new HashMap<>();
```
tivv: You can't define fields in the interface. This becomes global singletons and I think this is not what you want

tivv (changes requested):

wrangler-api/src/main/java/io/cdap/wrangler/api/schema/SchemaUpdater.java
vanathi-g: Yeah, was just experimenting, thought I'll keep a new `AbstractSchemaUpdater` class that the specific directives can implement. 

I think the easiest approach is to just have an interface with four methods to get collections of added, removed, renamed and type changed fields. They return empty by default, which is backwards compatible as we will generate schema in that case. If the UDD wants to implement the method, they have to override and return custom maps/lists.

Let me know if you have any other suggestions

wrangler-api/src/main/java/io/cdap/wrangler/api/schema/SchemaUpdater.java
```diff
@@ -0,0 +1,44 @@
...
+import io.cdap.cdap.api.data.schema.Schema;
+
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * This interface represents an entity that changes the schema of input data after applying a transformation.
+ */
+public interface SchemaUpdater {
```
tivv: I don't think we need to define it's strictly like this. A simple `@Nullable Schema getOutputSchema(Schema inputRecord)` would work. It would be much more generic as sometimes you do need input schema to define output schema. There are directives that work on a set of fields, like flatten/unflatten the nested struct. It won't be much harder for the directive as it just needs something like `Schema.recordOf(input.getName(), ImmutableList.builder().addAll(input.getFields()).add(Schema.Field.of("newField", nestedSchema)).build())` or similar. We can provide utilities if really needed

There is another question on things like regex parse where number of output fields is based on data, so not sure what are you going to do over there.

vanathi-g: Simplified it, this would mean more repeated code in directives though but I think having one method does make sense in terms of being generic. Since it is just one method, moved it to Executor interface as I don't see any other class in Wrangler that would need to produce an output schema.

wrangler-api/src/main/java/io/cdap/wrangler/api/Executor.java
```diff
@@ -80,5 +82,18 @@ O execute(I rows, ExecutorContext context)
    * correct at this phase of invocation.
    */
   void destroy();
-}
 
+  /**
+   * This method is used to get the updated schema of the data after the directive's transformation has been applied.
+   * @implNote By default, returns a null on the assumption that the schema is unchanged and it is inferred from the
```
tivv: I don't think there is a strong assumption that it's unchanged. At max, "unchanged for retained fields", but it's validated as well

wrangler-core/src/main/java/io/cdap/wrangler/executor/RecipePipelineExecutor.java
```diff
@@ -122,14 +136,25 @@ public List<Row> execute(List<Row> rows) throws RecipeException {
               if (cumulativeRows.size() < 1) {
                 break;
               }
+              if (schema != null) {
+                schema = directive.getOutputSchema(schema) != null ? directive.getOutputSchema(schema) :
```
tivv: Don't run it twice, it can actually be expensive

wrangler-core/src/main/java/io/cdap/wrangler/executor/RecipePipelineExecutor.java
```diff
@@ -122,14 +136,25 @@ public List<Row> execute(List<Row> rows) throws RecipeException {
...
+                  generateOutputSchema(schema, cumulativeRows);
+              }
             } catch (ReportErrorAndProceed e) {
               messages.add(String.format("%s (ecode: %d)", e.getMessage(), e.getCode()));
               collector
                 .add(new ErrorRecord(rows.subList(i, i + 1).get(0), String.join(",", messages), e.getCode(), true));
               cumulativeRows = new ArrayList<>();
               break;
+            } catch (RecordConvertorException e) {
+              throw new RecipeException("Problem generating schema for a field: " + e.getMessage(), e);
```
tivv: I don't think it's generating for "a field"

wrangler-core/src/main/java/io/cdap/wrangler/executor/RecipePipelineExecutor.java
```diff
@@ -122,14 +136,25 @@ public List<Row> execute(List<Row> rows) throws RecipeException {
...
             }
           }
+          if (schema != null) {
+            schema = context.getTransientStore().get(TransientStoreKeys.OUTPUT_SCHEMA) == null ?
+              schema : getSchemaUnion(context.getTransientStore().get(TransientStoreKeys.OUTPUT_SCHEMA), schema);
```
tivv: In what use case you need a union?

wrangler-core/src/main/java/io/cdap/wrangler/schema/Constants.java
```diff
@@ -0,0 +1,34 @@
...
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package io.cdap.wrangler.schema;
+
+/**
+ * Constants for schema handling in Wrangler
+ */
+public class Constants {
```
tivv: final

wrangler-core/src/main/java/io/cdap/wrangler/schema/Constants.java
```diff
@@ -0,0 +1,34 @@
...
+  /**
+   * Constants for storing Workspace schema in TransientStore
+   */
+  public static final class TransientStoreKeys {
```
tivv: Do you really need 2 levels? We already have a number of `Constants` classes in the project and it makes it hard to select which one with completion when writing code

wrangler-core/src/main/java/io/cdap/wrangler/executor/RecipePipelineExecutor.java
```diff
@@ -161,4 +186,38 @@ private List<Directive> getDirectives() throws RecipeException {
     }
     return directives;
   }
+
+  private Schema generateOutputSchema(Schema inputSchema, List<Row> output) throws RecordConvertorException {
+    Set<Field> outputFields = new HashSet<>();
```
tivv: I don't think Field Implements hashcode/equals. Also this way you loose the order

wrangler-core/src/main/java/io/cdap/wrangler/executor/RecipePipelineExecutor.java
```diff
@@ -161,4 +186,38 @@ private List<Directive> getDirectives() throws RecipeException {
...
+        Object fieldValue = rowField.getSecond();
+
+        Schema existing = inputSchema.getField(fieldName) != null ? inputSchema.getField(fieldName).getSchema() : null;
+        Schema generated = fieldValue != null && (existing == null || !isValidSchemaForValue(existing, fieldValue)) ?
+          generator.getSchema(fieldValue, fieldName) : null;
+
+        if (generated != null) {
+          outputFields.add(Schema.Field.of(fieldName, generated));
+        } else if (existing != null) {
+          outputFields.add(Schema.Field.of(fieldName, existing));
```
tivv: Do you want to add a field schema for each row?

wrangler-core/src/main/java/io/cdap/wrangler/executor/RecipePipelineExecutor.java
```diff
@@ -161,4 +186,38 @@ private List<Directive> getDirectives() throws RecipeException {
...
+  private Schema generateOutputSchema(Schema inputSchema, List<Row> output) throws RecordConvertorException {
+    Set<Field> outputFields = new HashSet<>();
+    for (Row row : output) {
+      for (Pair<String, Object> rowField : row.getFields()) {
+        String fieldName = rowField.getFirst();
+        Object fieldValue = rowField.getSecond();
+
+        Schema existing = inputSchema.getField(fieldName) != null ? inputSchema.getField(fieldName).getSchema() : null;
+        Schema generated = fieldValue != null && (existing == null || !isValidSchemaForValue(existing, fieldValue)) ?
+          generator.getSchema(fieldValue, fieldName) : null;
```
tivv: You may want to change existing to nullable if it is not nullable and current value is null

wrangler-core/src/main/java/io/cdap/wrangler/executor/RecipePipelineExecutor.java
vanathi-g: Since some directives can generate output rows that have different columns (ex parsers), we need to take a union of previous row schema and current one. Have modified the variable names and added comments to add clarity

wrangler-core/src/main/java/io/cdap/wrangler/executor/RecipePipelineExecutor.java
vanathi-g: Fixed by using a LinkedHashMap to map potentially multiple schemas to one field name (the last one is taken) and also preserve insertion order (current behaviour when schema is generated)

wrangler-core/src/main/java/io/cdap/wrangler/executor/RecipePipelineExecutor.java
vanathi-g: Could you clarify what exactly the issue here is?

wrangler-service/src/main/java/io/cdap/wrangler/service/directive/AbstractDirectiveHandler.java
```diff
@@ -264,6 +266,21 @@ protected WorkspaceValidationResult getWorkspaceSummary(List<Row> rows) throws E
     return new WorkspaceValidationResult(columnValidationResults, statistics);
   }
 
+  /**
+   * Method to get the list of columns across all the given rows, while maintaining the order
+   * @param rows list of rows
+   * @return list of columns (union across columns in all rows)
+   */
+  public static List<String> getAllColumns(List<Row> rows) {
```
saimukkamala: What order are we talking about?
If we have 2 rows with columns [A,B,D] and [A,B,C,D], this function will return [A,B,D,C]. Is this expected?

saimukkamala: This seems to be the existing behaviour in AbstractDirectiveHandler.generateExecutionResponse(). Probably `while maintaining the order` is a bit misleading.

wrangler-service/src/main/java/io/cdap/wrangler/service/directive/AbstractDirectiveHandler.java
```diff
@@ -156,8 +158,18 @@ protected DirectiveExecutionResponse generateExecutionResponse(
     List<Map<String, Object>> values = new ArrayList<>(rows.size());
     Map<String, String> types = new HashMap<>();
     Set<String> headers = new LinkedHashSet<>();
```
saimukkamala: Can we merge `headers` and `types` into a single `LinkedHashMap`? In this function, `types` are generated from output schema and `headers` are generated from row data. Both should be generated from output schema. Also maintaining 2 different objects containing same information can lead to issues.

vanathi-g: Done, verified that LinkedHashMap's keyset method also maintains ordering

saimukkamala (approved):

wrangler-core/src/main/java/io/cdap/wrangler/executor/RecipePipelineExecutor.java
```diff
@@ -161,4 +188,55 @@ private List<Directive> getDirectives() throws RecipeException {
...
   }
+
+  private Schema generateOutputSchema(Schema inputSchema, List<Row> output) throws RecordConvertorException {
+    Map<String, Schema> outputFieldMap = new LinkedHashMap<>();
+    for (Row row : output) {
+      for (Pair<String, Object> rowField : row.getFields()) {
+        String fieldName = rowField.getFirst();
+        Object fieldValue = rowField.getSecond();
+
+        Schema existing = inputSchema.getField(fieldName) != null ? inputSchema.getField(fieldName).getSchema() : null;
```
tivv: nit: better to use `Optional`

wrangler-core/src/main/java/io/cdap/wrangler/executor/RecipePipelineExecutor.java
tivv: I believe this is still applicable what should happen if all values are null, but the original schema is non-nullable

wrangler-core/src/main/java/io/cdap/wrangler/executor/RecipePipelineExecutor.java
```diff
@@ -122,14 +137,26 @@ public List<Row> execute(List<Row> rows) throws RecipeException {
               if (cumulativeRows.size() < 1) {
                 break;
               }
+              if (designTime && schema != null) {
+                Schema directiveOutputSchema = directive.getOutputSchema(schema);
+                schema = directiveOutputSchema != null ? directiveOutputSchema
+                  : generateOutputSchema(schema, cumulativeRows);
```
tivv: We should introduce a feature flag for this with default enabled.. If schema generator fails on some customer schema, there should be a way to disable it

wrangler-core/src/main/java/io/cdap/wrangler/executor/RecipePipelineExecutor.java
```diff
@@ -161,4 +188,55 @@ private List<Directive> getDirectives() throws RecipeException {
...
+      for (Pair<String, Object> rowField : row.getFields()) {
+        String fieldName = rowField.getFirst();
+        Object fieldValue = rowField.getSecond();
+
+        Schema existing = inputSchema.getField(fieldName) != null ? inputSchema.getField(fieldName).getSchema() : null;
+        Schema generated = fieldValue != null && !isValidSchemaForValue(existing, fieldValue) ?
+          generator.getSchema(fieldValue, fieldName) : null;
+
+        if (generated != null) {
+          outputFieldMap.put(fieldName, generated);
```
tivv: Does it mean you always use the schema from the last non-null row? If so, there is no reason to run generator many times

wrangler-core/src/main/java/io/cdap/wrangler/executor/RecipePipelineExecutor.java
```diff
@@ -161,4 +188,55 @@ private List<Directive> getDirectives() throws RecipeException {
...
+    }
+    Schema generated = generator.getSchema(value, TEMP_SCHEMA_FIELD_NAME);
+    generated = generated.isNullable() ? generated.getNonNullable() : generated;
+    schema = schema.isNullable() ? schema.getNonNullable() : schema;
+    return generated.getType().equals(schema.getType());
+  }
+
+  // Gets the union of fields in two schemas while maintaining insertion order and uniqueness of fields. If the same
+  // field exists with two different schemas, the second schema overwrites first one
+  private Schema getSchemaUnion(Schema first, Schema second) {
```
tivv: If which cases would it be useful? I believe we always generate schema for all field of the row. 
E.g. if original schema of the field is "string" and resulting generated schema is "nullable number", this algorythm with randomly (depending on a sample) generate either "string" or "nullable number"

wrangler-api/src/main/java/io/cdap/wrangler/api/Executor.java
```diff
@@ -80,5 +82,18 @@ O execute(I rows, ExecutorContext context)
...
+  /**
+   * This method is used to get the updated schema of the data after the directive's transformation has been applied.
+   * @implNote By default, returns a null and the schema is inferred from the data when necessary.
+   * <p>For consistent handling, override for directives that perform column renames,
+   * column data type changes or column additions with specific schemas.</p>
+   * @param inputSchema input {@link Schema} of the data before transformation
+   * @return output {@link Schema} of the transformed data
+   */
+  @Nullable
+  default Schema getOutputSchema(Schema inputSchema) {
```
tivv: What's the plan for directives that generate output schema based on input data (e.g. string parsing into multiple fields)? Are we going to rely on the detection in this case?

vanathi-g: Yes, the detection will take care of that. Even if we add logic to the directive, it will be redundant as it will be the same schema generation that will be done

wrangler-core/src/main/java/io/cdap/wrangler/executor/RecipePipelineExecutor.java
vanathi-g: That's how it is currently done (without schema addition). The last row of the sample's data is ultimately used to determine the type. So in this case, we do a schema union for a different reason. Lets say we have the json parsing directive which generates two rows with these columns:
- col_A: int, col_B: string
- col_B: string, col_C: float

Now, after the schema has been taken at the end of executing the directives on the first row, it will have columns A and B. We need to take a union after executing the directives on the second row as well to pick up all 3 columns in the schema.

wrangler-api/src/main/java/io/cdap/wrangler/api/Executor.java
tivv: Given we don't provide a full covering solution here, could you please introduce some interface, e.g. `OutputSchemaResolutionContext` with a single property `inputSchema` and pass it instead? This way we can later add more stuff as needed without contract change. I am still not sure if we would need to provide rows at some point

wrangler-core/src/main/java/io/cdap/wrangler/executor/RecipePipelineExecutor.java
```diff
@@ -161,4 +187,16 @@ private List<Directive> getDirectives() throws RecipeException {
     }
     return directives;
   }
+
+  public static Row getRowUnion(List<Row> rows) {
```
tivv: I would not do it, but just feed everything to `addDirectiveField` You are essentially do the same here and there

wrangler-core/src/main/java/io/cdap/wrangler/utils/OutputSchemaGenerator.java
```diff
@@ -0,0 +1,132 @@
...
+import java.util.Map;
+import javax.annotation.Nullable;
+
+/**
+ * This class can be used to generate the output schema after executing a set of directives. A list is maintained
+ * where each element is a map of fieldName --> value. Each element of the list corresponds to a directive in the list
+ * provided during initialization. Hence, each map represents the fields present across all output rows generated by a
+ * directive after execution.
+ */
+public class OutputSchemaGenerator {
```
tivv: With the current design this class is very tightly coupled with `RecipePipelineExecutor` in fragile ways (e.g. using indexes to something passed in constructor is fragile.
I would move `generateDirectiveOutputSchema` back to  `RecipePipelineExecutor`. This way this class does not need `directives` and essentially becomes a fixed-size table of rows.

Or alternatively, make it store just a single directive with all the data and make `RecipePipelineExecutor` manage the list. It already has a list of `directives`, you can add another one or replace the original list with a new one

vanathi-g: Changed it to `DirectiveOutputSchemaGenerator` and handled the list separately

wrangler-core/src/main/java/io/cdap/wrangler/utils/DirectiveOutputSchemaGenerator.java
```diff
@@ -0,0 +1,111 @@
...
+      for (Pair<String, Object> field : row.getFields()) {
+        String fieldName = field.getFirst();
+        Object fieldValue = field.getSecond();
+        if (outputFieldMap.containsKey(fieldName)) {
+          // If existing value is null, override with this non-null value
+          if (fieldValue != null && outputFieldMap.get(fieldName) == null) {
+            outputFieldMap.put(fieldName, fieldValue);
+          }
+        } else {
+          outputFieldMap.putIfAbsent(fieldName, fieldValue);
```
tivv: nit: Any specific reason to use `putIfAbsent`? You already checked that the field is not there. Also `putIfAbsent` may work weirdly for maps with null values.

wrangler-core/src/main/java/io/cdap/wrangler/utils/DirectiveOutputSchemaGenerator.java
```diff
@@ -0,0 +1,111 @@
...
+   */
+  public Schema getDirectiveOutputSchema(Schema inputSchema) throws RecordConvertorException {
+    Schema directiveOutputSchema = directive.getOutputSchema(inputSchema);
+    return directiveOutputSchema != null ? directiveOutputSchema : generateDirectiveOutputSchema(inputSchema);
+  }
+
+  // Given the schema from previous step and output of current directive, generates the directive output schema.
+  private Schema generateDirectiveOutputSchema(Schema inputSchema)
+    throws RecordConvertorException {
+    List<Schema.Field> outputFields = new LinkedList<>();
```
tivv: nit: In 90% of cases LinkedList is slower than ArrayList. Use ArrayList unless you have very specific reasons to use LinkedList

wrangler-core/src/main/java/io/cdap/wrangler/utils/DirectiveOutputSchemaGenerator.java
```diff
@@ -0,0 +1,111 @@
...
+    for (String fieldName : outputFieldMap.keySet()) {
+      Object fieldValue = outputFieldMap.get(fieldName);
```
tivv: nit: just iterate on the `entrySet` This will save you a lookup

wrangler-core/src/main/java/io/cdap/wrangler/utils/DirectiveOutputSchemaGenerator.java
```diff
@@ -0,0 +1,111 @@
...
+
+      Schema existing = inputSchema.getField(fieldName) != null ? inputSchema.getField(fieldName).getSchema() : null;
+      Schema generated = fieldValue != null && !isValidSchemaForValue(existing, fieldValue) ?
+        schemaGenerator.getSchema(fieldValue, fieldName) : null;
+
+      if (generated != null) {
+        outputFields.add(Schema.Field.of(fieldName, generated));
+      } else if (existing != null) {
+        outputFields.add(Schema.Field.of(fieldName, existing));
```
tivv: If you get here it may mean that all values were null. Ensure you make existing schema nullable if it's not

wrangler-api/src/main/java/io/cdap/wrangler/api/Executor.java
tivv: Reopening, please see previous comment

wrangler-service/src/main/java/io/cdap/wrangler/service/directive/AbstractDirectiveHandler.java
```diff
@@ -154,10 +157,19 @@ protected <E extends Exception> List<Row> executeDirectives(
...
     List<Row> rows, int limit) throws Exception {
     List<Map<String, Object>> values = new ArrayList<>(rows.size());
-    Map<String, String> types = new HashMap<>();
-    Set<String> headers = new LinkedHashSet<>();
-    SchemaConverter convertor = new SchemaConverter();
-
+    Map<String, String> types = new LinkedHashMap<>();
+
+    Schema outputSchema = TRANSIENT_STORE.get(TransientStoreKeys.OUTPUT_SCHEMA) != null ?
+      TRANSIENT_STORE.get(TransientStoreKeys.OUTPUT_SCHEMA) : TRANSIENT_STORE.get(TransientStoreKeys.INPUT_SCHEMA);
```
tivv, line 162: In which cases OUTPUT_SCHEMA will be null? This "else" part seems problematic for me.

vanathi-g: Initially only input schema is set in transient store (when workspace created for ex)

tivv (approved):

PR #646, opened 2023-06-20
Column Transform directives

shrverma:

- Copy
- Keep
- Merge
- Rename
- Trim
- Left-trim
- Right-trim
- Lowercase
- Uppercase
- Titlecase


wrangler-core/src/main/java/io/cdap/directives/column/Keep.java
```diff
@@ -93,4 +103,24 @@ public Mutation lineage() {
...
+    Optional<ExpressionFactory<String>> expressionFactory = getExpressionFactory(relationalTranformContext);
+    if (!expressionFactory.isPresent()) {
+      return new InvalidRelation("Cannot find an Expression Factory");
+    }
+    Map<String, Expression> keepCol = generateColumnExpMap(keep.stream().collect(Collectors.toList()),
+            expressionFactory.get());
+    return relation.select(keepCol);
+  }
+
+  private Optional<ExpressionFactory<String>> getExpressionFactory(RelationalTranformContext ctx) {
```
saimukkamala: This function seems to be common across directives. can we move this to `Directive` interface or move it to a util class

wrangler-core/src/main/java/io/cdap/directives/column/Keep.java
```diff
@@ -93,4 +103,24 @@ public Mutation lineage() {
...
+    return ctx.getEngine().getExpressionFactory(StringExpressionFactoryType.SQL);
+  }
+  private Map<String, Expression> generateColumnExpMap(List<String> columns, ExpressionFactory<String> factory) {
```
saimukkamala: Even this function seems like a util function rather than directive specific function. can we move this to util class or `Directive` interface.

shrverma: Yes. Will move it to a separate file.

wrangler-api/src/main/java/io/cdap/wrangler/api/Directive.java
```diff
@@ -137,4 +143,24 @@ default Relation transform(RelationalTranformContext relationalTranformContext,
     // no-op
     return relation;
   }
+
+  default Optional<ExpressionFactory<String>> getExpressionFactory(RelationalTranformContext ctx) {
```
vanathi-g: This can be moved to a separate interface as it doesn't make sense for a directive to have this. Maybe a new interface that extends `LinearRelationalTransform` that is internal to Wrangler?

wrangler-api/src/main/java/io/cdap/wrangler/api/Directive.java
```diff
@@ -137,4 +143,24 @@ default Relation transform(RelationalTranformContext relationalTranformContext,
...
+   return ctx.getEngine().getExpressionFactory(StringExpressionFactoryType.SQL);
+  }
+
+  default Map<String, Expression> generateColumnExpMap(List<String> columns, ExpressionFactory<String> factory) {
+    Map<String, Expression> columnExpMap = new LinkedHashMap<>();
+    columns.forEach((colName)-> columnExpMap.put(colName, factory.compile(colName)));
+    return columnExpMap;
+  }
 }
+
```
vanathi-g: nit: remove empty lines and other unnecessary spacing changes

wrangler-core/src/main/java/io/cdap/directives/column/ChangeColCaseNames.java
```diff
@@ -94,5 +104,40 @@ public Mutation lineage() {
       .all(Many.of())
       .build();
   }
+  @Override
+  public Relation transform(RelationalTranformContext relationalTranformContext,
+                            Relation relation) {
+    java.util.Optional<ExpressionFactory<String>> expressionFactory = getExpressionFactory(relationalTranformContext);
```
vanathi-g: nit: add import for Optional separately, change in other places too as necessary

wrangler-core/src/main/java/io/cdap/wrangler/utils/ColumnTypeExpression.java
```diff
@@ -0,0 +1,84 @@
...
+ *  License for the specific language governing permissions and limitations under
+ *  the License.
+ */
+
+package io.cdap.wrangler.utils;
+
+/**
+ * Utility class that returns a string of SQL expression for the given data type.
+ */
+public final class ColumnTypeExpression {
```
vanathi-g: Can rename to SQLExpressionGenerator and have all (future) utility methods related to SQL generation here. Ex: getColumnTypeSQLExpression

wrangler-core/src/main/java/io/cdap/directives/transformation/Upper.java
```diff
@@ -29,10 +35,18 @@
...
 import io.cdap.wrangler.api.parser.UsageDefinition;
 
+import java.util.ArrayList;
+import java.util.HashSet;
+import java.util.LinkedHashMap;
 import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import java.util.Set;
+import java.util.stream.Collectors;
```
vanathi-g: nit: Remove unused imports wherever necessary

wrangler-core/src/main/java/io/cdap/directives/column/ChangeColCaseNames.java
```diff
@@ -94,5 +104,40 @@ public Mutation lineage() {
...
+    List<String> columnNames = generateListCols(relationalTranformContext);
+    Map<String, Expression> colmap = generateColumnCaseMap(columnNames, expressionFactory.get());
+    return relation.select(colmap);
+  }
+
+  private List<String> generateListCols(RelationalTranformContext relationalTranformContext) {
+    List<String> colnames = new ArrayList<String>();
+    java.util.Set<String> s = relationalTranformContext.getInputRelationNames();
+    for (String inp : s) {
+      Schema schema = relationalTranformContext.getInputSchema(inp);
```
vanathi-g: Right now this is not an issue as we only support executing one directive in SQL. But later on if we add multiple directives support, input schema may not be the correct thing to use as columns may have been dropped/added/renamed in previous steps. Will the input schema passed have the previous directive's transforms applied in that case?

wrangler-core/src/main/java/io/cdap/directives/column/Keep.java
```diff
@@ -93,4 +103,18 @@ public Mutation lineage() {
...
     return builder.build();
   }
+  @Override
+  public Relation transform(RelationalTranformContext relationalTranformContext,
+                            Relation relation) {
+    Optional<ExpressionFactory<String>> expressionFactory = getExpressionFactory(relationalTranformContext);
+    if (!expressionFactory.isPresent()) {
+      return new InvalidRelation("Cannot find an Expression Factory");
+    }
+    Map<String, Expression> keepCol = generateColumnExpMap(keep.stream().collect(Collectors.toList()),
```
vanathi-g: To avoid having to stream and collect a set to a list, you can modify the function to take a collection as an argument. There are common methods to iterate over a collection

vanathi-g (changes requested):

wrangler-core/src/main/java/io/cdap/directives/column/Merge.java
```diff
@@ -108,4 +114,15 @@ public Mutation lineage() {
...
   }
+  public Relation transform(RelationalTranformContext relationalTranformContext,
+                            Relation relation) {
+    Optional<ExpressionFactory<String>> expressionFactory = SqlExpressionGenerator
+            .getExpressionFactory(relationalTranformContext);
+    if (!expressionFactory.isPresent()) {
+      return new InvalidRelation("Cannot find an Expression Factory");
+    }
+    return relation.setColumn(dest, expressionFactory.get()
+            .compile("CONCAT(" + col1 + ",'" + delimiter + "'," + col2 + ")"));
```
vanathi-g: Might be more readable to use String.format() in places like these

wrangler-core/src/main/java/io/cdap/wrangler/utils/SqlTypeExpGenerator.java
```diff
@@ -0,0 +1,85 @@
...
+ */
+
+package io.cdap.wrangler.utils;
+
+/**
+ * Utility class that returns a string of SQL expression for the given data type.
+ */
+public final class SqlTypeExpGenerator {
+
+    public static String getColumnTypeExp(String toType, String column, int scale) {
```
vanathi-g: Change this to have just one function and keep all the logic here. Mark the 3rd argument scale as nullable and you can pass in a null here. Then change the if-else logic in SetType where this method is called

wrangler-core/src/main/java/io/cdap/wrangler/utils/SqlTypeExpGenerator.java
```diff
@@ -0,0 +1,85 @@
...
+ *  License for the specific language governing permissions and limitations under
+ *  the License.
+ */
+
+package io.cdap.wrangler.utils;
+
+/**
+ * Utility class that returns a string of SQL expression for the given data type.
+ */
+public final class SqlTypeExpGenerator {
```
vanathi-g: Can consider moving the function in this class to SqlExpressionGenerator and delete this class altogether

saimukkamala (approved):

PR #648, opened 2023-07-11
UI change wrangler

shrverma:
Adding a UI toggle for the user to enable directive execution on SQL engine.
jexl
Screenshot from 2023-07-14 15-17-39


wrangler-transform/src/main/java/io/cdap/wrangler/Wrangler.java
```diff
@@ -752,10 +764,17 @@ public String getOnError() {
     public String getPreconditionLanguage() {
       if (Strings.isNullOrEmpty(preconditionLanguage)) {
```
saimukkamala: Its not a good idea to have logic in `Config` class. Logic of handling Null or empty string should be handled by the caller.

wrangler-transform/src/main/java/io/cdap/wrangler/Wrangler.java
```diff
@@ -752,10 +764,17 @@ public String getOnError() {
...
         // due to backward compatibility...
-        return PRECONDITION_LANGUAGE_JEXL;
+        return LANGUAGE_JEXL;
       }
       return preconditionLanguage;
     }
+    public String getExecutionLanguage() {
+      if (Strings.isNullOrEmpty(executionLanguage)) {
```
saimukkamala: Its not a good idea to have logic in Config class. Logic of handling Null or empty string should be handled by the caller.

wrangler-transform/src/main/java/io/cdap/wrangler/Wrangler.java
```diff
@@ -187,9 +185,13 @@ public void configurePipeline(PipelineConfigurer configurer) {
           directives = String.format("#pragma load-directives %s;%s", config.getUDDs(), config.getDirectives());
         }
       }
-
+      if (LANGUAGE_SQL.equalsIgnoreCase(config.getExecutionLanguage())
+              && Strings.isNullOrEmpty(config.preconditionLanguage)) {
+        config.preconditionLanguage = LANGUAGE_SQL;
```
saimukkamala: Seems like we are checking the if condition multiple times in this class. Can we move this to a common function and add some comments to make it more readable?

saimukkamala: We can determine and store the mode of execution in a global variable. I believe we have 2 modes
1. jexl execution mode.
2. sql engine execution mode.

wrangler-transform/src/main/java/io/cdap/wrangler/Wrangler.java
```diff
@@ -688,6 +693,12 @@ public static class Config extends PluginConfig {
     @Nullable
     private String preconditionLanguage;
 
+    @Name(NAME_EXECUTION_LANGUAGE)
```
saimukkamala: Rather than `EXECUTION_LANGUAGE`, I believe `USE_SQL_ENGINE` would make more sense.

wrangler-transform/src/main/java/io/cdap/wrangler/Wrangler.java
```diff
@@ -414,7 +417,7 @@ public void transform(StructuredRecord input, Emitter<StructuredRecord> emitter)
       }
 
       // If pre-condition is set, then evaluate the precondition
-      if (PRECONDITION_LANGUAGE_JEXL.equalsIgnoreCase(config.getPreconditionLanguage())
+      if (LANGUAGE_JEXL.equalsIgnoreCase(config.getExecutionLanguage())
           && checkPreconditionNotEmpty(false)) {
```
saimukkamala: This logic is executed on every record which is not efficient. Can we execute this logic once and store the result in a global variable?

wrangler-transform/src/main/java/io/cdap/wrangler/Wrangler.java
```diff
@@ -189,11 +190,11 @@ public void configurePipeline(PipelineConfigurer configurer) {
       }
 
       if (!config.containsMacro(Config.NAME_PRECONDITION_LANGUAGE)) {
-        if (PRECONDITION_LANGUAGE_SQL.equalsIgnoreCase(config.getPreconditionLanguage())) {
+        if (checkSQLExecution(config)) {
```
shrverma, line 195: These changes will allow users to run directives in SQL along with the existing precondition feature flag.

vanathi-g (approved):
LGTM, as long as allowing SQL execution of directives is alright even when just precondition SQL is enabled.

wrangler-transform/src/main/java/io/cdap/wrangler/Wrangler.java
```diff
@@ -547,6 +540,25 @@ private boolean checkPreconditionNotEmpty(Boolean isConditionSQL) {
...
+            || Feature.WRANGLER_EXECUTION_SQL.isEnabled(getContext()))) {
+      // disabling SQL execution for precondition and directives
+      return false;
+    }
+
+    if (!Strings.isNullOrEmpty(config.getPreconditionLanguage())) {
+      return SQL.equalsIgnoreCase(config.getPreconditionLanguage());
+    }
+
+    if (!Strings.isNullOrEmpty(config.getSqlExecution())) {
```
iam-divya, line 556: We are enabling SQL only in certain directives and not all. Where is that condition checked? Even when the SQL mode is enabled, If the given list of directives cant be supported in SQL, we should revert back to old way of execution.

wrangler-transform/src/main/java/io/cdap/wrangler/Wrangler.java
```diff
@@ -547,6 +540,25 @@ private boolean checkPreconditionNotEmpty(Boolean isConditionSQL) {
...
+      return SQL_ENABLED.equalsIgnoreCase(config.getSqlExecution());
+    }
```
iam-divya, line 558: Did we test this with biq query as well? Pushdown will enable running these in BQ, we need to test or handle it accordingly.

wrangler-transform/src/main/java/io/cdap/wrangler/Wrangler.java
```diff
@@ -547,6 +540,25 @@ private boolean checkPreconditionNotEmpty(Boolean isConditionSQL) {
...
+
+    // for backwards compatibility
+    return false;
+  }
```
iam-divya, line 562: Existing precondition filter should not be impacted. Please make sure it is well tested with this change.
1. Precondition with JEXL
2. Precondition with SQL
3. Precondition SQL with pushdown enabled.

wrangler-transform/src/main/java/io/cdap/wrangler/Wrangler.java
```diff
@@ -688,6 +696,12 @@ public static class Config extends PluginConfig {
     @Nullable
     private String preconditionLanguage;
 
+    @Name(NAME_SQL_EXECUTION)
+    @Description("Toggle to configure execution language between JEXL and SQL")
```
iam-divya, line 700: Hows this different from preconditionLanguage?

wrangler-transform/widgets/Wrangler-transform.json
```diff
@@ -50,7 +69,7 @@
           "label": "Precondition (SQL)",
           "name": "preconditionSQL",
           "widget-attributes" : {
-            "default" : "false"
+            "default" : "true"
```
iam-divya, line 72: Is this added for testing to enable Precondition SQL by default?

wrangler-transform/src/main/java/io/cdap/wrangler/Wrangler.java
shrverma: It is basically the same thing, just needed a new toggle for backwards compatibility.

wrangler-transform/src/main/java/io/cdap/wrangler/Wrangler.java
shrverma: Sure.

wrangler-transform/src/main/java/io/cdap/wrangler/Wrangler.java
shrverma: In that case if the directive is not supported, it should return an error and normal flow should work.

wrangler-transform/src/main/java/io/cdap/wrangler/Wrangler.java
shrverma: Yes BQ pushdown is running.

wrangler-transform/src/main/java/io/cdap/wrangler/Wrangler.java
iam-divya: Is that tested?

shrverma: Will add changes to validate supported directives in the next PR as it requires some changes to the directive interface. For now if the directive is not supported the normal execution takes place.

wrangler-transform/src/main/java/io/cdap/wrangler/Wrangler.java
```diff
@@ -414,8 +413,7 @@ public void transform(StructuredRecord input, Emitter<StructuredRecord> emitter)
       }
 
       // If pre-condition is set, then evaluate the precondition
-      if (PRECONDITION_LANGUAGE_JEXL.equalsIgnoreCase(config.getPreconditionLanguage())
-          && checkPreconditionNotEmpty(false)) {
+      if (!checkSQLExecution(config) && checkPreconditionNotEmpty(false)) {
```
saimukkamala, line 418: Only concern here, we are checking SQL execution for every row. This is not efficient. Can we have a global variable and initialize it in `initialize` function. afaik `initialize` is executed on all spark nodes.

saimukkamala (approved):
Approved with one suggested change

PR #649, opened 2023-07-14
Add custom schema handling logic to directives that rename columns

vanathi-g:
## Changes
- Override getOutputSchema method in `rename`, `set-headers`, `swap`, `cleanse-columns` directives
- All of these perform column renames, hence the necessary column schemas are also modified to avoid generating a new schema for renamed columns

### DIFF
--- a/wrangler-core/src/main/java/io/cdap/directives/column/CleanseColumnNames.java
+++ b/wrangler-core/src/main/java/io/cdap/directives/column/CleanseColumnNames.java
@@ -19,12 +19,14 @@ package io.cdap.directives.column;
 import io.cdap.cdap.api.annotation.Description;
 import io.cdap.cdap.api.annotation.Name;
 import io.cdap.cdap.api.annotation.Plugin;
+import io.cdap.cdap.api.data.schema.Schema;
 import io.cdap.wrangler.api.Arguments;
 import io.cdap.wrangler.api.Directive;
 import io.cdap.wrangler.api.DirectiveExecutionException;
 import io.cdap.wrangler.api.DirectiveParseException;
 import io.cdap.wrangler.api.ExecutorContext;
 import io.cdap.wrangler.api.Row;
+import io.cdap.wrangler.api.SchemaResolutionContext;
 import io.cdap.wrangler.api.annotations.Categories;
 import io.cdap.wrangler.api.lineage.Lineage;
 import io.cdap.wrangler.api.lineage.Many;
@@ -32,6 +34,7 @@ import io.cdap.wrangler.api.lineage.Mutation;
 import io.cdap.wrangler.api.parser.UsageDefinition;
 
 import java.util.List;
+import java.util.stream.Collectors;
 
 /**
  * A directive for cleanses columns names.
@@ -73,13 +76,7 @@ public final class CleanseColumnNames implements Directive, Lineage {
     for (Row row : rows) {
       for (int i = 0; i < row.width(); ++i) {
         String column = row.getColumn(i);
-        // Trims
-        column = column.trim();
-        // Lower case columns
-        column = column.toLowerCase();
-        // Filtering unwanted characters
-        column = column.replaceAll("[^a-zA-Z0-9_]", "_");
-        row.setColumn(i, column);
+        row.setColumn(i, cleanseColumnName(column));
       }
     }
     return rows;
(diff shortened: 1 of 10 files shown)


wrangler-core/src/main/java/io/cdap/directives/column/CleanseColumnNames.java
```diff
@@ -79,6 +84,7 @@ public List<Row> execute(List<Row> rows, ExecutorContext context) throws Directi
         column = column.toLowerCase();
         // Filtering unwanted characters
         column = column.replaceAll("[^a-zA-Z0-9_]", "_");
+        cleansedColumnNames.put(row.getColumn(i), column);
```
saimukkamala: we should not depend on row data. All we need is input schema to generate output schema. I believe we can apply the same column transformations on input schema to get the output schema.

wrangler-core/src/main/java/io/cdap/directives/column/SetHeader.java
```diff
@@ -75,6 +80,7 @@ public List<Row> execute(List<Row> rows, ExecutorContext context)
       int idx = 0;
       for (String name : columns) {
         if (idx < row.width()) {
+          mappedColumnNames.put(row.getColumn(idx), name);
```
saimukkamala: Same here. we should not use row data to get output schema.

wrangler-core/src/main/java/io/cdap/directives/column/CleanseColumnNames.java
vanathi-g: Done

wrangler-core/src/main/java/io/cdap/directives/column/Rename.java
```diff
@@ -82,4 +85,17 @@ public Mutation lineage() {
       .relation(source, target)
       .build();
   }
+
+  @Override
+  public Schema getOutputSchema(SchemaResolutionContext context) {
+    Schema inputSchema = context.getInputSchema();
+    return Schema.recordOf(
+      "outputSchema",
+      inputSchema.getFields().stream()
```
saimukkamala, line 94: nit: can we use `getField` function? https://github.com/cdapio/cdap/blob/develop/cdap-api-common/src/main/java/io/cdap/cdap/api/data/schema/Schema.java#L783 rather than streaming through all fields?

wrangler-core/src/main/java/io/cdap/directives/column/SetHeader.java
```diff
@@ -90,5 +92,14 @@ public Mutation lineage() {
...
       .build();
   }
-}
 
+  @Override
+  public Schema getOutputSchema(SchemaResolutionContext context) {
+    List<Schema.Field> fields = new ArrayList<>(context.getInputSchema().getFields());
+    for (int i = 0; i < columns.size(); i++) {
+      Schema.Field existing = fields.remove(i);
+      fields.add(i, Schema.Field.of(columns.get(i), existing.getSchema()));
```
saimukkamala: 1. Will there be a case where number of elements in `columns` are greater than number of elements in `inputSchema`? in that case, `existing.getSchema()` might throw a `NullPointerException`?
2. Why are we removing and adding elements to existing input schema rather than creating a new list? if we can, then we should for readability.

saimukkamala (approved):

wrangler-core/src/main/java/io/cdap/directives/column/Rename.java
vanathi-g: So the alternative approach is to create a new list of fields from inputSchema.getFields() and then iterate over every field to find the one to rename. We can't use any of the list compare/contains/replace methods since the Field object can't be compared directly (need to compare based on name manually)

wrangler-core/src/main/java/io/cdap/directives/column/SetHeader.java
```diff
@@ -95,11 +95,17 @@ public Mutation lineage() {
...
   @Override
   public Schema getOutputSchema(SchemaResolutionContext context) {
-    List<Schema.Field> fields = new ArrayList<>(context.getInputSchema().getFields());
+    List<Schema.Field> inputFields = context.getInputSchema().getFields();
+    List<Schema.Field> outputFields = new ArrayList<>();
     for (int i = 0; i < columns.size(); i++) {
-      Schema.Field existing = fields.remove(i);
-      fields.add(i, Schema.Field.of(columns.get(i), existing.getSchema()));
+      if (i < inputFields.size()) {
+        outputFields.add(Schema.Field.of(columns.get(i), inputFields.get(i).getSchema()));
```
saimukkamala: Each column name is trimmed in `execute` implementation. can we also add here `columns.get(i).trim()`

wrangler-core/src/main/java/io/cdap/directives/column/SetHeader.java
```diff
@@ -95,11 +95,17 @@ public Mutation lineage() {
...
+      }
+    }
+    // Leftover columns (not renamed)
+    for (int i = columns.size(); i < inputFields.size(); i++) {
```
saimukkamala, line 104: from this logic, if `columns` is larger than `inputFields`, the extra columns from `columns` are not added to the output schema. Or am I missing anything?

wrangler-core/src/main/java/io/cdap/directives/column/SetHeader.java
```diff
@@ -95,11 +95,17 @@ public Mutation lineage() {
 
   @Override
   public Schema getOutputSchema(SchemaResolutionContext context) {
-    List<Schema.Field> fields = new ArrayList<>(context.getInputSchema().getFields());
+    List<Schema.Field> inputFields = context.getInputSchema().getFields();
+    List<Schema.Field> outputFields = new ArrayList<>();
     for (int i = 0; i < columns.size(); i++) {
-      Schema.Field existing = fields.remove(i);
-      fields.add(i, Schema.Field.of(columns.get(i), existing.getSchema()));
+      if (i < inputFields.size()) {
```
saimukkamala: nit: this if condition can be added in for loop condition. `for (int i = 0; i < columns.size() && i < inputFields.size(); i++)`

PR #651, opened 2023-07-25
Transformation and Row Directives

shrverma:

- Create-record
- Swap
- Fixed-length-parser
- Record-regex-filter
- Set-record-delimiter
- Split-to-rows
- Fill-null-or-empty
- Generate-UUID
- Split-email
- Url-encode
- Url-decode
- Write-as-json-object

### DIFF
--- a/wrangler-core/src/main/java/io/cdap/directives/column/CreateRecord.java
+++ b/wrangler-core/src/main/java/io/cdap/directives/column/CreateRecord.java
@@ -19,6 +19,10 @@ package io.cdap.directives.column;
 import io.cdap.cdap.api.annotation.Description;
 import io.cdap.cdap.api.annotation.Name;
 import io.cdap.cdap.api.annotation.Plugin;
+import io.cdap.cdap.etl.api.relational.ExpressionFactory;
+import io.cdap.cdap.etl.api.relational.InvalidRelation;
+import io.cdap.cdap.etl.api.relational.Relation;
+import io.cdap.cdap.etl.api.relational.RelationalTranformContext;
 import io.cdap.wrangler.api.Arguments;
 import io.cdap.wrangler.api.Directive;
 import io.cdap.wrangler.api.DirectiveExecutionException;
@@ -32,6 +36,7 @@ import io.cdap.wrangler.api.lineage.Mutation;
 import io.cdap.wrangler.api.parser.ColumnNameList;
 import io.cdap.wrangler.api.parser.TokenType;
 import io.cdap.wrangler.api.parser.UsageDefinition;
+import io.cdap.wrangler.utils.SqlExpressionGenerator;
 
 import java.util.ArrayList;
 import java.util.Arrays;
@@ -100,5 +105,21 @@ public class CreateRecord implements Directive, Lineage {
       .readable("Created column based on values in columns '%s''", Arrays.asList(columns))
       .relation(Many.columns(columns), targetColumn)
       .build();
+  }
+
+  @Override
+  public Relation transform(RelationalTranformContext relationalTranformContext,
+                            Relation relation) {
+    java.util.Optional<ExpressionFactory<String>> expressionFactory = SqlExpressionGenerator
+            .getExpressionFactory(relationalTranformContext);
+    if (!expressionFactory.isPresent()) {
+      return new InvalidRelation("Cannot find an Expression Factory");
+    }
+    return relation.setColumn(targetColumn, expressionFactory.get().compile(String
+            .format("struct(%s)", getColumnString(columns))));
+  }
+
+  public String getColumnString(String[] columns) {
+      return String.join(",", columns);
   }
 }
--- a/wrangler-core/src/main/java/io/cdap/directives/column/CreateRecord.java
+++ b/wrangler-core/src/main/java/io/cdap/directives/column/CreateRecord.java
@@ -115,11 +115,9 @@ public class CreateRecord implements Directive, Lineage {
     if (!expressionFactory.isPresent()) {
       return new InvalidRelation("Cannot find an Expression Factory");
     }
+    String getColumnString = String.join(",", columns);
     return relation.setColumn(targetColumn, expressionFactory.get().compile(String
-            .format("struct(%s)", getColumnString(columns))));
+            .format("struct(%s)", getColumnString)));
   }
 
-  public String getColumnString(String[] columns) {
-      return String.join(",", columns);
-  }
 }
--- a/wrangler-core/src/main/java/io/cdap/directives/column/Swap.java
+++ b/wrangler-core/src/main/java/io/cdap/directives/column/Swap.java
@@ -19,6 +19,10 @@ package io.cdap.directives.column;
 import io.cdap.cdap.api.annotation.Description;
 import io.cdap.cdap.api.annotation.Name;
 import io.cdap.cdap.api.annotation.Plugin;
+import io.cdap.cdap.etl.api.relational.ExpressionFactory;
+import io.cdap.cdap.etl.api.relational.InvalidRelation;
+import io.cdap.cdap.etl.api.relational.Relation;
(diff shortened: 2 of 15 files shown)


wrangler-core/src/main/java/io/cdap/directives/transformation/SplitEmail.java
```diff
@@ -128,4 +135,24 @@ private Pair<String, String> extractDomainAndAccount(String emailId) {
...
+    if (!expressionFactory.isPresent()) {
+      return new InvalidRelation("Cannot find an Expression Factory");
+    }
+
+    Relation accountRelation = relation.setColumn(generatedAccountCol, getExpression(expressionFactory));
+    return accountRelation.setColumn(generatedDomainCol,
+            expressionFactory.get().compile(String.format("substring_index(%s, '@', -1)", column)));
+  }
+
+  Expression getExpression(Optional<ExpressionFactory<String>> expfactory) {
```
vanathi-g: Can just have two formatted strings with these expressions in the transform function itself, would be easier to understand

wrangler-core/src/main/java/io/cdap/directives/transformation/SplitEmail.java
```diff
@@ -128,4 +135,24 @@ private Pair<String, String> extractDomainAndAccount(String emailId) {
...
+    return expfactory.get().compile(String
+            .format("substring(%s, 1, char_length(%s) - locate('@', reverse(%s)))", column, column, column));
```
vanathi-g: Also, why not use the same substring_index(%s, '@', 1) to get everything before the @ symbol?

wrangler-core/src/main/java/io/cdap/directives/transformation/UrlEncode.java
```diff
@@ -101,4 +107,16 @@ public List<Row> execute(List<Row> rows, ExecutorContext context) throws Directi
...
+  public Relation transform(RelationalTranformContext relationalTranformContext,
+                            Relation relation) {
+    Optional<ExpressionFactory<String>> expressionFactory = SqlExpressionGenerator
+            .getExpressionFactory(relationalTranformContext);
+    if (!expressionFactory.isPresent()) {
+      return new InvalidRelation("Cannot find an Expression Factory");
+    }
+    return relation.setColumn(
+            column, expressionFactory.get().compile(
+                    String.format("reflect('java.net.url.Encoder', 'encode', %s, 'utf-8')", column)));
```
vanathi-g: Did you mean java.net.URLEncoder? (same thing for URLDecoder too, check once)

wrangler-core/src/main/java/io/cdap/directives/writer/WriteAsJsonObject.java
```diff
@@ -115,4 +120,20 @@ public Mutation lineage() {
...
+    java.util.Optional<ExpressionFactory<String>> expressionFactory = SqlExpressionGenerator
+            .getExpressionFactory(relationalTranformContext);
+    if (!expressionFactory.isPresent()) {
+      return new InvalidRelation("Cannot find an Expression Factory");
+    }
+    return relation.setColumn(
+            column, expressionFactory.get().compile(String.format("struct(%s)", getJSONColumns(columns))));
+  }
+
+  String getJSONColumns(List<String> columnList) {
```
vanathi-g: Unnecessary function can replace expression in String.format(), or consider renaming it to getJSONColumnList

vanathi-g (changes requested):

wrangler-core/src/main/java/io/cdap/directives/column/Swap.java
```diff
@@ -93,4 +99,19 @@ public Mutation lineage() {
...
+
+    Optional<ExpressionFactory<String>> expressionFactory = SqlExpressionGenerator
+            .getExpressionFactory(relationalTranformContext);
+    if (!expressionFactory.isPresent()) {
+      return new InvalidRelation("Cannot find an Expression Factory");
+    }
+
+    Relation tempRel =  relation.setColumn("tempColumn", expressionFactory.get().compile(right));
+    tempRel = tempRel.setColumn(right, expressionFactory.get().compile(left));
+    return tempRel.setColumn(left, expressionFactory.get().compile("tempColumn"));
```
saimukkamala, line 115: This adds a new column. Do we need to drop the temp column?

wrangler-core/src/main/java/io/cdap/directives/transformation/SplitEmail.java
shrverma: split-email directive returns the characters before the last '@' character, to have the split at the last occurrence of '@', I have used the above expression.

wrangler-core/src/main/java/io/cdap/directives/parser/FixedLengthParser.java
```diff
@@ -148,4 +153,32 @@ public Mutation lineage() {
...
+      return new InvalidRelation("Cannot find an Expression Factory");
+    }
+
+    int currentpos = 1;
+    int columncounter = 1;
+
+    for (int width : widths) {
+      relation = relation.setColumn(String.format("%s_%d", col, columncounter),
+              expressionFactory.get().compile(String
+                      .format("replace(substr(%s, %d, %d), '%s', \"\")"
```
saimukkamala: What happens when widths are greater than the length of the string?

wrangler-core/src/main/java/io/cdap/directives/column/Swap.java
shrverma: Dropping the temp column prevents the swap and returns both columns as is because the query plan somehow doesn't record the creation of the temp column.

wrangler-core/src/main/java/io/cdap/directives/parser/FixedLengthParser.java
shrverma: In that case the current directive behavior filters out that record. If the width is greater than the length of all the values in that column, no data is returned. To replicate the same behavior I have added the additional filter condition.

wrangler-core/src/main/java/io/cdap/directives/transformation/UrlEncode.java
shrverma: Yes, fixed it.

saimukkamala (approved):

wrangler-core/src/main/java/io/cdap/directives/column/CreateRecord.java
```diff
@@ -101,4 +106,20 @@ public Mutation lineage() {
...
+    java.util.Optional<ExpressionFactory<String>> expressionFactory = SqlExpressionGenerator
+            .getExpressionFactory(relationalTranformContext);
+    if (!expressionFactory.isPresent()) {
+      return new InvalidRelation("Cannot find an Expression Factory");
+    }
+    return relation.setColumn(targetColumn, expressionFactory.get().compile(String
+            .format("struct(%s)", getColumnString(columns))));
+  }
+
+  public String getColumnString(String[] columns) {
```
vanathi-g: Can just have a string variable instead of a function to make it easier to read. Similarly for WriteAsJsonObject (and other places where just String.join is used)

wrangler-core/src/main/java/io/cdap/directives/parser/FixedLengthParser.java
```diff
@@ -148,4 +153,32 @@ public Mutation lineage() {
...
+                              , col, currentpos, width, padding)))
+              .filter(expressionFactory.get().compile(String.format("%d <= (length(%s) - %d + 1)",
+                      width, col, currentpos)));
```
vanathi-g: Can move the `String.format`s to a variable to make it easier to read

vanathi-g (approved):

PR #654, opened 2023-07-26
[CDAP-15361] Add feature flag for schema management

vanathi-g:
## Changes 
- PR in CDAP: https://github.com/cdapio/cdap/pull/15256
- `ExecutorContext` now has a new field to check whether schema handling is enabled (set as true in `ServicePipelineContext` and `TestingPipelineContext`)
- The feature flag is also checked if enabled in 
    - `AbstractDirectiveHandler` - UI display of types (execution response)
    - `WorkspaceHandler` - generating plugin output schema in specification
- If the feature flag is disabled, fallback to old schema inference logic
- Also fixes CDAP-20740
### Testing
- Moved createUberRecord from AbstractDirectiveHandler to a new RowHelper utility class to ensure the same method is used consistently (including in tests)
- Refactored existing schema tests
- Added an additional test to check output schema inference when schema management is disabled

### DIFF
--- a/wrangler-core/src/main/java/io/cdap/wrangler/utils/RowHelper.java
+++ b/wrangler-core/src/main/java/io/cdap/wrangler/utils/RowHelper.java
@@ -1,0 +1,52 @@
+/*
+ * Copyright © 2023 Cask Data, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package io.cdap.wrangler.utils;
+
+import io.cdap.wrangler.api.Row;
+
+import java.util.List;
+
+/**
+ * Utility methods for {@link Row}
+ */
+public class RowHelper {
+  private RowHelper() {
+    throw new AssertionError("Cannot instantiate a static utility class");
+  }
+
+  /**
+   * Creates a merged record after iterating through all rows.
+   *
+   * @param rows list of all rows.
+   * @return A single record will rows merged across all columns.
+   */
+  public static Row createMergedRow(List<Row> rows) {
+    Row merged = new Row();
+    for (Row row : rows) {
+      for (int i = 0; i < row.width(); ++i) {
+        Object o = row.getValue(i);
+        if (o != null) {
+          int idx = merged.find(row.getColumn(i));
+          if (idx == -1) {
+            merged.add(row.getColumn(i), o);
+          }
+        }
+      }
+    }
+    return merged;
+  }
+}
--- a/wrangler-core/src/test/java/io/cdap/wrangler/TestingPipelineContext.java
+++ b/wrangler-core/src/test/java/io/cdap/wrangler/TestingPipelineContext.java
@@ -16,7 +16,6 @@
 
 package io.cdap.wrangler;
 
-import io.cdap.cdap.api.metrics.Metrics;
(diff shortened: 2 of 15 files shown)


wrangler-core/pom.xml
```diff
@@ -68,6 +68,11 @@
       <version>${cdap.version}</version>
       <scope>provided</scope>
     </dependency>
+    <dependency>
```
tivv: I don't think you use it in wrangler-core, do you?

tivv (commented):
Please also add/update tests

wrangler-core/pom.xml
vanathi-g: Was initially running into some circular dependency issues, resolved now

wrangler-core/src/main/java/io/cdap/wrangler/utils/RowHelper.java
```diff
@@ -0,0 +1,52 @@
...
+/**
+ * Utility methods for {@link Row}
+ */
+public class RowHelper {
+  private RowHelper() {
+    throw new AssertionError("Cannot instantiate a static utility class");
+  }
+
+  /**
+   * Creates a uber record after iterating through all rows.
```
tivv: nit: use "merged" instead of uber as it reflects the purpose more clear

wrangler-core/src/test/java/io/cdap/wrangler/TestingPipelineContext.java
```diff
@@ -142,4 +115,41 @@ public TransientStore getTransientStore() {
   public <T> Lookup<T> provide(String s, Map<String, String> map) {
     return null;
   }
+
+  @Override
+  public boolean schemaManagementEnabled() {
+    return schemaManagementEnabled;
+  }
+
+  protected class MockStageMetrics implements StageMetrics {
```
tivv: nit: Any reason not to use `Mockito`? Much less boilerplate code

wrangler-core/src/test/java/io/cdap/wrangler/TestingPipelineContext.java
```diff
@@ -35,46 +35,19 @@
...
-      }
+    this.schemaManagementEnabled = false;
+  }
 
-      @Override
-      public Map<String, String> getTags() {
-        return Collections.emptyMap();
-      }
-    };
+  public TestingPipelineContext(boolean schemaManagementEnabled) {
```
tivv: nit: consider using setter. `new TestingPipelineContext().setSchemaManagementEnabled(true)` is more clear when reading code. Boolean values are generally not very readable

wrangler-core/src/test/java/io/cdap/wrangler/TestingPipelineContext.java
```diff
@@ -142,4 +115,41 @@ public TransientStore getTransientStore() {
   public <T> Lookup<T> provide(String s, Map<String, String> map) {
     return null;
   }
+
+  @Override
+  public boolean schemaManagementEnabled() {
```
tivv: Please follow general JavaBean notation (`isSchemaManagementEnabled`)

wrangler-core/src/main/java/io/cdap/wrangler/utils/RowHelper.java
vanathi-g: Done

wrangler-core/src/test/java/io/cdap/wrangler/TestingPipelineContext.java
vanathi-g: Yeah that makes sense. I'd just moved the original code from the constructor to it's own class but I've used mockito now

wrangler-core/src/test/java/io/cdap/wrangler/TestingPipelineContext.java
vanathi-g: Understood, done

wrangler-core/src/test/java/io/cdap/wrangler/TestingPipelineContext.java
vanathi-g: Done

tivv (approved):

PR #655, opened 2023-07-27
Partially supported directives

shrverma:

- Change-column-case-names
- Quantization
- Cleanse-column-names
- Set-headers
- Set-type
- Diff-date
- Format-date
- Set-charset
- Fail
- Record-condition-filter
- Column-expression
- Decode
- Encode
- Write-as-CSV
- Write-as-JSON-Map

### DIFF
--- a/wrangler-core/src/main/java/io/cdap/directives/column/CleanseColumnNames.java
+++ b/wrangler-core/src/main/java/io/cdap/directives/column/CleanseColumnNames.java
@@ -19,6 +19,11 @@ package io.cdap.directives.column;
 import io.cdap.cdap.api.annotation.Description;
 import io.cdap.cdap.api.annotation.Name;
 import io.cdap.cdap.api.annotation.Plugin;
+import io.cdap.cdap.etl.api.relational.Expression;
+import io.cdap.cdap.etl.api.relational.ExpressionFactory;
+import io.cdap.cdap.etl.api.relational.InvalidRelation;
+import io.cdap.cdap.etl.api.relational.Relation;
+import io.cdap.cdap.etl.api.relational.RelationalTranformContext;
 import io.cdap.wrangler.api.Arguments;
 import io.cdap.wrangler.api.Directive;
 import io.cdap.wrangler.api.DirectiveExecutionException;
@@ -30,8 +35,12 @@ import io.cdap.wrangler.api.lineage.Lineage;
 import io.cdap.wrangler.api.lineage.Many;
 import io.cdap.wrangler.api.lineage.Mutation;
 import io.cdap.wrangler.api.parser.UsageDefinition;
+import io.cdap.wrangler.utils.SqlExpressionGenerator;
 
+import java.util.Collection;
+import java.util.LinkedHashMap;
 import java.util.List;
+import java.util.Map;
 
 /**
  * A directive for cleanses columns names.
@@ -93,4 +102,30 @@ public final class CleanseColumnNames implements Directive, Lineage {
       .all(Many.of())
       .build();
   }
+
+  @Override
+  public Relation transform(RelationalTranformContext relationalTranformContext,
+                            Relation relation) {
+    java.util.Optional<ExpressionFactory<String>> expressionFactory = SqlExpressionGenerator
+            .getExpressionFactory(relationalTranformContext);
+    if (!expressionFactory.isPresent()) {
+      return new InvalidRelation("Cannot find an Expression Factory");
+    }
+    List<String> columnNames = SqlExpressionGenerator.generateListCols(relationalTranformContext);
+    Map<String, Expression> colmap = generateCleanseColumnMap(columnNames, expressionFactory.get());
+    return relation.select(colmap);
+  }
+
+  public static Map<String, Expression> generateCleanseColumnMap(Collection columns,
+                                                                 ExpressionFactory<String> factory) {
+    Map<String, Expression> columnExpMap = new LinkedHashMap<>();
+    columns.forEach((colName)-> columnExpMap.put(String
+            .format(colName
+                    .toString()
+                    .toLowerCase()
+                    .replaceAll("[^a-zA-Z0-9_]", "_")), factory
+            .compile(colName.toString())));
+    return columnExpMap;
+  }
+
 }
--- a/wrangler-core/src/main/java/io/cdap/directives/column/CleanseColumnNames.java
+++ b/wrangler-core/src/main/java/io/cdap/directives/column/CleanseColumnNames.java
@@ -128,4 +128,9 @@ public final class CleanseColumnNames implements Directive, Lineage {
     return columnExpMap;
   }
 
+  @Override
+  public boolean isSQLSupported() {
+    return true;
+  }
+
 }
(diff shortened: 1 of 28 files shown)


wrangler-core/src/main/java/io/cdap/wrangler/utils/SqlExpressionGenerator.java
```diff
@@ -105,4 +109,19 @@ public static String getColumnTypeExp(String toType, String column, @Nullable In
                 return column;
         }
     }
+
+    public static List<String> generateListCols(RelationalTranformContext relationalTranformContext) {
```
vanathi-g: Can rename to generateColumnList or generateColumnNameList

wrangler-core/src/main/java/io/cdap/directives/column/SetHeader.java
```diff
@@ -90,5 +101,27 @@ public Mutation lineage() {
...
+    List<String> columnNames = SqlExpressionGenerator.generateListCols(relationalTranformContext);
+    Map<String, Expression> columnExpMap = new LinkedHashMap<>();
+
+    IntStream.range(0, Math.min(columnNames.size(), columns.size()))
+            .forEach(i -> columnExpMap.put(columns.get(i), expressionFactory.get().compile(columnNames.get(i))));
+
+    if (columnNames.size() > columns.size()) {
+      IntStream.range(columns.size(), columnNames.size())
+          .forEach(i -> columnExpMap.put(columnNames.get(i), expressionFactory.get().compile(columnNames.get(i))));
+    }
```
vanathi-g: can move to a function with a simple for loop

wrangler-core/src/main/java/io/cdap/directives/column/CleanseColumnNames.java
```diff
@@ -93,4 +102,30 @@ public Mutation lineage() {
...
+            .getExpressionFactory(relationalTranformContext);
+    if (!expressionFactory.isPresent()) {
+      return new InvalidRelation("Cannot find an Expression Factory");
+    }
+    List<String> columnNames = SqlExpressionGenerator.generateListCols(relationalTranformContext);
+    Map<String, Expression> colmap = generateCleanseColumnMap(columnNames, expressionFactory.get());
+    return relation.select(colmap);
+  }
+
+  public static Map<String, Expression> generateCleanseColumnMap(Collection columns,
```
vanathi-g: make private

wrangler-core/src/main/java/io/cdap/directives/column/CleanseColumnNames.java
```diff
@@ -93,4 +102,30 @@ public Mutation lineage() {
...
+
+  @Override
+  public Relation transform(RelationalTranformContext relationalTranformContext,
+                            Relation relation) {
+    java.util.Optional<ExpressionFactory<String>> expressionFactory = SqlExpressionGenerator
+            .getExpressionFactory(relationalTranformContext);
+    if (!expressionFactory.isPresent()) {
+      return new InvalidRelation("Cannot find an Expression Factory");
+    }
+    List<String> columnNames = SqlExpressionGenerator.generateListCols(relationalTranformContext);
```
vanathi-g: So this method gets the list of input columns from the transform context, which is not really related to "generating a SQL expression". I would either suggest keeping this in a different util class or you can just rename the method to getColumnList

wrangler-core/src/main/java/io/cdap/directives/column/SetType.java
```diff
@@ -108,4 +113,17 @@ public Mutation lineage() {
...
+
+  @Override
+  public Relation transform(RelationalTranformContext relationalTranformContext,
+                            Relation relation) {
+    java.util.Optional<ExpressionFactory<String>> expressionFactory = SqlExpressionGenerator
+            .getExpressionFactory(relationalTranformContext);
+    if (!expressionFactory.isPresent()) {
+      return new InvalidRelation("Cannot find an Expression Factory");
+    }
+    String expression = SqlExpressionGenerator.getColumnTypeExp(type, col, scale);
```
vanathi-g, line 128: Some of the directive specific methods are in their respective classes and some are in the SqlExpressionGenerator. We should try to maintain consistency

wrangler-core/src/main/java/io/cdap/directives/row/Fail.java
```diff
@@ -116,4 +121,20 @@ public List<EntityCountMetric> getCountMetrics() {
...
+  public Relation transform(RelationalTranformContext relationalTranformContext,
+                            Relation relation) {
+    java.util.Optional<ExpressionFactory<String>> expressionFactory = SqlExpressionGenerator
+            .getExpressionFactory(relationalTranformContext);
+    if (!expressionFactory.isPresent()) {
+      return new InvalidRelation("Cannot find an Expression Factory");
+    }
+
+    return relation.setColumn("tempColumn", expressionFactory.get().compile(
+            String.format("if(%s, raise_error(\"Condition '%s' evaluated to true. " +
```
vanathi-g: Can move this to a variable

wrangler-core/src/main/java/io/cdap/directives/row/RecordConditionFilter.java
```diff
@@ -129,4 +134,16 @@ public List<EntityCountMetric> getCountMetrics() {
...
+
+  @Override
+  public Relation transform(RelationalTranformContext relationalTranformContext,
+                            Relation relation) {
+    java.util.Optional<ExpressionFactory<String>> expressionFactory = SqlExpressionGenerator
+            .getExpressionFactory(relationalTranformContext);
+    if (!expressionFactory.isPresent()) {
+      return new InvalidRelation("Cannot find an Expression Factory");
+    }
+    return relation.filter(expressionFactory.get().compile(el.getScriptParsedText()));
```
vanathi-g, line 148: Add a code comment/TODO stating that potentially some JEXL expressions (not supported in SQL) can break this. Same in ColumnExpression and other places

wrangler-core/src/main/java/io/cdap/directives/transformation/Decode.java
```diff
@@ -155,4 +161,27 @@ public Mutation lineage() {
       .relation(column, column)
       .build();
   }
+
+  @Override
+  public Relation transform(RelationalTranformContext relationalTranformContext,
+                            Relation relation) {
+    Optional<ExpressionFactory<String>> expressionFactory = SqlExpressionGenerator
+            .getExpressionFactory(relationalTranformContext);
+    if (!expressionFactory.isPresent()) {
```
vanathi-g, line 169: Switch case, construct necessary string  then call relation.setColumn(column name, constructed string). Also, no base32 support?

wrangler-core/src/main/java/io/cdap/directives/transformation/Encode.java
```diff
@@ -150,4 +156,27 @@ public Mutation lineage() {
...
+
+  @Override
+  public Relation transform(RelationalTranformContext relationalTranformContext,
+                            Relation relation) {
+    Optional<ExpressionFactory<String>> expressionFactory = SqlExpressionGenerator
+            .getExpressionFactory(relationalTranformContext);
+    if (!expressionFactory.isPresent()) {
+      return new InvalidRelation("Cannot find an Expression Factory");
+    }
+    if (method.toString().equalsIgnoreCase("base64")) {
```
vanathi-g: Same as decode changes

vanathi-g (changes requested):

PR #657, opened 2023-08-23
Add custom schema handling logic to directives that rename columns

vanathi-g:
## Changes
These directives perform a "rename"-like operation in the following ways:
- `columns-replace` modifies column names in bulk based on a sed expression
- `change-column-case` changes the case of column names
- `copy`: the copied data has the same schema as the original data

Hence, the `getOutputSchema` method is overriden with custom logic to carry over the original schema to the renamed/copied columns

### DIFF
--- a/wrangler-core/src/main/java/io/cdap/directives/column/Copy.java
+++ b/wrangler-core/src/main/java/io/cdap/directives/column/Copy.java
@@ -19,6 +19,7 @@ package io.cdap.directives.column;
 import io.cdap.cdap.api.annotation.Description;
 import io.cdap.cdap.api.annotation.Name;
 import io.cdap.cdap.api.annotation.Plugin;
+import io.cdap.cdap.api.data.schema.Schema;
 import io.cdap.wrangler.api.Arguments;
 import io.cdap.wrangler.api.Directive;
 import io.cdap.wrangler.api.DirectiveExecutionException;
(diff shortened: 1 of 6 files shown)


wrangler-core/src/main/java/io/cdap/directives/column/Copy.java
```diff
@@ -110,4 +112,18 @@ public Mutation lineage() {
...
   }
+
+  @Override
+  public Schema getOutputSchema(SchemaResolutionContext context) {
+    Schema inputSchema = context.getInputSchema();
+    return Schema.recordOf(
+      "outputSchema",
+      inputSchema.getFields().stream()
+        .map(
+          field -> field.getName().equals(destination.value()) ?
```
saimukkamala: If the destination column does not exist in the input schema, new column is not added to output schema.

wrangler-core/src/test/java/io/cdap/directives/column/CopyTest.java
```diff
@@ -96,4 +99,33 @@ public void testForceCopy() throws Exception {
     Assert.assertEquals(rows.get(2).getValue("body_2"), rows.get(2).getValue("body_1"));
   }
 
+  @Test
+  public void testGetOutputSchemaForCopiedColumn() throws Exception {
+    String[] directives = new String[] {
+      "copy :col_B :col_A true",
```
saimukkamala, line 105: can we also add a test where the destination column is not present in input schema

saimukkamala (approved):

PR #658, opened 2023-08-24
Add custom schema handling logic to SetType

vanathi-g:
## Changes
- Created a schema type map for handling type conversion in set-type directive
- Future work:
    - Add precision as a directive argument and support setting precision as well

### DIFF
--- a/wrangler-core/src/main/java/io/cdap/wrangler/utils/ColumnConverter.java
+++ b/wrangler-core/src/main/java/io/cdap/wrangler/utils/ColumnConverter.java
@@ -83,7 +83,7 @@ public final class ColumnConverter {
       }
       try {
         Object converted = ColumnConverter.convertType(column, toType, object);
-        if (toType.equalsIgnoreCase("DECIMAL")) {
+        if (toType.equalsIgnoreCase(ColumnTypeNames.DECIMAL)) {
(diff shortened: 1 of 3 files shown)


wrangler-core/src/main/java/io/cdap/wrangler/utils/ColumnConverter.java
```diff
@@ -305,4 +311,38 @@ private static BigDecimal setDecimalScale(BigDecimal decimal, Integer scale, Rou
...
+          "Supported types are: int, short, long, double, decimal, boolean, string, bytes", type));
+      }
+      typeSchema = Schema.nullableOf(Schema.of(SCHEMA_TYPE_MAP.get(type)));
+    }
+    return typeSchema;
+  }
+
+  static {
+    Map<String, Schema.Type> schemaTypeMap = new HashMap<>();
+    schemaTypeMap.put("INTEGER", Schema.Type.INT);
```
saimukkamala: probably its a good idea to refactor String `Type` into constants. These are used across this file.

saimukkamala (approved):

wrangler-core/src/main/java/io/cdap/wrangler/utils/ColumnConverter.java
vanathi-g: Done

PR #670, opened 2023-09-25
[CDAP-20839] Fix grammar to identify and tokenize negative numbers

vanathi-g:
## Issue:
Wrangler's grammar does not recognize negative numbers, ex. unable to use negative number while incrementing a variable:

image

## Fix:
- Add '-' to the INT token
- Add test for negative scale in set-type to decimal directive

image


GnsP (approved):

vanathi-g:
Merging this since the e2e test build fails at "Run required e2e tests". This happens because there is no test annotated with Wrangler_required as of now. All other build and unit test steps have passed.

PR #673, opened 2023-10-05
[CDAP-15361] Fix set-type directive for decimal when scale is null

minurajeeve:
Fix set-type directive for decimal by adding a default value for scale in case it is null/not provided.

### DIFF
--- a/wrangler-core/src/main/java/io/cdap/wrangler/utils/ColumnConverter.java
+++ b/wrangler-core/src/main/java/io/cdap/wrangler/utils/ColumnConverter.java
@@ -317,7 +317,8 @@ public final class ColumnConverter {
     type = type.toUpperCase();
     if (type.equals(ColumnTypeNames.DECIMAL)) {
       // TODO make set-type support setting decimal precision
-      typeSchema = Schema.nullableOf(Schema.decimalOf(38, scale));
+      scale = scale != null ? scale : 38;
+      typeSchema = Schema.nullableOf(Schema.decimalOf(76, scale));
     } else {
       if (!SCHEMA_TYPE_MAP.containsKey(type)) {
         throw new DirectiveParseException(String.format("'%s' is an unsupported type. " +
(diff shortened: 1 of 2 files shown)


wrangler-core/src/main/java/io/cdap/wrangler/utils/ColumnConverter.java
```diff
@@ -317,7 +317,8 @@ public static Schema getSchemaForType(String type, Integer scale) throws Directi
     type = type.toUpperCase();
     if (type.equals(ColumnTypeNames.DECIMAL)) {
       // TODO make set-type support setting decimal precision
-      typeSchema = Schema.nullableOf(Schema.decimalOf(38, scale));
+      scale = scale != null ? scale : 38;
+      typeSchema = Schema.nullableOf(Schema.decimalOf(76, scale));
```
vanathi-g: I think we can keep precision as 77 because according to the docs, a max precision of 76.76 is allowed. Besides, the other plugins (like Oracle) follow 38, 77 as default scale, precision.

wrangler-core/src/test/java/io/cdap/directives/column/SetTypeTest.java
```diff
@@ -213,6 +213,17 @@ public void testToDecimalInvalidRoundingMode() throws Exception {
     TestingRig.execute(directives, rows);
   }
 
+  @Test
```
vanathi-g, line 216: This test should also check whether output schema is generated correctly with default scale and precision.

minurajeeve: Done

wrangler-core/src/main/java/io/cdap/wrangler/utils/ColumnConverter.java
minurajeeve: Done

saimukkamala (approved):

PR #678, opened 2023-11-09
[CDAP-20865] Fix copy directive schema generation bug

vanathi-g:
Fixes bug in existing copy directive's output schema generation when 'force' argument is true but a new column is specified.

### DIFF
--- a/wrangler-core/src/main/java/io/cdap/directives/column/Copy.java
+++ b/wrangler-core/src/main/java/io/cdap/directives/column/Copy.java
@@ -118,18 +118,18 @@ public class Copy implements Directive, Lineage {
     Schema inputSchema = context.getInputSchema();
     List<Schema.Field> outputFields = new ArrayList<>();
     Schema sourceSchema = inputSchema.getField(source.value()).getSchema();
+    boolean destinationExists = inputSchema.getField(destination.value()) != null;
 
     for (Schema.Field field : inputSchema.getFields()) {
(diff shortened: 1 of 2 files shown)


wrangler-core/src/main/java/io/cdap/directives/column/Copy.java
```diff
@@ -118,18 +118,23 @@ public Schema getOutputSchema(SchemaResolutionContext context) {
...
     List<Schema.Field> outputFields = new ArrayList<>();
     Schema sourceSchema = inputSchema.getField(source.value()).getSchema();
+    boolean added = false;
 
     for (Schema.Field field : inputSchema.getFields()) {
       if (field.getName().equals(destination.value())) {
-        outputFields.add(Schema.Field.of(destination.value(), sourceSchema));
+        if (force) {
+          outputFields.add(Schema.Field.of(destination.value(), sourceSchema));
+          added = true;
```
saimukkamala: I would rather name this variable `found` than `added`. and move this statement outside `if (force)` section. Currently we are throwing an error in `else` statement (if force is false). In case this changes in future and the code continues to execute a new column will not be added since `added` will never be true. I suggest we change `added` to `force`.

vanathi-g: Changed it to check if destination field exists in schema, if not, add the column in the end

saimukkamala (approved):

PR #686, opened 2023-11-27
modifying execute API to get column nullability state

minurajeeve:
Wrangler will now have two types of columns : Nullable and Non-nullable. This PR handles all the code changes that are required to set the column as Nullable or Non-nullable (column Nullability state).


wrangler-proto/src/main/java/io/cdap/wrangler/proto/workspace/v2/DirectiveExecutionRequest.java
harshdeeppruthi, line 1: Since we are ultimately setting the nullability mapping in Workspace (and it seems like a workspace property), can we instead just keep it as an attribute in workspace alone without modifying this file and update it on Workspace level when changed. Or is there a specific reason for keeping it in DirectiveExecutionRequest and then setting it in Workspace ?

wrangler-proto/src/main/java/io/cdap/wrangler/proto/workspace/v2/Workspace.java
```diff
@@ -79,6 +83,15 @@ public JsonObject getInsights() {
     return insights;
   }
 
+  public HashMap<String, UserDefinedAction> getColumnMappings() {
```
harshdeeppruthi: Can we keep the same name here too or make it more intuitive that column mappings refer to the nullability map

wrangler-proto/src/main/java/io/cdap/wrangler/proto/workspace/v2/Workspace.java
```diff
@@ -164,4 +177,15 @@ public Workspace build() {
...
+
+  /**
+   * UserDefinedAction enum.
+   */
+  public enum UserDefinedAction {
+    NO_ACTION,
+    SKIP_ROW,
+    SEND_TO_ERROR_COLLECTOR,
+    ERROR_PIPELINE,
+    NULLABLE
```
harshdeeppruthi: How would Nullable be different from No_Action ?

wrangler-service/src/main/java/io/cdap/wrangler/service/directive/AbstractDirectiveHandler.java
```diff
@@ -118,7 +121,8 @@ protected <E extends Exception> List<Row> executeDirectives(
       String namespace,
       List<String> directives,
       List<Row> sample,
-      GrammarWalker.Visitor<E> grammarVisitor) throws DirectiveParseException, E, RecipeException {
+      GrammarWalker.Visitor<E> grammarVisitor,
+      Workspace workspace) throws DirectiveParseException, E, RecipeException {
```
harshdeeppruthi: Do we need to add whole workspace as an argument. Can we use just nullabilityMap instead as the argument ?

wrangler-service/src/main/java/io/cdap/wrangler/service/directive/RemoteExecutionTask.java
```diff
@@ -107,7 +107,7 @@ public void run(RunnableTaskContext runnableTaskContext) throws Exception {
                                                                           namespace,
                                                                           ExecutorContext.Environment.SERVICE,
                                                                           systemAppContext,
-                                                                          new DefaultTransientStore()))) {
+                                                                          new DefaultTransientStore()), null)) {
```
harshdeeppruthi, line 110: Might be a good idea to set it to a default Hashmap everywhere instead of null where each column probably maps to NO_ACTION. I think it would be easier in later implementations.

wrangler-proto/src/main/java/io/cdap/wrangler/proto/workspace/v2/DirectiveExecutionRequest.java
minurajeeve: DirectiveExecutionRequest will be used to get the nullabilityMap set by the user in the UI. This is the request message of execute API used to convey the nullability map state to Workspace.

wrangler-proto/src/main/java/io/cdap/wrangler/proto/workspace/v2/Workspace.java
minurajeeve: Thanks for pointing that out. I missed that.

wrangler-proto/src/main/java/io/cdap/wrangler/proto/workspace/v2/Workspace.java
minurajeeve: It is not different, both these options will return the row as it is. I have kept a NULLABLE option for columns that were non- nullable earlier and were converted to Nullable column later. I might remove this option later if not required.

wrangler-service/src/main/java/io/cdap/wrangler/service/directive/AbstractDirectiveHandler.java
minurajeeve: I need the workspace object, for the subsequent changes

wrangler-proto/src/main/java/io/cdap/wrangler/proto/workspace/v2/DirectiveExecutionRequest.java
harshdeeppruthi: I meant can we directly do it in Workspace related workflow. Not sure about this but probably doing this as a part of ```workspaceUpdate``` of ```WorkspaceHandler.java``` might be a better option ? Because nullability won't change everytime a change is made to directive list. But this flow would set the nullabilityMap each time there is a change in directive list. It would be redundant in such a case to set the same value everytime

wrangler-service/src/main/java/io/cdap/wrangler/service/directive/RemoteExecutionTask.java
minurajeeve: if nullability map is null or empty it means all columns are nullable. So it's not required to set a default map.

wrangler-proto/src/main/java/io/cdap/wrangler/proto/workspace/v2/DirectiveExecutionRequest.java
minurajeeve: That's a good suggestion I'll check if we can do it this way.

minurajeeve: Unlike the execute API the updateWorkspace API does not have a response. There could be a case where the user has only set the column nullability and not directives. The user will expect a change in the Wrangler workspace UI's sample rows in case of null values(not implemented in this PR). This will not be possible with the update workspace API.

minurajeeve: Also every time nullabilityMap is changed the execute API will have to be called as the directives will have to be re-executed from the start with the new nullabilityMap. If we are using updateWorkspace we will have to call execute API whenever we are updating NullabilityMap.

wrangler-service/src/main/java/io/cdap/wrangler/service/directive/AbstractDirectiveHandler.java
minurajeeve: Removing the workspace object as it's not needed in subsequent changes.

wrangler-api/src/main/java/io/cdap/wrangler/api/NullHandlingException.java
```diff
@@ -0,0 +1,31 @@
+/*
+ * Copyright © 2016-2019 Cask Data, Inc.
```
albertshau: 2024

wrangler-api/src/main/java/io/cdap/wrangler/api/NullHandlingException.java
```diff
@@ -0,0 +1,31 @@
...
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package io.cdap.wrangler.api;
+
+/**
+ * A Null Handling specific exception used for communicating issues with Null Handling in a column.
+ */
+public class NullHandlingException extends Exception {
```
albertshau: I don't see this being thrown or caught anywhere, how are we expecting to use it? Would be better to leave this out of the PR and include it in whatever PR actually uses it.

wrangler-core/src/main/java/io/cdap/wrangler/executor/RecipePipelineExecutor.java
```diff
@@ -56,10 +59,13 @@ public final class RecipePipelineExecutor implements RecipePipeline<Row, Structu
   private final RecipeParser recipeParser;
   private final ExecutorContext context;
   private List<Directive> directives;
+  private HashMap<String, UserDefinedAction> nullabilityMap;
```
albertshau: HashMap -> Map. In general, when declaring variables we should be using the interface and not the specific implementation.

wrangler-core/src/main/java/io/cdap/wrangler/executor/RecipePipelineExecutor.java
```diff
@@ -56,10 +59,13 @@ public final class RecipePipelineExecutor implements RecipePipeline<Row, Structu
...
 
-  public RecipePipelineExecutor(RecipeParser recipeParser, @Nullable ExecutorContext context) {
+  public RecipePipelineExecutor(RecipeParser recipeParser, @Nullable ExecutorContext context,
+      HashMap<String, UserDefinedAction> nullabilityMap) {
```
albertshau: HashMap -> Map

wrangler-core/src/main/java/io/cdap/wrangler/executor/RecipePipelineExecutor.java
albertshau: this should also be final. Basically anything passed into the constructor should be final

wrangler-core/src/main/java/io/cdap/wrangler/executor/RecipePipelineExecutor.java
```diff
@@ -56,10 +59,13 @@ public final class RecipePipelineExecutor implements RecipePipeline<Row, Structu
...
     this.context = context;
     this.recipeParser = recipeParser;
+    this.nullabilityMap = nullabilityMap;
```
albertshau: should do a defensive copy to make sure the map cannot change.

wrangler-core/src/test/java/io/cdap/wrangler/TestingRig.java
```diff
@@ -89,7 +89,7 @@ public static List<Row> execute(String[] recipe, List<Row> rows, ExecutorContext
 
     String migrate = new MigrateToV2(recipe).migrate();
     RecipeParser parser = new GrammarBasedParser(Contexts.SYSTEM, migrate, registry);
-    return new RecipePipelineExecutor(parser, context).execute(rows);
+    return new RecipePipelineExecutor(parser, context, null).execute(rows);
```
albertshau: should not pass null for anything that is not annotated as `@Nullable`. Also should normally not pass null for any collection and instead pass an empty collection.

wrangler-proto/src/main/java/io/cdap/wrangler/proto/workspace/v2/DirectiveExecutionRequest.java
```diff
@@ -39,4 +45,8 @@ public int getLimit() {
   public List<String> getDirectives() {
     return directives == null ? Collections.emptyList() : directives;
   }
+
+  public HashMap<String, UserDefinedAction> getNullabilityMap() {
```
albertshau: HashMap -> Map

wrangler-proto/src/main/java/io/cdap/wrangler/proto/workspace/v2/DirectiveExecutionRequest.java
```diff
@@ -26,10 +28,14 @@
 public class DirectiveExecutionRequest {
   private final List<String> directives;
   private final int limit;
+  private final HashMap<String, UserDefinedAction> nullabilityMap;
 
-  public DirectiveExecutionRequest(List<String> directives, int limit) {
+
+  public DirectiveExecutionRequest(List<String> directives, int limit,
+      HashMap<String, UserDefinedAction> nullabilityMap) {
```
albertshau: HashMap -> Map

wrangler-proto/src/main/java/io/cdap/wrangler/proto/workspace/v2/DirectiveExecutionRequest.java
```diff
@@ -39,4 +45,8 @@ public int getLimit() {
   public List<String> getDirectives() {
     return directives == null ? Collections.emptyList() : directives;
   }
+
+  public HashMap<String, UserDefinedAction> getNullabilityMap() {
+    return nullabilityMap;
```
albertshau: need to check for null here, similar to getDirectives(). This object is created by deserializing the HTTP request body, so it's possible the caller is not setting this field.

wrangler-proto/src/main/java/io/cdap/wrangler/proto/workspace/v2/DirectiveExecutionRequest.java
```diff
@@ -26,10 +28,14 @@
 public class DirectiveExecutionRequest {
   private final List<String> directives;
   private final int limit;
+  private final HashMap<String, UserDefinedAction> nullabilityMap;
```
albertshau: This is very similar to using send-to-error and filter directives. Would like to understand why we need to introduce a new way to do almost the same thing.

wrangler-proto/src/main/java/io/cdap/wrangler/proto/workspace/v2/Workspace.java
```diff
@@ -79,6 +84,15 @@ public JsonObject getInsights() {
     return insights;
   }
 
+  public HashMap<String, UserDefinedAction> getNullabilityMap() {
+    return nullabilityMap;
+  }
+
+  public void setNullabilityMap(
```
albertshau: should usually not have any setter methods. Making the state mutable makes it very difficult to reason about correctness. If it needs to be changed, a new instance of the Workspace should be created using the Builder.

wrangler-proto/src/main/java/io/cdap/wrangler/proto/workspace/v2/Workspace.java
```diff
@@ -38,16 +39,20 @@ public class Workspace {
   // this is for insights page in UI
   private final JsonObject insights;
 
+  private HashMap<String, UserDefinedAction> nullabilityMap;
```
albertshau: HashMap -> Map (same for the rest of the class)

albertshau: this should also be final

wrangler-proto/src/main/java/io/cdap/wrangler/proto/workspace/v2/Workspace.java
```diff
@@ -38,16 +39,20 @@ public class Workspace {
...
-                    JsonObject insights) {
+                    JsonObject insights, HashMap<String, UserDefinedAction> nullabilityMap) {
     this.workspaceName = workspaceName;
     this.workspaceId = workspaceId;
     this.directives = directives;
     this.createdTimeMillis = createdTimeMillis;
     this.updatedTimeMillis = updatedTimeMillis;
     this.sampleSpec = sampleSpec;
     this.insights = insights;
+    this.nullabilityMap = nullabilityMap == null || nullabilityMap.isEmpty() ?
```
albertshau: by convention, only variables annotated as `@Nullable` should ever be null. Instead, should be creating an unmodifiable defensive copy here:

```
this.nullabilityMap = Collections.unmodifiableMap(new HashMap(nullabilityMap));
```

This is to prevent the map from ever changing once the object is created.

wrangler-proto/src/main/java/io/cdap/wrangler/proto/workspace/v2/Workspace.java
```diff
@@ -159,9 +175,23 @@ public Builder setInsights(JsonObject insights) {
...
       return new Workspace(workspaceName, workspaceId, directives, createdTimeMillis, updatedTimeMillis, sampleSpec,
-                           insights);
+                           insights, nullabilityMap);
     }
   }
+
+  /**
+   * UserDefinedAction enum.
+   */
+  public enum UserDefinedAction {
```
albertshau: should move this to its own class instead of being nested.

wrangler-proto/src/main/java/io/cdap/wrangler/proto/workspace/v2/Workspace.java
```diff
@@ -159,9 +175,23 @@ public Builder setInsights(JsonObject insights) {
...
+    SKIP_ROW,
+    SEND_TO_ERROR_COLLECTOR,
+    ERROR_PIPELINE,
```
albertshau: pipeline is not a great name, as wrangler can be used outside of a pipeline when making HTTP requests. This can just be ERROR

wrangler-proto/src/main/java/io/cdap/wrangler/proto/workspace/v2/Workspace.java
```diff
@@ -159,9 +175,23 @@ public Builder setInsights(JsonObject insights) {
...
+                           insights, nullabilityMap);
     }
   }
+
+  /**
+   * UserDefinedAction enum.
+   */
+  public enum UserDefinedAction {
+    SKIP_ROW,
+    SEND_TO_ERROR_COLLECTOR,
```
albertshau: ERROR_COLLECTOR is a pipeline specific concept, this can just be SEND_TO_ERROR

wrangler-proto/src/main/java/io/cdap/wrangler/proto/workspace/v2/Workspace.java
```diff
@@ -159,9 +175,23 @@ public Builder setInsights(JsonObject insights) {
...
-                           insights);
+                           insights, nullabilityMap);
     }
   }
+
+  /**
+   * UserDefinedAction enum.
+   */
+  public enum UserDefinedAction {
+    SKIP_ROW,
```
albertshau: a better name would be FILTER

wrangler-service/src/main/java/io/cdap/wrangler/service/directive/WorkspaceHandler.java
```diff
@@ -472,6 +474,12 @@ private DirectiveExecutionResponse execute(NamespaceSummary ns, HttpServiceReque
 
     WorkspaceDetail detail = wsStore.getWorkspaceDetail(workspaceId);
     UserDirectivesCollector userDirectivesCollector = new UserDirectivesCollector();
+    HashMap<String, UserDefinedAction> nullabilityMap = executionRequest.getNullabilityMap() == null ?
```
albertshau: this logic should be in the getNullabilityMap() class

wrangler-service/src/main/java/io/cdap/wrangler/service/directive/WorkspaceHandler.java
```diff
@@ -484,6 +492,18 @@ private DirectiveExecutionResponse execute(NamespaceSummary ns, HttpServiceReque
...
+  private void changeNullability(HashMap<String, UserDefinedAction> columnMappings,
+       WorkspaceId workspaceId) throws Exception {
+    try {
+      Workspace workspace = wsStore.getWorkspace(workspaceId);
+      workspace.setNullabilityMap(columnMappings);
+      wsStore.updateWorkspace(workspaceId, workspace);
+    } catch (Exception e) {
+      throw new RuntimeException("Error in setting nullabilityMap of columns ", e);
+    }
+    }
```
albertshau: indentation

wrangler-service/src/main/java/io/cdap/wrangler/service/directive/WorkspaceHandler.java
```diff
@@ -472,6 +474,12 @@ private DirectiveExecutionResponse execute(NamespaceSummary ns, HttpServiceReque
 
     WorkspaceDetail detail = wsStore.getWorkspaceDetail(workspaceId);
     UserDirectivesCollector userDirectivesCollector = new UserDirectivesCollector();
+    HashMap<String, UserDefinedAction> nullabilityMap = executionRequest.getNullabilityMap() == null ?
+        new HashMap<>() : executionRequest.getNullabilityMap();
+    if (!nullabilityMap.isEmpty()) {
+      //change nullabilityMap in Workspace Object
+      changeNullability(nullabilityMap, workspaceId);
```
albertshau, line 480: should update the workspace after executing the directives (this is already happening), not before. Otherwise the execution can fail and now there's a partially updated workspace.

albertshau (commented):
left general style/practice comments, but would like to understand why we want to introduce null specific handling when existing directives can already do mostly the same thing with more flexibility.

wrangler-core/src/main/java/io/cdap/wrangler/executor/RecipePipelineExecutor.java
minurajeeve: Done

wrangler-proto/src/main/java/io/cdap/wrangler/proto/workspace/v2/Workspace.java
minurajeeve: Done, But how does using the Builder instead of directly changing it ensure correctness?

wrangler-core/src/main/java/io/cdap/wrangler/executor/RecipePipelineExecutor.java
minurajeeve: Done

wrangler-core/src/main/java/io/cdap/wrangler/executor/RecipePipelineExecutor.java
minurajeeve: Done

wrangler-core/src/test/java/io/cdap/wrangler/TestingRig.java
minurajeeve: Done

wrangler-proto/src/main/java/io/cdap/wrangler/proto/workspace/v2/DirectiveExecutionRequest.java
minurajeeve: Done

wrangler-proto/src/main/java/io/cdap/wrangler/proto/workspace/v2/DirectiveExecutionRequest.java
minurajeeve: Done

wrangler-proto/src/main/java/io/cdap/wrangler/proto/workspace/v2/DirectiveExecutionRequest.java
minurajeeve: Done

wrangler-proto/src/main/java/io/cdap/wrangler/proto/workspace/v2/Workspace.java
minurajeeve: Done

wrangler-proto/src/main/java/io/cdap/wrangler/proto/workspace/v2/Workspace.java
minurajeeve: Done

wrangler-proto/src/main/java/io/cdap/wrangler/proto/workspace/v2/Workspace.java
minurajeeve: Done

wrangler-proto/src/main/java/io/cdap/wrangler/proto/workspace/v2/Workspace.java
minurajeeve: But by SEND_TO_ERROR_COLLECTOR, I actually mean the pipeline specific error collector hence the name.

wrangler-proto/src/main/java/io/cdap/wrangler/proto/workspace/v2/Workspace.java
minurajeeve: Done

wrangler-proto/src/main/java/io/cdap/wrangler/proto/workspace/v2/Workspace.java
minurajeeve: Done

wrangler-service/src/main/java/io/cdap/wrangler/service/directive/WorkspaceHandler.java
minurajeeve: Done

wrangler-service/src/main/java/io/cdap/wrangler/service/directive/WorkspaceHandler.java
minurajeeve: We need the updated nullabilityMap before directives are executed.

wrangler-service/src/main/java/io/cdap/wrangler/service/directive/WorkspaceHandler.java
minurajeeve: Done

wrangler-api/src/main/java/io/cdap/wrangler/api/NullHandlingException.java
minurajeeve: It was for future use, removed

PR #688, opened 2023-12-05
bug fix for using set-type when it's the first directive in the recipie

minurajeeve:


wrangler-core/src/main/java/io/cdap/directives/column/SetType.java
```diff
@@ -134,33 +134,35 @@ public Schema getOutputSchema(SchemaResolutionContext context) {
               if (field.getName().equals(col)) {
                 Integer outputScale = scale;
                 Integer outputPrecision = precision;
-                Schema fieldSchema = field.getSchema().getNonNullable();
-                Pair<Integer, Integer> scaleAndPrecision = getPrecisionAndScale(fieldSchema);
-                Integer inputSchemaScale = scaleAndPrecision.getSecond();
-                Integer inputSchemaPrecision = scaleAndPrecision.getFirst();
+                if (type.equalsIgnoreCase("decimal") && field.getSchema().isNullable()) {
```
vanathi-g, line 137: What happens when schema is non nullable but set type directive is applied with 'decimal' as the type?

wrangler-core/src/main/java/io/cdap/directives/column/SetType.java
```diff
@@ -134,33 +134,35 @@ public Schema getOutputSchema(SchemaResolutionContext context) {
...
+                  Schema fieldSchema = field.getSchema().getNonNullable();
```
vanathi-g, line 138: Also, I think this giant block that calculates precision and scale to use can be moved to a different function. Can refactor the function used in .map()

wrangler-core/src/main/java/io/cdap/directives/column/SetType.java
minurajeeve: If the schema is non nullable it will skip this if condition and will go to the return statement in line no:167. By default outputScale and outputPrecision are the scale and precision specified by the user. In this case we don't validate if the user specified scale and precision values clash with the original scale and precision values of that column.

wrangler-core/src/main/java/io/cdap/directives/column/SetType.java
minurajeeve: Done

PR #690, opened 2023-12-11
Setup for Wrangler SQL Execution Support

harshdeeppruthi:
This PR is responsible for the following:
1. Adding the initial setup for the SQL Execution support for Wrangler Directives using CDAP's relational API
2. Adding a UI toggle to enable this SQL Execution of directives

This includes implementation of sql support of `drop` directive as an example. Once this is merged, the SQL support for other directives can be extended in a similar way

### DIFF
--- a/wrangler-api/src/main/java/io/cdap/wrangler/api/Directive.java
+++ b/wrangler-api/src/main/java/io/cdap/wrangler/api/Directive.java
@@ -16,6 +16,8 @@
 
 package io.cdap.wrangler.api;
 
+import io.cdap.cdap.etl.api.relational.Relation;
+import io.cdap.cdap.etl.api.relational.RelationalTranformContext;
 import io.cdap.wrangler.api.parser.UsageDefinition;
 
 import java.util.List;
@@ -51,7 +53,8 @@ import java.util.List;
  *   }
  * </code>
  */
-public interface Directive extends Executor<List<Row>, List<Row>>, EntityMetrics {
+public interface Directive extends Executor<List<Row>, List<Row>>, EntityMetrics,
+    DirectiveRelationalTransform {
   /**
    * This defines a interface variable that is static and final for specify
    * the {@code type} of the plugin this interface would provide.
@@ -125,5 +128,12 @@ public interface Directive extends Executor<List<Row>, List<Row>>, EntityMetrics
   default List<EntityCountMetric> getCountMetrics() {
     // no op
     return null;
+  }
+
+  @Override
+  default Relation transform(RelationalTranformContext relationalTranformContext,
+      Relation relation) {
+    // no-op
+    return relation;
   }
 }
--- a/wrangler-transform/src/main/java/io/cdap/wrangler/Wrangler.java
+++ b/wrangler-transform/src/main/java/io/cdap/wrangler/Wrangler.java
@@ -38,7 +38,6 @@ import io.cdap.cdap.etl.api.TransformContext;
 import io.cdap.cdap.etl.api.relational.Expression;
 import io.cdap.cdap.etl.api.relational.ExpressionFactory;
 import io.cdap.cdap.etl.api.relational.InvalidRelation;
-import io.cdap.cdap.etl.api.relational.LinearRelationalTransform;
 import io.cdap.cdap.etl.api.relational.Relation;
 import io.cdap.cdap.etl.api.relational.RelationalTranformContext;
 import io.cdap.cdap.etl.api.relational.StringExpressionFactoryType;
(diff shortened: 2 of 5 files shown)


wrangler-api/src/main/java/io/cdap/wrangler/api/Directive.java
```diff
@@ -126,4 +129,11 @@ default List<EntityCountMetric> getCountMetrics() {
     // no op
     return null;
   }
+
```
minurajeeve, line 132: You could add a description for this method similar to getCountMetrics()

wrangler-api/src/main/java/io/cdap/wrangler/api/RelationalDirective.java
```diff
@@ -0,0 +1,34 @@
...
+package io.cdap.wrangler.api;
+
+import io.cdap.cdap.etl.api.relational.LinearRelationalTransform;
+import io.cdap.cdap.etl.api.relational.Relation;
+import io.cdap.cdap.etl.api.relational.RelationalTranformContext;
+
+/**
+ * Directive interface which supports Relational transformations
+ */
+public interface RelationalDirective extends Directive, LinearRelationalTransform {
```
minurajeeve: Did not understand the need for this interface

harshdeeppruthi: Had pushed this accidentally. It's not needed. Thanks fr pointing it out

wrangler-transform/src/main/java/io/cdap/wrangler/Wrangler.java
```diff
@@ -186,11 +197,11 @@ public void configurePipeline(PipelineConfigurer configurer) {
       }
 
       if (!config.containsMacro(Config.NAME_PRECONDITION_LANGUAGE)) {
-        if (PRECONDITION_LANGUAGE_SQL.equalsIgnoreCase(config.getPreconditionLanguage())) {
```
minurajeeve, line 189: Please clarify why we are modifying the precondition check

harshdeeppruthi: From now on, if the SQL execution feature flag would be enabled, SQL precondition would only be allowed if SQL execution of directives is enabled; and JEXl precondition would only be supported with non-SQL execution.

harshdeeppruthi: Also changed the precondition checking logic. PTAL

wrangler-api/src/main/java/io/cdap/wrangler/api/Directive.java
```diff
@@ -126,4 +129,11 @@ default List<EntityCountMetric> getCountMetrics() {
...
+  @Override
+  default Relation transform(RelationalTranformContext relationalTranformContext,
+      Relation relation) {
+    // no-op
```
albertshau, line 136: Looking just at this class, it seems like every existing directive would become a no-op in SQL, which would be incorrect. It only makes sense when seeing the isSQLSupported() method in the RelationalTransform interface, and that it defaults to false. 

I think it would be cleaner if the Directive interface did not extend the RelationalTransform interface, with only directives that do support SQL (like Drop) implementing it. Then the isSQLSupported() method is not needed, Wrangler just checks if the directive is an instance of RelationTransform. This way it is impossible to implement a directive where the transform() and isSQLSupported() methods don't match up.

wrangler-transform/src/main/java/io/cdap/wrangler/Wrangler.java
```diff
@@ -646,6 +717,11 @@ public static class Config extends PluginConfig {
     static final String NAME_SCHEMA = "schema";
     static final String NAME_ON_ERROR = "on-error";
 
+    @Name(NAME_SQL_EXECUTION)
+    @Description("Toggle to configure execution language between JEXL and SQL")
+    @Macro
+    @Nullable
+    private String sqlExecution;
```
albertshau, line 724: nit: add a newline after this line

wrangler-transform/src/main/java/io/cdap/wrangler/Wrangler.java
```diff
@@ -646,6 +717,11 @@ public static class Config extends PluginConfig {
     static final String NAME_SCHEMA = "schema";
     static final String NAME_ON_ERROR = "on-error";
 
+    @Name(NAME_SQL_EXECUTION)
+    @Description("Toggle to configure execution language between JEXL and SQL")
```
albertshau, line 721: JEXL is a technical term that many users won't understand. It's also not technically correct, as many directives are direct java implementations that don't use jexl scripting. Not sure what would be a better term though. Native? Default?

wrangler-transform/src/main/java/io/cdap/wrangler/Wrangler.java
albertshau: It's better to name this something like 'executionMode' and have an internal Enum for it. This is more extensible than a boolean toggle for SQL or no SQL, as it allows us to add new modes in the future if needed. (As a side note, when a property is meant to be a toggle, it should be of Boolean type, not String type).

For example, we may eventually want to add an 'auto' mode that lets the engine choose, or maybe some BigQuery specific SQL mode or something like that.

PR #695, opened 2024-01-23
CDAP-20934 - adding new param to congiure if string value needs to co…

sahusanket:
…nverted or not

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -83,7 +83,7 @@
     <aws.sdk.version>1.11.133</aws.sdk.version>
     <bigquery.connector.hadoop2.version>0.10.2-hadoop2</bigquery.connector.hadoop2.version>
     <bouncycastle.version>1.56</bouncycastle.version>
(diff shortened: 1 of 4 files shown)


pom.xml
```diff
@@ -83,7 +83,7 @@
     <aws.sdk.version>1.11.133</aws.sdk.version>
     <bigquery.connector.hadoop2.version>0.10.2-hadoop2</bigquery.connector.hadoop2.version>
     <bouncycastle.version>1.56</bouncycastle.version>
-    <cdap.version>6.10.0-SNAPSHOT</cdap.version>
+    <cdap.version>6.11.0-SNAPSHOT</cdap.version>
```
itsankit-google: use `6.10.0` unless there is any new dependency added in `6.11.x` which is required to be used by the plugin.

itsankit-google (approved):

pom.xml
sahusanket: Done

PR #697, opened 2024-02-06
[cherry-pick][6.9]CDAP-20934 - adding new param to congiure if string value needs to co…

sahusanket:
…nverted or not

Original PR : https://github.com/data-integrations/wrangler/pull/695


itsankit-google (changes requested):
The version needs to be bumped to next snapshot: https://github.com/data-integrations/wrangler/blob/97dc7a80f34ee81623ab4f0b87a88190fb83d626/pom.xml#L22

itsankit-google (approved):

PR #700, opened 2024-02-12
[PLUGIN-1736] Initialize dq_failure before use in send-to-error-and-continue directive

DJSagarAhire:
This PR adds an initialization to the `dq_failure` transitory variable used in the send-to-error Wrangler directive. Without this initialization, if the `dq_failure` variable is used in Wrangler without it ever being incremented, the directive fails as the variable is not found.

### DIFF
--- a/wrangler-core/src/main/java/io/cdap/directives/row/SendToErrorAndContinue.java
+++ b/wrangler-core/src/main/java/io/cdap/directives/row/SendToErrorAndContinue.java
@@ -125,6 +125,8 @@ public class SendToErrorAndContinue implements Directive, Lineage {
             context.getTransientStore().increment(TransientVariableScope.LOCAL, "dq_failure", 1);
           }
           throw new ReportErrorAndProceed(message, 1);
+        } else if (context != null && !context.getTransientStore().getVariables().contains("dq_failure")) {
+            context.getTransientStore().set(TransientVariableScope.LOCAL, "dq_failure", 0L);
         }
       } catch (ELException e) {
         throw new DirectiveExecutionException(NAME, e.getMessage(), e);
--- a/wrangler-core/src/test/java/io/cdap/directives/row/SendToErrorAndContinueTest.java
+++ b/wrangler-core/src/test/java/io/cdap/directives/row/SendToErrorAndContinueTest.java
@@ -107,4 +107,29 @@ public class SendToErrorAndContinueTest {
     Assert.assertEquals(2, errors.size());
     Assert.assertEquals(3, results.size());
   }
+
+  @Test
+  public void testErrorConditionFalseAndContinueWithTransientVariable() throws Exception {
+    String[] directives = new String[] {
+            "parse-as-csv body , true",
+            "drop body",
+            "send-to-error-and-continue exp:{body_3 == 'xyzw'} 'invalid value'",
+            "send-to-error-and-continue exp:{body_4=='1000'} 'junk' ",
+            "send-to-error exp:{dq_failure >= 1} "
+    };
+
+    List<Row> rows = Arrays.asList(
+            new Row("body", "1020134.298,,1,2,2 "),
+            new Row("body", "1020134.298,,xx,1,3"),
+            new Row("body", "1020134.298,,4,1,4"),
+            new Row("body", "1020134.298,,4,2,5"),
+            new Row("body", "1020134.298,,1,2,1")
+    );
+
+    RecipePipeline pipeline = TestingRig.execute(directives);
+    List<Row> results = pipeline.execute(rows);
+    List<ErrorRecord> errors = pipeline.errors();
+    Assert.assertEquals(0, errors.size());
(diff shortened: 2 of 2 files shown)


itsankit-google (commented):
JIRA also mentions another issue: `Even though send-to-error condition is false, record was send to error port.`

Please add unit tests.

wrangler-core/src/main/java/io/cdap/wrangler/executor/RecipePipelineExecutor.java
```diff
@@ -127,6 +127,7 @@ public List<Row> execute(List<Row> rows) throws RecipeException {
         // Resets the scope of local variable.
         if (context != null) {
           context.getTransientStore().reset(TransientVariableScope.LOCAL);
+          context.getTransientStore().set(TransientVariableScope.LOCAL, "dq_failure", 0L);
```
itsankit-google: Please add a comment why this variable is being initialized here.

iam-divya: If you need this variable to be maintained across the rows, it should be made global. Any reason it is kept as local? 
Also there is another question in the bug if it will be set per record or for all records. Can you please test that by setting a global vs local variable.

DJSagarAhire: As per the [documentation](https://cdap.atlassian.net/wiki/spaces/DOCS/pages/382140616/Send+to+Error+and+continue+processing+directive), the `dq_failure` variable is intended to be local as it calculates the number of errors within a record to figure out if it is worth proceeding with processing the record. So this variable is intentionally not retained across rows.

itsankit-google: +1 to @DJSagarAhire's comment.

iam-divya: Got it, it makes sense to keep it as local. Customer is also asking for a way to count the errored rows across the rows. Do we show that metric anywhere in the pipeline? Do we have to create a feature request for that?

DJSagarAhire: To the best of my knowledge there is no count of rows that have errors but are not sent to the error port. For rows that are sent it is easy to find out but for rows that are not sent we'll need another code change.

iam-divya: Please check if feature request is needed for that.

iam-divya (approved):

wrangler-core/src/main/java/io/cdap/wrangler/executor/RecipePipelineExecutor.java
DJSagarAhire: Done.

itsankit-google (changes requested):
Why can't we add an else condition here: https://github.com/data-integrations/wrangler/blob/ddff0132eec8a359178837388e81d0cbae415da3/wrangler-core/src/main/java/io/cdap/directives/row/SendToErrorAndContinue.java#L117-L128

And check if the variable `dq_failure` does not exist then initialize it?

DJSagarAhire:
@itsankit-google I feel it's better to initialize the variable at the very beginning where it will be picked up regardless of what happens in subsequently used directives rather than rely on some potentially brittle condition. Lmk if you agree.

itsankit-google:
> @itsankit-google I feel it's better to initialize the variable at the very beginning where it will be picked up regardless of what happens in subsequently used directives rather than rely on some potentially brittle condition. Lmk if you agree.

According to the [documentation](https://cdap.atlassian.net/wiki/spaces/DOCS/pages/382140616/Send+to+Error+and+continue+processing+directive#Example), we don't expect the variable to be used unless `SendToErrorAndContinue` directive is used. Otherwise the same goes with `dq_total` variable if without using `SendToErrorAndContinue` directive, the recipe tries to access `dq_total` variable they can still get a NPE.

itsankit-google:
> @itsankit-google I feel it's better to initialize the variable at the very beginning where it will be picked up regardless of what happens in subsequently used directives rather than rely on some potentially brittle condition. Lmk if you agree.

Also I didn't understand how is this condition brittle and depending on subsequently used directive, can you please give an example?

DJSagarAhire:
> > @itsankit-google I feel it's better to initialize the variable at the very beginning where it will be picked up regardless of what happens in subsequently used directives rather than rely on some potentially brittle condition. Lmk if you agree.
> 
> Also I didn't understand how is this condition brittle and depending on subsequently used directive, can you please give an example?

Never mind, I figured that referring to the variables without using Send-To-Error-And-Continue is an error anyway, so I made the change.

itsankit-google (approved):

PR #709, opened 2024-04-29
[PLUGIN-1778] Add schema management to remote task execution flow in Wrangler service

vanathi-g:
## Changes
- Add new inputSchema field to `RemoteDirectiveRequest`
- Set the inputSchema obtained from request in `TransientStore` created in `RemoteExecutionTask`
- Add a new `RemoteDirectiveResponse` class that wraps the output rows and output schema together for returning from RemoteExecutionTask

### DIFF
--- a/wrangler-api/src/main/java/io/cdap/wrangler/api/RemoteDirectiveResponse.java
+++ b/wrangler-api/src/main/java/io/cdap/wrangler/api/RemoteDirectiveResponse.java
@@ -1,0 +1,50 @@
+/*
+ * Copyright © 2024 Cask Data, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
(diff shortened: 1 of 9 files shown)


tivv (commented):
This looks like a specific case handling. Any other TransientStore usage won't work. Should we just serialize and send the whole store over the wire both ways?

vanathi-g:
Thanks for pointing that out @tivv. I did initially implement that approach with the type handler for serializing the transient store: https://github.com/data-integrations/wrangler/commit/eb981ab5bc9b579b2ed29b60be3102cce4f15639

But then later I thought that sending the entire transient store might add unnecessary overhead and if there are serialization/deserialization issues with some other variable in the store, schema handling would break. 

We can add it back in as a follow up item with more rigorous testing maybe? Right now this issue is a release blocker so I tried to fix it with minimal changes

vanathi-g:
@samdgupi seeing that you have added a kryo serializer for rows in a previous PR under a feature flag, you have to add support for schema as well now

tivv (approved):
Please add a comment to TransientStoreKeys to update this code on any additonal values

samdgupi:
> @samdgupi seeing that you have added a kryo serializer for rows in a previous PR under a feature flag, you have to add support for schema as well now

@vanathi-g Updated KryoSeializer to support RemoteDirectiveResponse. Please check the PR and give LGTM

vanathi-g (commented):
LGTM, looks like there are build issues due to checkstyle

PR #710, opened 2024-04-29
PLUGIN-1755: Use Kryo for serialisation

samdgupi:
This change adds the support to serialize and deserialize wrangler `Row` objects using Kryo. This is used while communicating with `TaskWorker` to execute directives.

1. We have to register all classes to be serialized with `Kryo`. The list of this class is taken from `SchemaConverter`
2. `Kryo` is not thread safe hence each serializer uses its own instance
3. The version of `Kryo` is kept same as `CDAP`

We will feature gate it in a future PR


tivv (approved):
LGTM given a feature gate is added

PR #726, opened 2025-01-29
[PLUGIN-1856] Error management for Wrangler plugin

Amit-CloudSufi:
https://cdap.atlassian.net/browse/PLUGIN-1856

image

```json
[
  {
    "stageName": "Wrangler",
    "errorCategory": "Plugin-'Wrangler'",
    "errorReason": "Error in stage 'Wrangler'. Format of output schema specified is invalid. Please check the format. com.google.gson.stream.MalformedJsonException: Unterminated object at line 3 column 4 path $.type",
    "errorMessage": "Error in stage 'Wrangler'. Format of output schema specified is invalid. Please check the format. com.google.gson.stream.MalformedJsonException: Unterminated object at line 3 column 4 path $.type",
    "errorType": "USER",
    "dependency": "false"
  }
]

```

```
2025-01-30 12:36:48,237 - ERROR [spark-submitter-phase-1-bfc7d018-ded8-11ef-8496-0000009adf90:o.a.s.i.i.SparkHadoopWriter@98] - Aborting job job_202501301236463168859500037448187_0005.
org.apache.spark.SparkException: Job aborted due to stage failure: Task 0 in stage 0.0 failed 1 times, most recent failure: Lost task 0.0 in stage 0.0 (TID 0) (192.168.1.100 executor driver): org.apache.spark.SparkException: Task failed while writing rows
	at org.apache.spark.internal.io.SparkHadoopWriter$.executeTask(SparkHadoopWriter.scala:163)
	at org.apache.spark.internal.io.SparkHadoopWriter$.$anonfun$write$1(SparkHadoopWriter.scala:88)
	at org.apache.spark.scheduler.ResultTask.runTask(ResultTask.scala:90)
	at org.apache.spark.scheduler.Task.run(Task.scala:136)
	at org.apache.spark.executor.Executor$TaskRunner.$anonfun$run$3(Executor.scala:548)
	at org.apache.spark.util.Utils$.tryWithSafeFinally(Utils.scala:1504)
	at org.apache.spark.executor.Executor$TaskRunner.run(Executor.scala:551)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
	at java.lang.Thread.run(Thread.java:750)
Caused by: java.util.concurrent.ExecutionException: Error when transforming stage Wrangler: com.google.common.util.concurrent.UncheckedExecutionException: io.cdap.cdap.api.exception.WrappedStageException: Stage 'Wrangler' encountered : io.cdap.cdap.api.exception.ProgramFailureException: Error in stage 'Wrangler'. Format of output schema specified is invalid. Please check the format. com.google.gson.stream.MalformedJsonException: Unterminated object at line 3 column 4 path $.type
	at io.cdap.cdap.etl.spark.function.TransformFunction.call(TransformFunction.java:61)
	at org.apache.spark.api.java.JavaRDDLike.$anonfun$flatMap$1(JavaRDDLike.scala:125)
	at scala.collection.Iterator$$anon$11.nextCur(Iterator.scala:486)
	at scala.collection.Iterator$$anon$11.hasNext(Iterator.scala:492)
	at scala.collection.Iterator$$anon$11.hasNext(Iterator.scala:491)
	at scala.collection.Iterator$$anon$11.hasNext(Iterator.scala:491)
	at org.apache.spark.internal.io.SparkHadoopWriter$.$anonfun$executeTask$1(SparkHadoopWriter.scala:136)
	at org.apache.spark.util.Utils$.tryWithSafeFinallyAndFailureCallbacks(Utils.scala:1538)
	at org.apache.spark.internal.io.SparkHadoopWriter$.executeTask(SparkHadoopWriter.scala:135)
	... 9 more
```

```
Loading Directive exception:
```
image
image

```
Precondition excdeption:
```
Screenshot from 2025-02-10 10-45-02

```
Directive Parsing Exception
```
Screenshot from 2025-02-10 11-02-06

```
RecordConvertorException
```
Screenshot from 2025-02-10 13-05-49

### DIFF
--- a/wrangler-transform/src/main/java/io/cdap/wrangler/Wrangler.java
+++ b/wrangler-transform/src/main/java/io/cdap/wrangler/Wrangler.java
@@ -24,6 +24,9 @@ import io.cdap.cdap.api.annotation.Name;
 import io.cdap.cdap.api.annotation.Plugin;
 import io.cdap.cdap.api.data.format.StructuredRecord;
 import io.cdap.cdap.api.data.schema.Schema;
+import io.cdap.cdap.api.exception.ErrorCategory;
+import io.cdap.cdap.api.exception.ErrorType;
+import io.cdap.cdap.api.exception.ErrorUtils;
 import io.cdap.cdap.api.metrics.Metrics;
 import io.cdap.cdap.api.plugin.PluginConfig;
 import io.cdap.cdap.api.plugin.PluginProperties;
@@ -53,6 +56,7 @@ import io.cdap.wrangler.api.DirectiveParseException;
 import io.cdap.wrangler.api.EntityCountMetric;
 import io.cdap.wrangler.api.ErrorRecord;
 import io.cdap.wrangler.api.ExecutorContext;
+import io.cdap.wrangler.api.RecipeException;
 import io.cdap.wrangler.api.RecipeParser;
 import io.cdap.wrangler.api.RecipePipeline;
 import io.cdap.wrangler.api.RecipeSymbol;
@@ -243,9 +247,17 @@ public class Wrangler extends Transform<StructuredRecord, StructuredRecord> impl
           }
         }
       } catch (CompileException e) {
-        collector.addFailure("Compilation error occurred : " + e.getMessage(), null);
+        collector.addFailure(
+          String.format("Compilation error occurred, %s: %s ", e.getClass().getName(),
+            e.getMessage()), null);
       } catch (DirectiveParseException e) {
-        collector.addFailure(e.getMessage(), null);
+        collector.addFailure(
+          String.format("Error parsing directive, %s: %s", e.getClass().getName(),
+            e.getMessage()), null);
+      } catch (DirectiveLoadException e) {
+        collector.addFailure(
+          String.format("Error loading directive, %s: %s", e.getClass().getName(),
+            e.getMessage()), null);
       }
 
       // Based on the configuration create output schema.
@@ -254,18 +266,22 @@ public class Wrangler extends Transform<StructuredRecord, StructuredRecord> impl
           oSchema = Schema.parseJson(config.schema);
         }
       } catch (IOException e) {
-        collector.addFailure("Invalid output schema.", null)
-          .withConfigProperty(Config.NAME_SCHEMA).withStacktrace(e.getStackTrace());
+        collector.addFailure(
+          String.format("Invalid output schema %s: %s", e.getClass().getName(), e.getMessage()),
+          null).withConfigProperty(Config.NAME_SCHEMA).withStacktrace(e.getStackTrace());
       }
 
       // Check if jexl pre-condition is not null or empty and if so compile expression.
-      if (!config.containsMacro(Config.NAME_PRECONDITION) && !config.containsMacro(Config.NAME_PRECONDITION_LANGUAGE)) {
+      if (!config.containsMacro(Config.NAME_PRECONDITION) && !config.containsMacro(
+        Config.NAME_PRECONDITION_LANGUAGE)) {
         if (PRECONDITION_LANGUAGE_JEXL.equalsIgnoreCase(config.getPreconditionLanguage())
           && checkPreconditionNotEmpty(false)) {
           try {
             new Precondition(config.getPreconditionJEXL());
           } catch (PreconditionException e) {
-            collector.addFailure(e.getMessage(), null).withConfigProperty(Config.NAME_PRECONDITION);
+            collector.addFailure(String.format("Error compiling precondition expression, %s: %s",
+                e.getClass().getName(), e.getMessage()), null)
+              .withConfigProperty(Config.NAME_PRECONDITION);
           }
         }
       }
@@ -274,11 +290,10 @@ public class Wrangler extends Transform<StructuredRecord, StructuredRecord> impl
       if (oSchema != null) {
         configurer.getStageConfigurer().setOutputSchema(oSchema);
       }
-
     } catch (Exception e) {
-      LOG.error(e.getMessage());
       collector.addFailure("Error occurred : " + e.getMessage(), null).withStacktrace(e.getStackTrace());
     }
+    collector.getOrThrowException();
   }
 
   /**
@@ -319,7 +334,14 @@ public class Wrangler extends Transform<StructuredRecord, StructuredRecord> impl
     // Parse the recipe and extract all the instances of directives
     // to be processed for extracting lineage.
     RecipeParser recipe = getRecipeParser(context);
-    List<Directive> directives = recipe.parse();
+    List<Directive> directives;
+    try {
+      directives = recipe.parse();
+    } catch (Exception e) {
+      String errorReason = "Unable to parse recipe and extract all instances of directives.";
+      throw WranglerErrorUtil.getProgramFailureExceptionDetailsFromChain(e, errorReason, null,
+          ErrorType.USER);
+    }
     emitDirectiveMetrics(directives, context.getMetrics());
 
     LineageOperations lineageOperations = new LineageOperations(input, output, directives);
@@ -344,11 +366,13 @@ public class Wrangler extends Transform<StructuredRecord, StructuredRecord> impl
     // Based on the configuration create output schema.
     try {
       oSchema = Schema.parseJson(config.schema);
-    } catch (IOException e) {
-      throw new IllegalArgumentException(
-        String.format("Stage:%s - Format of output schema specified is invalid. Please check the format.",
-                      context.getStageName()), e
-      );
+    } catch (Exception e) {
+      String errorReason = "Invalid output schema format.";
+      String errorMessage = String.format(
+          "Format of output schema specified is invalid. Please check the format. %s: %s",
+          e.getClass().getName(), e.getMessage());
+      throw WranglerErrorUtil.getProgramFailureExceptionDetailsFromChain(e, errorReason,
+          errorMessage, ErrorType.USER);
     }
 
     // Check if jexl pre-condition is not null or empty and if so compile expression.
@@ -357,8 +381,10 @@ public class Wrangler extends Transform<StructuredRecord, StructuredRecord> impl
         && checkPreconditionNotEmpty(false)) {
         try {
           condition = new Precondition(config.getPreconditionJEXL());
-        } catch (PreconditionException e) {
-          throw new IllegalArgumentException(e.getMessage(), e);
+        } catch (Exception e) {
+          String errorReason = "Failed to evaluate precondition due to an invalid JEXL expression.";
+          throw WranglerErrorUtil.getProgramFailureExceptionDetailsFromChain(e, errorReason, null,
+              ErrorType.USER);
         }
       }
     }
@@ -367,7 +393,12 @@ public class Wrangler extends Transform<StructuredRecord, StructuredRecord> impl
       // Create the pipeline executor with context being set.
       pipeline = new RecipePipelineExecutor(recipe, ctx);
     } catch (Exception e) {
-      throw new Exception(String.format("Stage:%s - %s", getContext().getStageName(), e.getMessage()), e);
+      String errorReason = "Unable to compile the recipe and execute directives.";
+      String errorMessage = String.format(
+          "Error compiling the recipe and executing directives. " + "%s: %s",
+          e.getClass().getName(), e.getMessage());
+      throw WranglerErrorUtil.getProgramFailureExceptionDetailsFromChain(e, errorReason,
+          errorMessage, ErrorType.USER);
     }
 
     String defaultStrategy = context.getArguments().get(ERROR_STRATEGY_DEFAULT);
@@ -437,8 +468,11 @@ public class Wrangler extends Transform<StructuredRecord, StructuredRecord> impl
         }
         if (WRANGLER_FAIL_PIPELINE_FOR_ERROR.isEnabled(getContext())
             && onErrorStrategy.equalsIgnoreCase(ON_ERROR_FAIL_PIPELINE)) {
-          throw new Exception(
-              String.format("Errors in Wrangler Transformation - %s", errorMessages));
+          String errorReason = String.format("Errors in Wrangler Transformation - %s",
+              errorMessages);
+          throw ErrorUtils.getProgramFailureException(
+              new ErrorCategory(ErrorCategory.ErrorCategoryEnum.PLUGIN), errorReason, errorReason,
+              ErrorType.UNKNOWN, false, null);
         }
       }
     } catch (Exception e) {
@@ -457,8 +491,12 @@ public class Wrangler extends Transform<StructuredRecord, StructuredRecord> impl
                                    getContext().getStageName(), e.getMessage()),
           "value", String.valueOf(errorCounter)
         ));
-        throw new Exception(String.format("Stage:%s - Failing pipeline due to error : %s",
-                                          getContext().getStageName(), e.getMessage()), e);
+        String errorReason = "Error occurred while processing input data, possibly due to invalid "
+            + "transformation or schema mismatch.";
+        String errorMessage = String.format("Pipeline failed at stage:%s, %s: %s",
+            getContext().getStageName(), e.getClass().getName(), e.getMessage());
+        throw WranglerErrorUtil.getProgramFailureExceptionDetailsFromChain(e, errorReason,
+            errorMessage, ErrorType.UNKNOWN);
       }
       // If it's 'skip-on-error' we continue processing and don't emit any error records.
       return;
@@ -553,18 +591,29 @@ public class Wrangler extends Transform<StructuredRecord, StructuredRecord> impl
    * @throws DirectiveLoadException
    * @throws DirectiveParseException
    */
-  private RecipeParser getRecipeParser(StageContext context)
-    throws DirectiveLoadException, DirectiveParseException {
+  private RecipeParser getRecipeParser(StageContext context) {
 
     registry = new CompositeDirectiveRegistry(SystemDirectiveRegistry.INSTANCE, new UserDirectiveRegistry(context));
-    registry.reload(context.getNamespace());
+    try {
+      registry.reload(context.getNamespace());
+    } catch (Exception e) {
+      String errorReason = "Unable to load directive from the artifacts.";
+      throw WranglerErrorUtil.getProgramFailureExceptionDetailsFromChain(e, errorReason, null,
+          ErrorType.USER);
+    }
 
     String directives = config.getDirectives();
     if (config.getUDDs() != null && !config.getUDDs().trim().isEmpty()) {
       directives = String.format("#pragma load-directives %s;%s", config.getUDDs(), config.getDirectives());
     }
 
-    return new GrammarBasedParser(context.getNamespace(), new MigrateToV2(directives).migrate(), registry);
+    try {
+      return new GrammarBasedParser(context.getNamespace(), new MigrateToV2(directives).migrate(), registry);
+    } catch (Exception e) {
+      String errorReason = "Unable to parse the directives.";
+      throw WranglerErrorUtil.getProgramFailureExceptionDetailsFromChain(e, errorReason, null,
+          ErrorType.USER);
+    }
   }
 
   @Override
@@ -573,7 +622,10 @@ public class Wrangler extends Transform<StructuredRecord, StructuredRecord> impl
             && checkPreconditionNotEmpty(true)) {
 
       if (!Feature.WRANGLER_PRECONDITION_SQL.isEnabled(relationalTranformContext)) {
-        throw new RuntimeException("SQL Precondition feature is not available");
+        String errorReason = "SQL Precondition feature is not available";
+        throw ErrorUtils.getProgramFailureException(
+            new ErrorCategory(ErrorCategory.ErrorCategoryEnum.PLUGIN), errorReason, errorReason,
+            ErrorType.SYSTEM, false, null);
       }
 
       Optional<ExpressionFactory<String>> expressionFactory = getExpressionFactory(relationalTranformContext);
@@ -598,11 +650,18 @@ public class Wrangler extends Transform<StructuredRecord, StructuredRecord> impl
    * @param directives a list of Wrangler directives
    * @param metrics CDAP {@link Metrics} object using which metrics can be emitted
    */
-  private void emitDirectiveMetrics(List<Directive> directives, Metrics metrics) throws DirectiveLoadException {
+  private void emitDirectiveMetrics(List<Directive> directives, Metrics metrics) {
     for (Directive directive : directives) {
       // skip emitting metrics if the directive is not system directive
-      if (registry.get(Contexts.SYSTEM, directive.define().getDirectiveName()) == null) {
-        continue;
+      try {
+        if (registry.get(Contexts.SYSTEM, directive.define().getDirectiveName()) == null) {
+          continue;
+        }
+      } catch (Exception e) {
+        String errorReason = String.format("Unable to load directive %s",
+            directive.define().getDirectiveName());
+        throw WranglerErrorUtil.getProgramFailureExceptionDetailsFromChain(e, errorReason, null,
+            ErrorType.USER);
       }
       List<EntityCountMetric> countMetrics = new ArrayList<>();
 
--- a/wrangler-transform/src/main/java/io/cdap/wrangler/WranglerErrorUtil.java
+++ b/wrangler-transform/src/main/java/io/cdap/wrangler/WranglerErrorUtil.java
@@ -1,0 +1,118 @@
+/*
+ * Copyright © 2025 Cask Data, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package io.cdap.wrangler;
+
+import com.google.common.base.Throwables;
+import com.google.common.collect.ImmutableMap;
+import io.cdap.cdap.api.exception.ErrorCategory;
+import io.cdap.cdap.api.exception.ErrorType;
+import io.cdap.cdap.api.exception.ErrorUtils;
+import io.cdap.cdap.api.exception.ProgramFailureException;
+import io.cdap.wrangler.api.DirectiveExecutionException;
+import io.cdap.wrangler.api.DirectiveLoadException;
+import io.cdap.wrangler.api.DirectiveNotFoundException;
+import io.cdap.wrangler.api.DirectiveParseException;
+import io.cdap.wrangler.api.RecipeException;
+import io.cdap.wrangler.expression.ELException;
+import io.cdap.wrangler.utils.RecordConvertorException;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Error util file to handle exceptions caught in Wrangler plugin
+ */
+public final class WranglerErrorUtil {
+
+
+  private static final Map<String, String> TERMINAL_EXCEPTIONS = ImmutableMap.<String, String>builder()
+      .put(DirectiveParseException.class.getName(), "Parsing-Directive")
+      .put(PreconditionException.class.getName(), "Precondition")
+      .put(DirectiveExecutionException.class.getName(), "Executing-Directive")
+      .put(DirectiveLoadException.class.getName(), "Loading-Directive")
+      .put(DirectiveNotFoundException.class.getName(), "Directive-Not-Found")
+      .put(RecordConvertorException.class.getName(), "Record-Conversion")
+      .put(ELException.class.getName(), "ExpressionLanguage-Parsing").build();
+
+  private static final Map<String, String> NON_TERMINAL_EXCEPTIONS = ImmutableMap.<String, String>builder()
+      .put(RecipeException.class.getName(), "Executing-Recipe").build();
+
+  /**
+   * Private constructor to prevent instantiation of this utility class.
+   * <p>
+   * This class is designed to contain only static utility methods for handling exceptions and
+   * should not be instantiated. Any attempt to create an instance of this class will result in an
+   * {@link IllegalStateException}.
+   */
+  private WranglerErrorUtil() {
+    throw new IllegalStateException("Utility class");
+  }
+
+  /**
+   * Traverses the causal chain of the given Throwable to find specific exceptions. If a terminal
+   * exception is found, it returns a corresponding ProgramFailureException. If a non-terminal
+   * exception is found, it is stored as a fallback. Otherwise, a generic ProgramFailureException is
+   * returned.
+   *
+   * @param e            the Throwable to analyze
+   * @param errorReason  the error reason to tell the cause of error
+   * @param errorMessage default error message if no terminal exception is found
+   * @param errorType    the error type to categorize the failure
+   * @return a ProgramFailureException with specific or generic error details
+   */
+  public static ProgramFailureException getProgramFailureExceptionDetailsFromChain(Throwable e,
+      String errorReason, String errorMessage, ErrorType errorType) {
+    List<Throwable> causalChain = Throwables.getCausalChain(e);
+    Throwable nonTerminalException = null;
+    for (Throwable t : causalChain) {
+      if (t instanceof ProgramFailureException) {
+        return null; // Avoid multiple wrap
+      }
+      if (NON_TERMINAL_EXCEPTIONS.containsKey(t.getClass().getName())) {
+        nonTerminalException = t; // Store non-terminal exception as fallback
+        continue;
+      }
+      String errorSubCategory = TERMINAL_EXCEPTIONS.get(t.getClass().getName());
+      if (errorSubCategory != null) {
+        return getProgramFailureException(t, errorReason, errorSubCategory);
+      }
+    }
+
+    if (nonTerminalException != null) {
+      return getProgramFailureException(nonTerminalException, errorReason,
+          NON_TERMINAL_EXCEPTIONS.get(nonTerminalException.getClass().getName()));
+    }
+
+    return ErrorUtils.getProgramFailureException(
+        new ErrorCategory(ErrorCategory.ErrorCategoryEnum.PLUGIN), errorReason, errorMessage,
+        errorType, false, e);
+  }
+
+  /**
+   * Constructs a ProgramFailureException using the provided exception details.
+   *
+   * @param exception        the exception to wrap
+   * @param errorSubCategory specific subcategory of the error
+   * @return a new ProgramFailureException with the extracted details
+   */
+  private static ProgramFailureException getProgramFailureException(Throwable exception,
+      String errorReason, String errorSubCategory) {
+    String errorMessage = exception.getMessage();
+    return ErrorUtils.getProgramFailureException(
+        new ErrorCategory(ErrorCategory.ErrorCategoryEnum.PLUGIN, errorSubCategory), errorReason,
+        errorMessage, ErrorType.USER, false, exception);
+  }
+}


wrangler-transform/src/main/java/io/cdap/wrangler/WranglerUtil.java
```diff
@@ -0,0 +1,79 @@
+/*
+ * Copyright © 2016-2019 Cask Data, Inc.
```
psainics: Change Year

wrangler-transform/src/main/java/io/cdap/wrangler/WranglerUtil.java
```diff
@@ -0,0 +1,79 @@
...
+ */
+public class WranglerUtil {
+
+  private WranglerUtil() {
+    throw new IllegalStateException("Utility class");
+  }
+
+  public static ProgramFailureException getProgramFailureExceptionDetailsFromChain(Throwable e,
+      String errorMessage, ErrorType errorType, boolean dependency) {
+    List<Throwable> causalChain = Throwables.getCausalChain(e);
```
psainics: Why are we looping when e is null ?

wrangler-transform/src/main/java/io/cdap/wrangler/WranglerUtil.java
Amit-CloudSufi: updated

wrangler-transform/src/main/java/io/cdap/wrangler/WranglerUtil.java
Amit-CloudSufi: added a Null check to avoid any NPE

wrangler-transform/src/main/java/io/cdap/wrangler/Precondition.java
```diff
@@ -72,20 +71,33 @@ public boolean apply(Row row) throws PreconditionException {
...
-      }
+      throw getProgramFailureExceptionDueToPrecondition(condition, e);
+    }
+  }
+
+  private static ProgramFailureException getProgramFailureExceptionDueToPrecondition(
+      String condition, ScriptException e) {
+    String errorMessage;
+    if (e.getCause() != null) {
+      Log.error(String.format("Error in evaluating precondition '%s'. %s: %s", condition,
```
psainics: Don't use this Log Object

wrangler-transform/src/main/java/io/cdap/wrangler/Wrangler.java
```diff
@@ -276,8 +288,11 @@ && checkPreconditionNotEmpty(false)) {
       }
 
     } catch (Exception e) {
-      LOG.error(e.getMessage());
-      collector.addFailure("Error occurred : " + e.getMessage(), null).withStacktrace(e.getStackTrace());
+      LOG.error("Error occurred during configuration of the plugin, {}: {}", e.getClass().getName(),
```
psainics, line 291: Same comment here for LOG object

wrangler-transform/src/main/java/io/cdap/wrangler/Wrangler.java
```diff
@@ -24,6 +24,9 @@
 import io.cdap.cdap.api.annotation.Plugin;
 import io.cdap.cdap.api.data.format.StructuredRecord;
 import io.cdap.cdap.api.data.schema.Schema;
+import io.cdap.cdap.api.exception.ErrorCategory;
```
psainics, line 27: Where is this import used ?

wrangler-transform/src/main/java/io/cdap/wrangler/Wrangler.java
```diff
@@ -24,6 +24,9 @@
...
+import io.cdap.cdap.api.exception.ErrorType;
+import io.cdap.cdap.api.exception.ErrorUtils;
```
psainics, line 29: Where is this import used ?

wrangler-transform/src/main/java/io/cdap/wrangler/WranglerUtil.java
```diff
@@ -0,0 +1,83 @@
...
+        }
+        if (t instanceof DirectiveParseException) {
+          return getProgramFailureException((DirectiveParseException) t, errorType, dependency);
+        }
+        if (t instanceof RecipeException) {
+          return getProgramFailureException((RecipeException) t, errorType, dependency);
+        }
+        if (t instanceof PreconditionException) {
+          return getProgramFailureException((PreconditionException) t, errorType, dependency);
+        }
```
psainics: since we are using the same method we can use `||` to combine them.

wrangler-transform/src/main/java/io/cdap/wrangler/Precondition.java
Amit-CloudSufi: Updated

wrangler-transform/src/main/java/io/cdap/wrangler/Wrangler.java
Amit-CloudSufi: removed

wrangler-transform/src/main/java/io/cdap/wrangler/Precondition.java
```diff
@@ -72,20 +74,33 @@ public boolean apply(Row row) throws PreconditionException {
...
+      String condition, ScriptException e) {
+    String errorMessage;
+    if (e.getCause() != null) {
+      LOG.error("Error in evaluating precondition '{}'. {}: {}", condition, e.getClass().getName(),
+          e.getCause().getMessage());
+      errorMessage = e.getCause().getMessage();
+    } else {
+      LOG.error("Error in evaluating precondition '{}'. {}: {}", condition, e.getClass().getName(),
+          e.getMessage());
+      errorMessage = e.getMessage();
```
psainics: Same log in if and else can be extracted outside

wrangler-transform/src/main/java/io/cdap/wrangler/Wrangler.java
Amit-CloudSufi: removed

wrangler-transform/src/main/java/io/cdap/wrangler/WranglerUtil.java
Amit-CloudSufi: updated

wrangler-transform/src/main/java/io/cdap/wrangler/Precondition.java
Amit-CloudSufi: updated

wrangler-transform/src/main/java/io/cdap/wrangler/WranglerUtil.java
```diff
@@ -0,0 +1,75 @@
...
+  /**
+   * Get a ProgramFailureException with the given error information from {@link Exception}.
+   *
+   * @param exception  The Exception to get the error information from.
+   * @param errorType  The ErrorType to get the error type information.
+   * @param dependency The dependency to show if it depends on external source or not
+   * @return A ProgramFailureException with the given error information.
+   */
+  private static ProgramFailureException getProgramFailureException(Exception exception,
+      ErrorType errorType, boolean dependency) {
```
psainics, line 68: Is this method required ?

wrangler-transform/src/main/java/io/cdap/wrangler/Precondition.java
```diff
@@ -72,20 +74,31 @@ public boolean apply(Row row) throws PreconditionException {
...
-      } else {
-        throw new PreconditionException(e.getMessage());
-      }
+      throw getProgramFailureExceptionDueToPrecondition(condition, e);
+    }
+  }
+
+  private static ProgramFailureException getProgramFailureExceptionDueToPrecondition(
+      String condition, ScriptException e) {
+    String errorMessage;
```
psainics: Is this method required ?

wrangler-transform/src/main/java/io/cdap/wrangler/Precondition.java
```diff
@@ -72,20 +74,31 @@ public boolean apply(Row row) throws PreconditionException {
       scriptContext.setBindings(bindings, ScriptContext.ENGINE_SCOPE);
       Object result = script.eval(scriptContext);
       if (!(result instanceof Boolean)) {
-        throw new PreconditionException(
-          String.format("Precondition '%s' does not result in true or false.", condition)
-        );
+        String errorMessage = String.format("Precondition '%s' does not result in true or false.",
+            condition);
+        throw WranglerUtil.getProgramFailureExceptionDetailsFromChain(null, errorMessage,
+            ErrorType.USER, false);
```
psainics: Why user ?

wrangler-transform/src/main/java/io/cdap/wrangler/Wrangler.java
```diff
@@ -457,8 +481,10 @@ && checkPreconditionNotEmpty(false)) {
                                    getContext().getStageName(), e.getMessage()),
           "value", String.valueOf(errorCounter)
         ));
-        throw new Exception(String.format("Stage:%s - Failing pipeline due to error : %s",
-                                          getContext().getStageName(), e.getMessage()), e);
+        String errorMessage = String.format("Pipeline failed at stage:%s, %s: %s",
+            getContext().getStageName(), e.getClass().getName(), e.getMessage());
+        throw WranglerUtil.getProgramFailureExceptionDetailsFromChain(e, errorMessage,
+            ErrorType.SYSTEM, true);
```
psainics: Why dependency true ?

wrangler-transform/src/main/java/io/cdap/wrangler/Precondition.java
Amit-CloudSufi: I think, user entered the wrong precondition, let me know if you think it can be system error?

psainics (changes requested):
Please Fix Unit Tests

psainics:
E2E Fixed in https://github.com/data-integrations/wrangler/pull/727 !

wrangler-transform/src/test/java/io/cdap/wrangler/PreconditionTest.java
```diff
@@ -34,7 +35,7 @@ public void testPrecondition() throws Exception {
     Assert.assertEquals(false, new Precondition("false").apply(row));
   }
 
-  @Test(expected = PreconditionException.class)
+  @Test(expected = ProgramFailureException.class)
```
psainics: Please match error message to verify we are catching the correct error !

Amit-CloudSufi: done matched

wrangler-transform/src/main/java/io/cdap/wrangler/Wrangler.java
```diff
@@ -345,10 +362,11 @@ public void initialize(TransformContext context) throws Exception {
     try {
       oSchema = Schema.parseJson(config.schema);
     } catch (IOException e) {
-      throw new IllegalArgumentException(
-        String.format("Stage:%s - Format of output schema specified is invalid. Please check the format.",
-                      context.getStageName()), e
-      );
+      String errorMessage = String.format("Error in stage '%s'. Format of output schema specified "
```
itsankit-google: please don't use `Error in stage` anywhere. 

`stageName` is already present in `errorCategory`.

wrangler-transform/src/main/java/io/cdap/wrangler/WranglerUtil.java
```diff
@@ -0,0 +1,75 @@
...
+import io.cdap.cdap.api.exception.ProgramFailureException;
+import io.cdap.wrangler.api.DirectiveLoadException;
+import io.cdap.wrangler.api.DirectiveParseException;
+import io.cdap.wrangler.api.RecipeException;
+import java.util.List;
+
+/**
+ * Util file to handle exceptions caught in Wrangler plugin
+ */
+public class WranglerUtil {
```
itsankit-google: make this class final. 

rename it to `WranglerErrorUtil`

wrangler-transform/src/main/java/io/cdap/wrangler/Wrangler.java
```diff
@@ -367,7 +385,11 @@ && checkPreconditionNotEmpty(false)) {
       // Create the pipeline executor with context being set.
       pipeline = new RecipePipelineExecutor(recipe, ctx);
     } catch (Exception e) {
-      throw new Exception(String.format("Stage:%s - %s", getContext().getStageName(), e.getMessage()), e);
+      String errorMessage = String.format(
+          "Error in stage '%s'. Please check the configuration or input data. %s: %s",
+          context.getStageName(), e.getClass().getName(), e.getMessage());
+      throw WranglerUtil.getProgramFailureExceptionDetailsFromChain(e, errorMessage,
+          ErrorType.SYSTEM, false);
```
itsankit-google: why `system` error?

wrangler-transform/src/main/java/io/cdap/wrangler/Wrangler.java
```diff
@@ -345,10 +362,11 @@ public void initialize(TransformContext context) throws Exception {
...
     } catch (IOException e) {
-      throw new IllegalArgumentException(
-        String.format("Stage:%s - Format of output schema specified is invalid. Please check the format.",
-                      context.getStageName()), e
-      );
+      String errorMessage = String.format("Error in stage '%s'. Format of output schema specified "
+              + "is invalid. Please check the format. %s: %s", context.getStageName(),
+          e.getClass().getName(), e.getMessage());
+      throw WranglerUtil.getProgramFailureExceptionDetailsFromChain(
+          new IllegalArgumentException(errorMessage, e), errorMessage, ErrorType.USER, false);
```
itsankit-google: why rewrapping the exception? 

You can directly throw `ErrorUtils.getProgramFailureException()`

Amit-CloudSufi: updated

wrangler-transform/src/main/java/io/cdap/wrangler/Wrangler.java
Amit-CloudSufi: updated

wrangler-transform/src/main/java/io/cdap/wrangler/WranglerUtil.java
Amit-CloudSufi: updated

wrangler-transform/src/main/java/io/cdap/wrangler/Wrangler.java
Amit-CloudSufi: removed

wrangler-transform/src/main/java/io/cdap/wrangler/Precondition.java
itsankit-google: why not handle it at `WranglerErrorUtil` level?

For all DirectiveParseException, DirectiveLoadException, RecipeException & PreconditionException error type can be `USER`.

wrangler-transform/src/test/java/io/cdap/wrangler/PreconditionTest.java
itsankit-google: We are still not matching `errorMessage`, only matching `Exception` class, this comment is still not resolved.

wrangler-transform/src/main/java/io/cdap/wrangler/Precondition.java
Amit-CloudSufi: At few places, we have System error ,  that's why have made this method generic. If we put `null` at error type, by default it will be set to `USER` in WranglerErrorUtil class.

wrangler-transform/src/main/java/io/cdap/wrangler/Precondition.java
Amit-CloudSufi: reverted

wrangler-transform/src/main/java/io/cdap/wrangler/Wrangler.java
Amit-CloudSufi: updated to false

psainics (approved):

wrangler-transform/src/main/java/io/cdap/wrangler/WranglerErrorUtil.java
```diff
@@ -0,0 +1,76 @@
...
+      List<Throwable> causalChain = Throwables.getCausalChain(e);
+      for (Throwable t : causalChain) {
+        if (t instanceof ProgramFailureException) {
+          // Avoid double wrap
+          return (ProgramFailureException) t;
+        }
+        if (t instanceof DirectiveLoadException || t instanceof DirectiveParseException
+            || t instanceof RecipeException || t instanceof PreconditionException) {
+          return getProgramFailureException((Exception) t,
+              errorType != null ? errorType : ErrorType.USER, false);
```
itsankit-google: based on exception, there can be different `subCategory` like `Parsing-Directive`, `Loading-Directive`, `Executing-Recipe` & `Precondition`

wrangler-transform/src/main/java/io/cdap/wrangler/Precondition.java
```diff
@@ -72,9 +73,10 @@ public boolean apply(Row row) throws PreconditionException {
       scriptContext.setBindings(bindings, ScriptContext.ENGINE_SCOPE);
       Object result = script.eval(scriptContext);
       if (!(result instanceof Boolean)) {
-        throw new PreconditionException(
-          String.format("Precondition '%s' does not result in true or false.", condition)
-        );
+        String errorMessage = String.format("Precondition '%s' does not result in true or false.",
+            condition);
+        throw WranglerErrorUtil.getProgramFailureExceptionDetailsFromChain(null, errorMessage,
+            ErrorType.USER);
```
itsankit-google: why this change?

Why can't we let it throw `PreConditionException` and catch it in `Wrangler.java` to handle it using `WranglerErrorUtil`?

wrangler-transform/src/main/java/io/cdap/wrangler/Wrangler.java
```diff
@@ -276,8 +286,11 @@ && checkPreconditionNotEmpty(false)) {
       }
 
     } catch (Exception e) {
-      LOG.error(e.getMessage());
-      collector.addFailure("Error occurred : " + e.getMessage(), null).withStacktrace(e.getStackTrace());
+      LOG.error("Error occurred during configuration of the plugin, {}: {}", e.getClass().getName(),
```
itsankit-google: be specific in error message what `configuration` are we referring to here?

wrangler-transform/src/main/java/io/cdap/wrangler/Wrangler.java
```diff
@@ -437,8 +460,10 @@ && checkPreconditionNotEmpty(false)) {
         }
         if (WRANGLER_FAIL_PIPELINE_FOR_ERROR.isEnabled(getContext())
             && onErrorStrategy.equalsIgnoreCase(ON_ERROR_FAIL_PIPELINE)) {
-          throw new Exception(
-              String.format("Errors in Wrangler Transformation - %s", errorMessages));
+          String errorReason = String.format("Errors in Wrangler Transformation - %s",
+              errorMessages);
+          throw WranglerErrorUtil.getProgramFailureExceptionDetailsFromChain(null, errorReason,
+              ErrorType.SYSTEM);
```
itsankit-google: why `SYSTEM`?

wrangler-transform/src/main/java/io/cdap/wrangler/Wrangler.java
```diff
@@ -457,8 +482,10 @@ && checkPreconditionNotEmpty(false)) {
                                    getContext().getStageName(), e.getMessage()),
           "value", String.valueOf(errorCounter)
         ));
-        throw new Exception(String.format("Stage:%s - Failing pipeline due to error : %s",
-                                          getContext().getStageName(), e.getMessage()), e);
+        String errorMessage = String.format("Pipeline failed at stage:%s, %s: %s",
+            getContext().getStageName(), e.getClass().getName(), e.getMessage());
+        throw WranglerErrorUtil.getProgramFailureExceptionDetailsFromChain(e, errorMessage,
+            ErrorType.SYSTEM);
```
itsankit-google: why `SYSTEM`?

wrangler-transform/src/main/java/io/cdap/wrangler/Wrangler.java
```diff
@@ -553,18 +580,32 @@ private boolean checkPreconditionNotEmpty(Boolean isConditionSQL) {
...
     registry = new CompositeDirectiveRegistry(SystemDirectiveRegistry.INSTANCE, new UserDirectiveRegistry(context));
-    registry.reload(context.getNamespace());
+    try {
+      registry.reload(context.getNamespace());
+    } catch (DirectiveLoadException e) {
+      LOG.error("Failed to reload the directive registry for namespace "
+              + "'{}' at stage '{}'. Please verify the namespace and ensure the directives are "
+              + "correctly configured. {}: {}", context.getNamespace(), context.getStageName(),
+          e.getClass().getName(), e.getMessage());
+      throw WranglerErrorUtil.getProgramFailureExceptionDetailsFromChain(e, null, ErrorType.USER);
```
itsankit-google: why are we logging if we are already throwing the exception?
This is an anti-pattern. Please do not log anywhere if we are already throwing the exception. Instead pass proper descriptive errorMessage in `ProgramFailureException`.

If you log the exception and then throw it, there's a chance that the exception will be logged again further up the call stack, resulting in two log events for the same error.

This comment applies to whole PR.

wrangler-transform/src/test/java/io/cdap/wrangler/PreconditionTest.java
itsankit-google: The above `expected = ProgramFailureException` is not needed if you are already catching exception in the test.

itsankit-google (changes requested):
Lots of basic issues in the PR. PTAL, thanks!

itsankit-google (commented):
please add evidences/screenshots of all four known exceptions being captured properly in ErrorManagement UI.

`Parsing-Directive, Loading-Directive, Executing-Recipe & Precondition`

wrangler-transform/src/main/java/io/cdap/wrangler/Wrangler.java
```diff
@@ -169,115 +171,124 @@ public void configurePipeline(PipelineConfigurer configurer) {
...
-        } else {
-          directives = String.format("#pragma load-directives %s;%s", config.getUDDs(), config.getDirectives());
-        }
+    String directives = config.getDirectives();
+    if (config.getUDDs() != null && !config.getUDDs().trim().isEmpty()) {
+      if (config.containsMacro("directives")) {
+        directives = String.format("#pragma load-directives %s;", config.getUDDs());
+      } else {
+        directives = String.format("#pragma load-directives %s;%s", config.getUDDs(),
+            config.getDirectives());
```
itsankit-google: indentation

wrangler-transform/src/main/java/io/cdap/wrangler/Wrangler.java
```diff
@@ -169,115 +171,124 @@ public void configurePipeline(PipelineConfigurer configurer) {
     super.configurePipeline(configurer);
     FailureCollector collector = configurer.getStageConfigurer().getFailureCollector();
 
-    try {
```
itsankit-google: why this try-catch is removed?

wrangler-transform/src/main/java/io/cdap/wrangler/Wrangler.java
```diff
@@ -169,115 +171,124 @@ public void configurePipeline(PipelineConfigurer configurer) {
...
       }
+    }
 
-      // Set the output schema.
-      if (oSchema != null) {
-        configurer.getStageConfigurer().setOutputSchema(oSchema);
-      }
-
-    } catch (Exception e) {
-      LOG.error(e.getMessage());
```
itsankit-google, line 279: we could have just removed this log line instead of whole try-catch, any specific reason?

wrangler-transform/src/main/java/io/cdap/wrangler/Wrangler.java
```diff
@@ -598,11 +633,16 @@ private Optional<ExpressionFactory<String>> getExpressionFactory(RelationalTranf
...
       // skip emitting metrics if the directive is not system directive
-      if (registry.get(Contexts.SYSTEM, directive.define().getDirectiveName()) == null) {
-        continue;
+      try {
+        if (registry.get(Contexts.SYSTEM, directive.define().getDirectiveName()) == null) {
+          continue;
+        }
+      } catch (DirectiveLoadException e) {
+        throw WranglerErrorUtil.getProgramFailureExceptionDetailsFromChain(e, null,
+            ErrorType.SYSTEM);
```
itsankit-google: This error can come while loading a user directive, why is this system error?

wrangler-transform/src/main/java/io/cdap/wrangler/Wrangler.java
```diff
@@ -274,9 +288,7 @@ && checkPreconditionNotEmpty(false)) {
       if (oSchema != null) {
         configurer.getStageConfigurer().setOutputSchema(oSchema);
       }
-
     } catch (Exception e) {
-      LOG.error(e.getMessage());
       collector.addFailure("Error occurred : " + e.getMessage(), null).withStacktrace(e.getStackTrace());
     }
```
itsankit-google, line 295: please do `collector.getOrThrowException()` at the end

wrangler-transform/src/main/java/io/cdap/wrangler/Wrangler.java
```diff
@@ -274,10 +288,9 @@ && checkPreconditionNotEmpty(false)) {
       if (oSchema != null) {
         configurer.getStageConfigurer().setOutputSchema(oSchema);
       }
-
     } catch (Exception e) {
-      LOG.error(e.getMessage());
       collector.addFailure("Error occurred : " + e.getMessage(), null).withStacktrace(e.getStackTrace());
+      collector.getOrThrowException();
```
itsankit-google: This needs to be done outside the `catch` block

wrangler-transform/src/main/java/io/cdap/wrangler/WranglerErrorUtil.java
```diff
@@ -0,0 +1,87 @@
...
+   * @return A ProgramFailureException with the given error information.
+   */
+  private static ProgramFailureException getProgramFailureException(Exception exception,
+      ErrorType errorType, boolean dependency, String errorSubCategory) {
+    String errorMessage = exception.getMessage();
+    return ErrorUtils.getProgramFailureException(
+        new ErrorCategory(ErrorCategory.ErrorCategoryEnum.PLUGIN, errorSubCategory), errorMessage,
+        errorMessage, errorType, dependency, exception);
+  }
+
```
itsankit-google: nit: remove empty line

wrangler-transform/src/main/java/io/cdap/wrangler/Wrangler.java
```diff
@@ -274,11 +288,11 @@ && checkPreconditionNotEmpty(false)) {
       if (oSchema != null) {
         configurer.getStageConfigurer().setOutputSchema(oSchema);
       }
-
     } catch (Exception e) {
-      LOG.error(e.getMessage());
       collector.addFailure("Error occurred : " + e.getMessage(), null).withStacktrace(e.getStackTrace());
     }
+
```
itsankit-google: nit: remove empty line

wrangler-transform/src/main/java/io/cdap/wrangler/WranglerErrorUtil.java
```diff
@@ -0,0 +1,86 @@
...
+/**
+ * Error util file to handle exceptions caught in Wrangler plugin
+ */
+public final class WranglerErrorUtil {
+
+  private WranglerErrorUtil() {
+    throw new IllegalStateException("Utility class");
+  }
+
+  public static ProgramFailureException getProgramFailureExceptionDetailsFromChain(Throwable e,
```
itsankit-google, line 76: please add javadocs for public methods.

wrangler-transform/src/main/java/io/cdap/wrangler/WranglerErrorUtil.java
```diff
@@ -0,0 +1,107 @@
...
+              errorType != null ? errorType : ErrorType.USER, false, "Parsing-Directive");
+        }
+        if (t instanceof RecipeException) {
+          return getProgramFailureException((RecipeException) t,
+              errorType != null ? errorType : ErrorType.USER, false, "Executing-Recipe");
+        }
+        if (t instanceof PreconditionException) {
+          return getProgramFailureException((PreconditionException) t,
+              errorType != null ? errorType : ErrorType.USER, false, "Precondition");
+        }
```
itsankit-google: There are few more known exceptions like `DirectiveNotFoundException`, `DirectiveExecutionException`

wrangler-transform/pom.xml
```diff
@@ -84,6 +84,12 @@
         </exclusion>
       </exclusions>
     </dependency>
+    <dependency>
+      <groupId>io.cdap.wrangler</groupId>
+      <artifactId>wrangler-storage</artifactId>
+      <version>4.12.0-SNAPSHOT</version>
+      <scope>compile</scope>
+    </dependency>
```
itsankit-google: It is not recommended to add this dependency in wrangler-transform.

Please revert this change.

wrangler-transform/src/main/java/io/cdap/wrangler/WranglerErrorUtil.java
```diff
@@ -0,0 +1,202 @@
...
+   * it creates and returns a ProgramFailureException with appropriate error details.
+   * <p>
+   * If no predefined exception is found in the causal chain, a generic ProgramFailureException is
+   * returned.
+   *
+   * @param exception the Throwable to get the error information from
+   * @param errorType the type of error to be used if no specific exception is found
+   * @return a ProgramFailureException with the given error information
+   */
+  private static ProgramFailureException getProgramFailureExceptionWrappedUnderRecipeException(
```
itsankit-google: There is no need to create these new methods.

You can just check while traversing the original causal chain, if the exception `isTerminal`, if not continue traversing the causal chain and keep on storing/overring the result.

You can create a list of terminal exceptions and error category corresponding to it.

At the end if result is not null return it, otherwise return generic programFailureException.

itsankit-google (commented):
`wrangler-storage` is not used in `wrangler-transform` no need to wrap exceptions from that module.

wrangler-transform/src/main/java/io/cdap/wrangler/WranglerErrorUtil.java
```diff
@@ -0,0 +1,116 @@
...
+  static {
+    TERMINAL_EXCEPTIONS = new HashMap<>();
+    TERMINAL_EXCEPTIONS.put(DirectiveParseException.class, "Parsing-Directive");
+    TERMINAL_EXCEPTIONS.put(RecipeException.class, "Executing-Recipe");
+    TERMINAL_EXCEPTIONS.put(PreconditionException.class, "Precondition");
+    TERMINAL_EXCEPTIONS.put(DirectiveExecutionException.class, "Executing-Directive");
+    TERMINAL_EXCEPTIONS.put(DirectiveLoadException.class, "Loading-Directive");
+    TERMINAL_EXCEPTIONS.put(DirectiveNotFoundException.class, "NotFound-Directive");
+    TERMINAL_EXCEPTIONS.put(RecordConvertorException.class, "RecordConvertor");
+    TERMINAL_EXCEPTIONS.put(ELException.class, "ExpressionLanguage-Parsing");
```
itsankit-google: if everything is terminal then again it will not go further in causal chain when `DirectiveNotFoundException` was wrapped in `RecipeException`.

According to the finding `RecipeException` should not have been terminal right?

wrangler-transform/src/main/java/io/cdap/wrangler/WranglerErrorUtil.java
```diff
@@ -0,0 +1,116 @@
...
+      String errorMessage, ErrorType errorType) {
+    if (e == null) {
+      return ErrorUtils.getProgramFailureException(
+          new ErrorCategory(ErrorCategory.ErrorCategoryEnum.PLUGIN), errorMessage, errorMessage,
+          errorType, false, null);
+    }
+
+    ProgramFailureException result = null;
+    List<Throwable> causalChain = Throwables.getCausalChain(e);
+    for (int i = causalChain.size() - 1; i >= 0; i--) {
```
itsankit-google: please use `for each` and there is no reason to go from bottom to top as explained above.

wrangler-transform/src/main/java/io/cdap/wrangler/WranglerErrorUtil.java
Amit-CloudSufi: it will go, since we are looping from backward, then we will first get DirectiveNotFoundException, then once we get the child exception we will break out of loop, no need to go further to check the parent exception I.e. Recipe exception

itsankit-google: It does not matter whether you loop from front or back, the actual problem considering an exception as terminal which is not.

If you consider `RecipeException` as non-terminal then we can still further go in causal chain and there will be no need for you to loop in reverse order.

wrangler-transform/src/main/java/io/cdap/wrangler/WranglerErrorUtil.java
```diff
@@ -0,0 +1,125 @@
...
+    for (Throwable t : causalChain) {
+      if (t instanceof ProgramFailureException) {
+        return null; // Avoid multiple wrap
+      }
+      if (t instanceof PreconditionException) {
+        return getProgramFailureException((PreconditionException) t, errorType, "Precondition");
+      }
+      if (t instanceof RecipeException) {
+        // Store RecipeException as fallback, if there is only Recipe exception in casual chain
+        recipeException = (RecipeException) t;
```
itsankit-google: can we also create a list of `non-terminal_exceptions` with category to make it extensible?

wrangler-transform/src/main/java/io/cdap/wrangler/WranglerErrorUtil.java
```diff
@@ -0,0 +1,128 @@
...
+    TERMINAL_EXCEPTIONS.put(DirectiveParseException.class, "Parsing-Directive");
+    TERMINAL_EXCEPTIONS.put(DirectiveExecutionException.class, "Executing-Directive");
+    TERMINAL_EXCEPTIONS.put(DirectiveLoadException.class, "Loading-Directive");
+    TERMINAL_EXCEPTIONS.put(DirectiveNotFoundException.class, "NotFound-Directive");
+    TERMINAL_EXCEPTIONS.put(RecordConvertorException.class, "RecordConvertor");
+    TERMINAL_EXCEPTIONS.put(ELException.class, "ExpressionLanguage-Parsing");
+
+    NON_TERMINAL_EXCEPTIONS = new HashMap<>();
+    NON_TERMINAL_EXCEPTIONS.put(RecipeException.class, "Executing-Recipe");
+    NON_TERMINAL_EXCEPTIONS.put(PreconditionException.class, "Precondition");
```
itsankit-google: why is `PreconditionException` non-terminal?

Amit-CloudSufi: this is not getting wrapped anywhere, we are directly returning Precondition exception if provided precondition is wrong

itsankit-google: `Non-terminal` exception means it can wrap any exception like `DirectiveParseException` which are terminal exceptions.

wrangler-transform/src/main/java/io/cdap/wrangler/WranglerErrorUtil.java
```diff
@@ -0,0 +1,128 @@
...
+    TERMINAL_EXCEPTIONS.put(DirectiveExecutionException.class, "Executing-Directive");
+    TERMINAL_EXCEPTIONS.put(DirectiveLoadException.class, "Loading-Directive");
+    TERMINAL_EXCEPTIONS.put(DirectiveNotFoundException.class, "NotFound-Directive");
+    TERMINAL_EXCEPTIONS.put(RecordConvertorException.class, "RecordConvertor");
+    TERMINAL_EXCEPTIONS.put(ELException.class, "ExpressionLanguage-Parsing");
+    TERMINAL_EXCEPTIONS.put(PreconditionException.class, "Precondition");
+
+    NON_TERMINAL_EXCEPTIONS = new HashMap<>();
+    NON_TERMINAL_EXCEPTIONS.put(RecipeException.class, "Executing-Recipe");
+  }
```
itsankit-google: In Java, we generally create an immutable map instead using a hashMap, if it is not meant to be modified later:

```
private static final Map<String, String> TERMINAL_EXCEPTIONS = ImmutableMap.<String, String>builder()
            .put(DirectiveParseException.class.getName(), "Parsing-Directive")
            ........
            .build();

private static final Map<String, String> NON_TERMINAL_EXCEPTIONS = ImmutableMap.<String>, String>builder()
            .put(RecipeException.class.getName(), "Executing-Recipe")
            .build();
```

wrangler-transform/src/main/java/io/cdap/wrangler/WranglerErrorUtil.java
```diff
@@ -0,0 +1,124 @@
...
+public final class WranglerErrorUtil {
+
+
+  private static final Map<String, String> TERMINAL_EXCEPTIONS = ImmutableMap.<String, String>builder()
+      .put(DirectiveParseException.class.getName(), "Parsing-Directive")
+      .put(PreconditionException.class.getName(), "Precondition")
+      .put(DirectiveExecutionException.class.getName(), "Executing-Directive")
+      .put(DirectiveLoadException.class.getName(), "Loading-Directive")
+      .put(DirectiveNotFoundException.class.getName(), "NotFound-Directive")
+      .put(RecordConvertorException.class.getName(), "RecordConvertor")
```
itsankit-google: nit: `Record-Conversion`

wrangler-transform/src/main/java/io/cdap/wrangler/WranglerErrorUtil.java
```diff
@@ -0,0 +1,124 @@
...
+ */
+public final class WranglerErrorUtil {
+
+
+  private static final Map<String, String> TERMINAL_EXCEPTIONS = ImmutableMap.<String, String>builder()
+      .put(DirectiveParseException.class.getName(), "Parsing-Directive")
+      .put(PreconditionException.class.getName(), "Precondition")
+      .put(DirectiveExecutionException.class.getName(), "Executing-Directive")
+      .put(DirectiveLoadException.class.getName(), "Loading-Directive")
+      .put(DirectiveNotFoundException.class.getName(), "NotFound-Directive")
```
itsankit-google: `Directive-Not-Found`

wrangler-transform/src/main/java/io/cdap/wrangler/WranglerErrorUtil.java
```diff
@@ -0,0 +1,124 @@
...
+      if (t instanceof ProgramFailureException) {
+        return null; // Avoid multiple wrap
+      }
+      if (NON_TERMINAL_EXCEPTIONS.containsKey(t.getClass().getName())) {
+        nonTerminalException = t; // Store non-terminal exception as fallback
+        continue;
+      }
+      String errorSubCategory = TERMINAL_EXCEPTIONS.get(t.getClass().getName());
+      if (errorSubCategory != null) {
+        return getProgramFailureException((Exception) t, ErrorType.USER, errorSubCategory);
```
itsankit-google: why is this typecast needed?

wrangler-transform/src/main/java/io/cdap/wrangler/WranglerErrorUtil.java
```diff
@@ -0,0 +1,124 @@
...
+
+  /**
+   * Constructs a ProgramFailureException using the provided exception details.
+   *
+   * @param exception        the exception to wrap
+   * @param errorType        type of error categorization
+   * @param errorSubCategory specific subcategory of the error
+   * @return a new ProgramFailureException with the extracted details
+   */
+  private static ProgramFailureException getProgramFailureException(Exception exception,
```
itsankit-google: please change it to accept Throwable instead

wrangler-transform/src/main/java/io/cdap/wrangler/Wrangler.java
```diff
@@ -319,7 +332,12 @@ public void prepareRun(StageSubmitterContext context) throws Exception {
     // Parse the recipe and extract all the instances of directives
     // to be processed for extracting lineage.
     RecipeParser recipe = getRecipeParser(context);
-    List<Directive> directives = recipe.parse();
+    List<Directive> directives;
+    try {
+      directives = recipe.parse();
+    } catch (RecipeException e) {
```
itsankit-google: `catch(Exception e)` here similarly at all other places in the PR.

wrangler-transform/src/main/java/io/cdap/wrangler/Wrangler.java
```diff
@@ -319,7 +332,12 @@ public void prepareRun(StageSubmitterContext context) throws Exception {
...
+      throw WranglerErrorUtil.getProgramFailureExceptionDetailsFromChain(e, null, ErrorType.USER);
```
itsankit-google: Please also take `errorReason` as parameter in the function.

`String errorReason = Unable to parse recipe and extract all instances of directives`.

similarly on all other places in the PR.

wrangler-transform/src/main/java/io/cdap/wrangler/Wrangler.java
```diff
@@ -553,18 +588,29 @@ private boolean checkPreconditionNotEmpty(Boolean isConditionSQL) {
...
-  private RecipeParser getRecipeParser(StageContext context)
-    throws DirectiveLoadException, DirectiveParseException {
+  private RecipeParser getRecipeParser(StageContext context) {
 
     registry = new CompositeDirectiveRegistry(SystemDirectiveRegistry.INSTANCE, new UserDirectiveRegistry(context));
-    registry.reload(context.getNamespace());
+    try {
+      registry.reload(context.getNamespace());
+    } catch (Exception e) {
+      String errorReason = "Unable to re-load directive from the artifacts.";
```
itsankit-google: `Unable to load .....`

wrangler-transform/src/main/java/io/cdap/wrangler/Wrangler.java
```diff
@@ -553,18 +588,29 @@ private boolean checkPreconditionNotEmpty(Boolean isConditionSQL) {
...
     String directives = config.getDirectives();
     if (config.getUDDs() != null && !config.getUDDs().trim().isEmpty()) {
       directives = String.format("#pragma load-directives %s;%s", config.getUDDs(), config.getDirectives());
     }
 
-    return new GrammarBasedParser(context.getNamespace(), new MigrateToV2(directives).migrate(), registry);
+    try {
+      return new GrammarBasedParser(context.getNamespace(), new MigrateToV2(directives).migrate(), registry);
+    } catch (Exception e) {
+      String errorReason = "Unable to parse directive from the artifacts, few directives are missing.";
```
itsankit-google: `Unable to parse the directives.`

wrangler-transform/src/main/java/io/cdap/wrangler/WranglerErrorUtil.java
```diff
@@ -0,0 +1,123 @@
...
+   * returned.
+   *
+   * @param e            the Throwable to analyze
+   * @param errorMessage default error message if no terminal exception is found
+   * @param errorType    the error type to categorize the failure
+   * @return a ProgramFailureException with specific or generic error details
+   */
+  public static ProgramFailureException getProgramFailureExceptionDetailsFromChain(Throwable e,
+      String errorMessage, String errorReason, ErrorType errorType) {
+    if (e == null) {
```
itsankit-google: why do we want to use this method if `e` is `null`?

Can't we just directly throw 

```
ErrorUtils.getProgramFailureException(new ErrorCategory(ErrorCategory.ErrorCategoryEnum.PLUGIN), errorReason, errorMessage, errorType, false, null)
```

Please call this method only when `e` is `notNull`.

wrangler-transform/src/main/java/io/cdap/wrangler/WranglerErrorUtil.java
```diff
@@ -0,0 +1,123 @@
...
+   * exception is found, it is stored as a fallback. Otherwise, a generic ProgramFailureException is
+   * returned.
+   *
+   * @param e            the Throwable to analyze
+   * @param errorMessage default error message if no terminal exception is found
+   * @param errorType    the error type to categorize the failure
+   * @return a ProgramFailureException with specific or generic error details
+   */
+  public static ProgramFailureException getProgramFailureExceptionDetailsFromChain(Throwable e,
+      String errorMessage, String errorReason, ErrorType errorType) {
```
itsankit-google: At all other places we are taking `errorReason` before `errorMessage`. Please keep the order of params consistent.

wrangler-transform/src/main/java/io/cdap/wrangler/Wrangler.java
```diff
@@ -437,8 +466,10 @@ && checkPreconditionNotEmpty(false)) {
         }
         if (WRANGLER_FAIL_PIPELINE_FOR_ERROR.isEnabled(getContext())
             && onErrorStrategy.equalsIgnoreCase(ON_ERROR_FAIL_PIPELINE)) {
-          throw new Exception(
-              String.format("Errors in Wrangler Transformation - %s", errorMessages));
+          String errorReason = String.format("Errors in Wrangler Transformation - %s",
+              errorMessages);
+          throw WranglerErrorUtil.getProgramFailureExceptionDetailsFromChain(null, errorReason,
```
itsankit-google: similar comment here, can we please throw `ErrorUtils.getProgramFailureException` directly from here?

wrangler-transform/src/main/java/io/cdap/wrangler/Wrangler.java
```diff
@@ -573,7 +619,9 @@ public Relation transform(RelationalTranformContext relationalTranformContext, R
             && checkPreconditionNotEmpty(true)) {
 
       if (!Feature.WRANGLER_PRECONDITION_SQL.isEnabled(relationalTranformContext)) {
-        throw new RuntimeException("SQL Precondition feature is not available");
+        String errorReason = "SQL Precondition feature is not available";
+        throw WranglerErrorUtil.getProgramFailureExceptionDetailsFromChain(null, errorReason,
```
itsankit-google: similar comment here

itsankit-google: Also this method is private, how was it tested?

wrangler-transform/src/main/java/io/cdap/wrangler/WranglerErrorUtil.java
```diff
@@ -0,0 +1,123 @@
...
+   *
+   * @param e            the Throwable to analyze
+   * @param errorReason  the error reason to tell the cause of error
+   * @param errorMessage default error message if no terminal exception is found
+   * @param errorType    the error type to categorize the failure
+   * @return a ProgramFailureException with specific or generic error details
+   */
+  public static ProgramFailureException getProgramFailureExceptionDetailsFromChain(Throwable e,
+      String errorReason, String errorMessage, ErrorType errorType) {
+    if (e != null) {
```
itsankit-google: Do we need this `null` check? We should always call this method only when `e` is not null.

itsankit-google (approved):
LGTM

Please squash commits before merge.

Please remember to cherry-pick the PR in `release/4.11`.

PR #777, opened 2025-04-12
Add ByteSize, TimeDuration support and AggregateStats directive to wrangler-core

Nitinkumar432:
Features Implemented:
🔹 New Utility Classes
ByteSize: Parses and converts data size values like "10KB", "2.5MB", etc. into canonical bytes.

TimeDuration: Parses and converts time values like "5s", "120ms", etc. into canonical nanoseconds.

🔹 New Directive: aggregate-stats
Aggregates data size and time duration across multiple rows.

Supports aggregation types: total (default) and average.

Converts the final result into user-defined units (MB, GB, seconds, minutes, etc.).

Accepts the following arguments:

Input byte column

Input time column

Output byte column

Output time column

(Optional) Unit for bytes

(Optional) Unit for time

(Optional) Aggregation type (total or average)

🔹 Parser Enhancements
Added new grammar rules to support byte size and time duration formats in recipe parsing.

Added visit methods:

visitByteSizeArg(ctx)

visitTimeDurationArg(ctx)

Modified visitValue() to support new types.

Registered new token types to the TokenGroup.

### DIFF
--- a/README.md
+++ b/README.md
@@ -226,3 +226,38 @@ To do this, run:
 
 ```bash
 mvn clean compile
+
+
+New Directive: aggregate-stats
+This update introduces a new aggregate-stats directive to the wrangler-core module that performs aggregation on byte sizes and time durations with support for unit conversion and aggregation types.
+
+🔧 Features Added
+✅ New Directive: aggregate-stats
+
+✅ New Parser Rules for:
+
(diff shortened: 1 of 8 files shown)


Nitinkumar432:
ok

PR #995, opened 2025-05-06
Added lexer and parser rules in accordance with task 1

Renegade2345:
Added lexer and parser rules


Renegade2345:
Hi,

This PR adds lexer and parser rules in accordance with Task 1.  
Requesting review and workflow approval when convenient. Thank you!

PR #1041, opened 2026-09-02
CDAP-21271 : Adding Enforcement checks for Directiveshandler and RecipeHandler

sahusanket:

PR #1045, opened 2026-09-06
[cherry-pick][6.10] CDAP-21271 : Adding Enforcement checks for Directiveshandler and RecipeHandler- #1041

sahusanket:
Original PR : #1041

### DIFF
--- a/wrangler-proto/src/main/java/io/cdap/wrangler/proto/id/DirectiveConfigEntityId.java
+++ b/wrangler-proto/src/main/java/io/cdap/wrangler/proto/id/DirectiveConfigEntityId.java
@@ -1,0 +1,29 @@
+/*
+ * Copyright © 2026 Cask Data, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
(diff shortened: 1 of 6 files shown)


wrangler-proto/src/main/java/io/cdap/wrangler/proto/id/DirectiveConfigEntityId.java
```diff
@@ -0,0 +1,33 @@
...
+import io.cdap.cdap.proto.id.SystemAppEntityId;
+
+/**
+ * Uniquely identifies the system-level dataprep directive config entity.
+ */
+public class DirectiveConfigEntityId extends SystemAppEntityId {
+
+  public static final String APP_NAME = "dataprep";
+  public static final String ENTITY_TYPE = "config";
+  public static final String ENTITY_NAME = "directives";
```
vsethi09: Does this need to be public? Can it be used inline like `wrangler-proto/src/main/java/io/cdap/wrangler/proto/id/RecipeEntityId.java`?

vsethi09 (approved):

PR #1053, opened 2026-09-17
[CDAP-21270] Initialise EL with DirectiveContext during recipe parsing

riyaa14:


wrangler-core/src/main/java/io/cdap/wrangler/expression/CompileOptions.java
sahusanket, line 31: Do we need this anymore ?

wrangler-core/src/main/java/io/cdap/wrangler/expression/EL.java
```diff
@@ -56,6 +58,12 @@ public final class EL {
 
   private static volatile boolean used;
 
+  /**
+   * Options applied to every script compiled by this class. Set through
+   * {@link #initialize(DirectiveContext)} and defaults to allowlist disabled.
+   */
+  private static CompileOptions compileOptions = CompileOptions.getDefaultCompileOptions();
```
sahusanket: Is there any place where `EL.initialize` won't be called before EL.compile ? 
If not, let's not keep any default. 

I think it is better to break rather than silently taking default value even if directiveConfig was different.

wrangler-core/src/main/java/io/cdap/wrangler/expression/CompileOptions.java
```diff
@@ -34,7 +31,7 @@ public class CompileOptions {
   private final boolean allowlistEnabled;
   private final ImmutableList<JexlAllowlist> jexlAllowlist;
 
-  private CompileOptions(boolean allowlistEnabled, @Nullable List<JexlAllowlist> jexlAllowlist) {
+  public CompileOptions(boolean allowlistEnabled, @Nullable List<JexlAllowlist> jexlAllowlist) {
```
vsethi09: Instead of using parameterized public constructor, use the static creator function to build compile options.

Something like:

```java
public static CompileOptions fromContext(DirectiveContext context) {
   ...
   ...
   // Call private parameterized constructor.
   return new CompileOptions(....)
}
```

This is for maintainability. If more options need to be added in the future, callers of public constructor won't need to be modified. We can expect most of the options to be derived from the DirectiveContext.

wrangler-core/src/main/java/io/cdap/wrangler/expression/EL.java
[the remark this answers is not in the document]
vsethi09: +1 either use ThreadLocal<> or AtomicReference<> or static volatile variable whichever suits best.

vsethi09: On the safer side for UDDs, we can consider using Default compile option.

vsethi09 (changes requested):

wrangler-core/src/main/java/io/cdap/wrangler/expression/CompileOptions.java
riyaa14: Discussed with @vsethi09 , EL will never compile with default compile options as long as EL.compile(...) is used in UDD, but just to be on the safer side with user code, we keep default compile options as a fallback.

vsethi09 (approved):

